PR #4, opened 2015-03-30
Update mocha.js to correct typos

MystK:
Corrected two typos


mousetraps:
Thanks for the PR. I know it's a small change, but would you mind signing https://cla.microsoft.com/? For future PRs, the request will be automatic... so kudos for making a PR before we even set that up. :smiley:

AndreSayreMSFT:
Looks good to me!

PR #97, opened 2015-05-05
Fixes race condition during debugger launch

### LINKED ISSUE #31: VS2015 F5 is inconsistent
Oftentimes it doesn't go any further than "debugger listening on port 5858"

(using basic express app)

### LINKED ISSUE #83: Execution doesn't starts on launch
At times when I run the project, it just shows a message "Debugger listening on port xxxx" and doesn't proceeds further. I've to manually pause which then shows the execution paused at first line of the start-up file and then click continue to make it work.

mousetraps:
#83 Execution doesn't start on launch
#31 VS2015 F5 is inconsistent

closes #83, closes #31

### DIFF
--- a/Nodejs/Product/Nodejs/Debugger/Communication/DebuggerClient.cs
+++ b/Nodejs/Product/Nodejs/Debugger/Communication/DebuggerClient.cs
@@ -74,7 +74,7 @@ namespace Microsoft.NodejsTools.Debugger.Communication {
         /// </remarks>
         public static async void RunWithRequestExceptionsHandled(Func<Task> action) {
             try {
-                await action();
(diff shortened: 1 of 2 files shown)


AndreSayreMSFT:
Looks good to me

AndreSayreMSFT:
Can any of the other sends of Debugger Events cause a race condition, as they're not within a lock? With a very cursory inspection, I see what may be other potential race conditions, but it depends on what sends the triggering events.

mousetraps:
Potentially. We've mainly been fixing issues as they are reported, and while the issues are not too frequent, it may be worth doing a full code review of the debugger at some point.

Issue #19, opened 2015-04-02
Test discovery is not working when default processor architecture set to x64


Ruslan-B:
In Visual Studio 2013. (node.js x64) 
Tests cannot be discovered or displayed in grey after this settings is applied:
Test->Test Settings -> Default Processor Architecture -> x64

lukedgr:
I was facing a similar issue, where I had another test adapter that was causing the vstest.discoveryengine.exe to run (instead of vstest.discoveryengine.x86.exe) and this resulted in the tests not being discovered.  However, I found the following work around. 

First, download the source code, and recompile the TestAdapter project with 'Any CPU' as the target platform. Then, replace the test adapter dll in the install directory with the dll that was generated from this compilation and re-open Visual Studio.

mousetraps:
@lukedgr Indeed that works! Just verified we don't have any 32-bit-required dependencies in the project and submitted a PR.

mousetraps:
Fixed in #111

mousetraps:
Available in the [latest dev build](https://github.com/Microsoft/nodejstools/releases)

toddpi314:
NTVS 1.0 - Mar 25, 2015 - Where not seeing any tests get picked up, regardless of configuration, on VS 2015 RC.
5-13-2015 Dev Build - Confirmed resolved for all architectures.

Thanks!
Todd

PR #111, opened 2015-05-09
#19 Test discovery is not working when default processor architecture is

mousetraps:
set to x64
- TestAdapter project was targeting x86, but none of its assembly dependencies
  require 32-bit, so there is no reason that it cannot target AnyCPU. This
  enables it to run vstest.executionengine.exe instead of
  vstest.executionengine.x86.exe


AndreSayreMSFT:
Do any of our assemblies/dependencies require x86, or can we holistically switch things to AnyCPU?

mousetraps:
I know some do because that's why we moved most of our assemblies to x86 in the first place, but I'm not sure which ones require it. We can create another issue to verify other assemblies (though I would say this falls more into the category of fixing up our public api), but this is the only spot where things are actually broken because of it.

Ruslan-B:
Thanks!!!

PR #122, opened 2015-05-14
Eliminated periods from template headers

### LINKED ISSUE #33: Should not end template headers and subheaders with periods.
It's pretty weird and not a good default. 
image

paladique:
Fix #33. Found 6 templates similar to the screenshot listed in issue.


paladique:
This isn't ready yet. I missed the subheaders!

PR #123, opened 2015-05-14
Fix #33 - Eliminated periods from template files.

### LINKED ISSUE #33: Should not end template headers and subheaders with periods.

paladique:
Found 6 templates similar to the screenshot listed in issue #33.


mousetraps:
:+1: Thanks for your contribution! The headers with periods always looked off :smiley:

FYI in the future, you can simply update a pull request by pushing a new commit to your branch. For instance, my second commit to #111 was after I had already created the pull request.

Issue #121, opened 2015-05-14
Update PostInstallation fwlink to point to GitHub docs


mousetraps:
http://go.microsoft.com/fwlink/?LinkID=519208

mousetraps:
updated the fwlink to point to https://github.com/Microsoft/nodejstools/wiki/Install-Node.js-and-get-started-with-NTVS

PR #129, opened 2015-05-19
Update Post-Install link

paladique:
Link needs to point to GitHub docs Per #121


mousetraps:
Thanks for the pull request! Unfortunately we can't take it because we don't actually have to update the code to make this one happen (just the fwlink, which is not changeable by the external community.) Updating the fwlink rather than the absolute link means that we don't have to worry about updating the product when links change. I went ahead and took care of this one, so http://go.microsoft.com/fwlink/?LinkID=519208 should now point to the right location.

Perhaps we should not mark issues like this one as up-for-grabs since it's not up-for-grabs for *everyone* (only those on the core team.)

Issue #88, opened 2015-04-30
.ntvs_analysis.dat size and mem use


objelke:
Node: v 0.12.2
VS 2012 and 2013, both Ultimate
NTVS 1.0

On larger projects, memory use becomes a significant issue. I suspect it's related to the analyze function; the .ntvs_analysis.dat files for individual projects range from a few MB to >400MB.

The advise for using AnalysisIgnoredDirectories to exclude modules has not resulted in any measurable improvements. Also the .dat file sizes stayed const - even after playing with reference pathes etc.
(note: every project reload after a change takes ~ 10 minutes until the new .dat is completed so hard to test many scenarios).

Combining multiple projects in a single solution had to be given up at an early stage already (sadly) - the VS process was quickly approaching 4GB mem use and then often crashing. If it survived, VS became unusable (many seconds between possible keystrokes, often pausing 20sec+ before VS could be used again).

But even when just loading individual projects, load time is a drag - on slower/older laptops, 10 min.+ for a single project. Even on fast machines (SSD/16GB RAM etc.) many minutes.

Disabling intellisense may be an option - but that would invalidate the use of VS in the first place for this.

Once it's all loaded, editing performance is actually quite all right. 
Launching the debugger takes a bit of time but survivable. 
While debugging though, every keystroke is followed by painfully slow execution.

The projects are admittedly complex, a typical dependency list:

"dependencies": {
    "bcrypt-nodejs": "0.0.3",
    "bluebird": "^2.9.14",
    "chai": "^2.1.2",
    "chai-as-promised": "^4.3.0",
    "expiring-lru-cache": "^2.1.0",
    "jsonschema": "^1.0.1",
    "knex": "^0.7.6",
    "lodash": "^3.6.0",
    "moment": "^2.9.0",
    "monitor": "^0.6.10",
    "mysql": "^2.5.5",
    "node-cache": "^1.1.0",
    "node-uuid": "^1.4.3",
    "nodemailer": "^1.3.2",
    "nodemailer-smtp-transport": "^1.0.2",
    "passport": "^0.2.1",
    "passport-http": "^0.2.2",
    "remove": "^0.1.5",
    "restify": "^3.0.1",
    "simple-lru-cache": "0.0.1",
    "when": "^3.7.2"
  },

(needles to say every dependency has a significant node_modules subfolder itself and so forth...)

Looking for advise 
- any useful action which could be taken without giving up on intellisense?

[...]

PR #146, opened 2015-05-28
Fix #138 Add medium level of IntelliSense / limit to the analysis depth (Related to #88).

### LINKED ISSUE #138: Medium-power IntelliSense mode
This would be helpful for users who are more CPU-bound, but still want some degree of IntelliSense. Right now the only options are on/off.

kant2002:
Limit depth analysys to relative paths which contains only 2 node_modules. This should be sufficient enough to capture
a) Capture API of the root package
b) Capture API of the dependencies. This is obiviously the must to analyze (1 node_modules)
c) Capture API of subdependencies (2 node_modules). This is also the must, since a lot of packages are wrapper/glue for another library. For example using gulp-typescript require that gulp-typescript should be parsed (level b) and typescript should be parsed (level c) since typescritp provide some config options for gulp-typescript.

Related to #88 

#138 Added Medium level of Intellisense. This level now same as Full level, but limit analysis depth to the 2 nested modules depth. Full mode is now has limit to 4 modules depth which I almost sure practical enough, but could be increased if needed. Low level is analyse 1 level depth.

### DIFF
--- a/Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
+++ b/Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
@@ -34,7 +34,7 @@ namespace Microsoft.NodejsTools.Analysis {
             MaxObjectLiteralProperties = 50;
             MaxObjectKeysTypes = 5;
             MaxMergeTypes = 5;
-            NestedModulesLimit = 4;
+            NestedModulesLimit = AnalysisConstants.MaxAnalysisDepthQualty;
         }
 
         /// <summary>
@@ -43,7 +43,7 @@ namespace Microsoft.NodejsTools.Analysis {
         /// <returns>An <see cref="AnalysisLimits"/> object representing medium level Initellisense settings.</returns>
         public static AnalysisLimits MakeMediumAnalysisLimits() {
             return new AnalysisLimits() {
-                NestedModulesLimit = 2
+                NestedModulesLimit = AnalysisConstants.MaxAnalysisDepthFast
             };
         }
 
--- a/Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
+++ b/Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
@@ -58,7 +58,6 @@ namespace Microsoft.NodejsTools.Analysis {
             //
             // All these examples are highly speculative and I specifically try to create such deep level. 
             // If you develop on windows with such deep level you already close to your limit, your maximum is probably 10. 
-            // </remarks>
             NestedModulesLimit = 4;
         }
 
--- a/Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
+++ b/Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
@@ -34,6 +34,17 @@ namespace Microsoft.NodejsTools.Analysis {
             MaxObjectLiteralProperties = 50;
             MaxObjectKeysTypes = 5;
             MaxMergeTypes = 5;
+            NestedModulesLimit = 4;
+        }
+
+        /// <summary>
+        /// Creates instance of the <see cref="AnalysisLimits"/> for medium level of Intellisense support.
+        /// </summary>
+        /// <returns>An <see cref="AnalysisLimits"/> object representing medium level Initellisense settings.</returns>
+        public static AnalysisLimits MakeMediumAnalysisLimits() {
+            return new AnalysisLimits() {
+                NestedModulesLimit = 2
+            };
         }
 
         public static AnalysisLimits MakeLowAnalysisLimits() {
@@ -43,7 +54,8 @@ namespace Microsoft.NodejsTools.Analysis {
                 DictKeyTypes = 1,
                 DictValueTypes = 1,
                 IndexTypes = 1,
-                InstanceMembers = 1
+                InstanceMembers = 1,
+                NestedModulesLimit = 1
             };
         }
 
@@ -111,6 +123,11 @@ namespace Microsoft.NodejsTools.Analysis {
         /// </summary>
         public int MaxMergeTypes { get; set; }
 
+        /// <summary>
+        /// Gets the maximum levl  of dependency modules which could be analyzed.
+        /// </summary>
+        public int NestedModulesLimit { get; set; }
+
         public override bool Equals(object obj) {
             AnalysisLimits other = obj as AnalysisLimits;
             if (other != null) {
@@ -125,7 +142,8 @@ namespace Microsoft.NodejsTools.Analysis {
                     other.MaxArrayLiterals == MaxArrayLiterals &&
                     other.MaxObjectLiteralProperties == MaxObjectLiteralProperties &&
                     other.MaxObjectKeysTypes == MaxObjectKeysTypes &&
-                    other.MaxMergeTypes == MaxMergeTypes;
+                    other.MaxMergeTypes == MaxMergeTypes &&
+                    other.NestedModulesLimit == NestedModulesLimit;
             }
             return false;
         }
@@ -142,7 +160,8 @@ namespace Microsoft.NodejsTools.Analysis {
                 MaxArrayLiterals +
                 MaxObjectLiteralProperties +
                 MaxObjectKeysTypes +
-                MaxMergeTypes;
+                MaxMergeTypes +
+                NestedModulesLimit;
         }
     }
 }
--- a/Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
+++ b/Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
@@ -151,9 +151,31 @@ namespace Microsoft.NodejsTools.Analysis {
         public int MaxMergeTypes { get; set; }
 
         /// <summary>
-        /// Gets the maximum levl  of dependency modules which could be analyzed.
+        /// Gets the maximum level of dependency modules which could be analyzed.
         /// </summary>
         public int NestedModulesLimit { get; set; }
+
+        /// <summary>
+        /// Checks whether relative path exceed the nested module limit.
+        /// </summary>
+        /// <param name="path">Path to module file which has to be checked for depth limit.</param>
+        /// <returns>True if path too deep in nesting tree; false overwise.</returns>
+        public bool IsPathExceedNestingLimit(string path) { 
+            int nestedModulesCount = 0;
+            int startIndex = 0;
+            int index = path.IndexOf(AnalysisConstants.NodeModulesFolder, startIndex, StringComparison.OrdinalIgnoreCase);
+            while (index != -1) {
+                nestedModulesCount++;
+                if (nestedModulesCount > this.NestedModulesLimit){
+                    return true;
+                }
+
+                startIndex = index + AnalysisConstants.NodeModulesFolder.Length;
+                index = path.IndexOf(AnalysisConstants.NodeModulesFolder, startIndex, StringComparison.OrdinalIgnoreCase);
+            }
+
+            return false;
+        }
 
         public override bool Equals(object obj) {
             AnalysisLimits other = obj as AnalysisLimits;
--- a/Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
+++ b/Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
@@ -34,7 +34,32 @@ namespace Microsoft.NodejsTools.Analysis {
             MaxObjectLiteralProperties = 50;
             MaxObjectKeysTypes = 5;
             MaxMergeTypes = 5;
-            NestedModulesLimit = AnalysisConstants.MaxAnalysisDepthQualty;
+
+            // There no practical reasons to go deeper in dependencies analysis.
+            // Number 4 is very practical. Here the examples which hightlight idea.
+            // 
+            // Example 1
+            // 
+            // Level 0 - Large system. Some code should use properties of the object on level 4 here. 
+            // Level 1 - Subsystem - pass data from level 4
+            // Level 2 - Internal dependency for company - Do some work and pass data to level 1
+            // Level 3 - Framework on top of which created Internal dependency.
+            // Level 4 - Dependency of the framework. Objects from here still should be available.
+            // Level 5 - Dependency of the framework provide some usefull primitive which would be used very often on the level of whole system. Hmmm.
+            // 
+            // Example 2 (reason why I could increase to 5)
+            // 
+            // Level 0 - Large system. Some code should use properties of the object on level 4 here. 
+            // Level 1 - Subsystem - pass data from level 4
+            // Level 2 - Internal dependency for company - Wrap access to internal library and perform business logic. Do some work and pass data to level 1
+            // Level 3 - Internal library which wrap access to API.
+            // Level 4 - Http library.
+            // Level 5 - Promise Polyfill.
+            //
+            // All these examples are highly speculative and I specifically try to create such deep level. 
+            // If you develop on windows with such deep level you already close to your limit, your maximum is probably 10. 
+            // </remarks>
+            NestedModulesLimit = 4;
         }
 
         /// <summary>
@@ -43,7 +68,9 @@ namespace Microsoft.NodejsTools.Analysis {
         /// <returns>An <see cref="AnalysisLimits"/> object representing medium level Initellisense settings.</returns>
         public static AnalysisLimits MakeMediumAnalysisLimits() {
             return new AnalysisLimits() {
-                NestedModulesLimit = AnalysisConstants.MaxAnalysisDepthFast
+
+                // Maximum practical limit for the dependencies analysis oriented on producing faster results.
+                NestedModulesLimit = 2
             };
         }
 
--- a/Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.cs
+++ b/Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.cs
@@ -42,12 +42,10 @@ namespace Microsoft.NodejsTools.Options {
             get {
                 if (_fullIntelliSenseRadioButton.Checked) {
                     return AnalysisLevel.High;
+                } else if (_mediumIntelliSenseRadioButton.Checked) {
+                    return AnalysisLevel.Medium;
                 } else {
-                    if (_mediumIntelliSenseRadioButton.Checked) {
-                        return AnalysisLevel.Medium;
-                    } else {
-                        return AnalysisLevel.None;
-                    }
+                    return AnalysisLevel.None;
                 }
             }
             set {
--- a/Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.cs
+++ b/Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.cs
@@ -43,13 +43,20 @@ namespace Microsoft.NodejsTools.Options {
                 if (_fullIntelliSenseRadioButton.Checked) {
                     return AnalysisLevel.High;
                 } else {
-                    return AnalysisLevel.None;
+                    if (_mediumIntelliSenseRadioButton.Checked) {
+                        return AnalysisLevel.Medium;
+                    } else {
+                        return AnalysisLevel.None;
+                    }
                 }
             }
             set {
                 switch (value) {
                     case AnalysisLevel.High:
                         _fullIntelliSenseRadioButton.Checked = true;
+                        break;
+                    case AnalysisLevel.Medium:
+                        _mediumIntelliSenseRadioButton.Checked = true;
                         break;
                     case AnalysisLevel.None:
                         _noIntelliSenseRadioButton.Checked = true;
--- a/Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.resx
+++ b/Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.resx
@@ -139,13 +139,13 @@
     <value>NoControl</value>
   </data>
   <data name="_fullIntelliSenseRadioButton.Location" type="System.Drawing.Point, System.Drawing">
-    <value>6, 3</value>
+    <value>8, 4</value>
   </data>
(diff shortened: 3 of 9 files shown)


kant2002:
I add the configuration option for the Middle level of Intellisense (#138). Right now analysis level not configurable, but if you think this would be valuable, I could put additional config option for that.

mousetraps:
I'll take a look at this PR when I've had more sleep :smiley:. But lets make this configurable in the per-project setting as well - it shouldn't override the users general setting, but rather serve as an additional filter. This would enable users who are sharing a project to figure out good project-specific settings rather than relying on every single person on the team to use the same defaults. Because ultimately this is a project-specific perf issue. 

For the sake of UI simplicity, Let's hold off on making it configurable from general project settings unless/until we start getting requests for it.

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.resx
```diff
@@ -322,6 +322,51 @@
...
+    <value>8, 33</value>
+  </data>
+  <data name="_mediumIntelliSenseRadioButton.Margin" type="System.Windows.Forms.Padding, System.Windows.Forms">
+    <value>8, 4, 8, 4</value>
+  </data>
+  <data name="_mediumIntelliSenseRadioButton.Size" type="System.Drawing.Size, System.Drawing">
+    <value>154, 21</value>
+  </data>
+  <data name="_mediumIntelliSenseRadioButton.TabIndex" type="System.Int32, mscorlib">
+    <value>5</value>
```
mousetraps: Check the tab index here. WinForms is annoying in that it won't automatically update all the controls on the page.

Nodejs/Tests/AnalysisTests/AnalysisFile.cs
```diff
@@ -131,6 +131,21 @@ static class Analysis {
             List<AnalysisFile> files = new List<AnalysisFile>();
             foreach (var file in Directory.GetFiles(directory, "*", SearchOption.AllDirectories)) {
                 if (String.Equals(Path.GetExtension(file), ".js", StringComparison.OrdinalIgnoreCase)) {
+                    var relativeFile = file.Substring(directory.Length);
+                    int nestedModulesCount = 0;
+                    int index = relativeFile.IndexOf("node_modules");
+                    while (index != -1)
+                    {
```
mousetraps: brace on same line here and everywhere else

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.cs
```diff
@@ -43,7 +43,11 @@ public partial class NodejsIntellisenseOptionsControl : UserControl {
                 if (_fullIntelliSenseRadioButton.Checked) {
                     return AnalysisLevel.High;
                 } else {
-                    return AnalysisLevel.None;
+                    if (_mediumIntelliSenseRadioButton.Checked) {
```
mousetraps: this if/else block doesn't need to be nested

Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
```diff
@@ -34,6 +34,17 @@ internal class AnalysisLimits {
             MaxObjectLiteralProperties = 50;
             MaxObjectKeysTypes = 5;
             MaxMergeTypes = 5;
+            NestedModulesLimit = 4;
```
mousetraps: Do we really want to mess with the high analysis limits here? How did you decide on 4?

Nodejs/Tests/AnalysisTests/AnalysisFile.cs
```diff
@@ -131,6 +131,21 @@ static class Analysis {
             List<AnalysisFile> files = new List<AnalysisFile>();
             foreach (var file in Directory.GetFiles(directory, "*", SearchOption.AllDirectories)) {
                 if (String.Equals(Path.GetExtension(file), ".js", StringComparison.OrdinalIgnoreCase)) {
+                    var relativeFile = file.Substring(directory.Length);
+                    int nestedModulesCount = 0;
+                    int index = relativeFile.IndexOf("node_modules");
```
mousetraps: use StringComparison.OrdinalIgnoreCase for this. Also we have this in a resource string, so may as well use that (NodejsConstants.NodeModulesFolder)

Nodejs/Tests/AnalysisTests/AnalysisFile.cs
```diff
@@ -131,6 +131,21 @@ static class Analysis {
...
+                    while (index != -1)
+                    {
+                        nestedModulesCount++;
+                        relativeFile = relativeFile.Substring(index + 1);
```
mousetraps: instead of creating a whole new string, you can change the search position in IndexOf

Nodejs/Tests/AnalysisTests/AnalysisFile.cs
```diff
@@ -131,6 +131,21 @@ static class Analysis {
...
+                        index = relativeFile.IndexOf("node_modules");
+                    }
+
+                    if (nestedModulesCount > limits.NestedModulesLimit)
```
mousetraps: You should be able figure this out before you've traversed the entire string.

Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
kant2002: Reason for 4 following - I try to be very practical about numbers. 

For small applications this number more then enough, and they don't have any issues dealing with some potential errors in Intellisense.

But if we have large application with Full setting, then VS Tools for Node.JS as a product immediately became unusable for that users.

I try to imagine dependency tree which exceed this 4 depth limit. Here the examples of trees about which I thought:
## Example 1
Level 0 - Large system. Some code should use properties of the object on level 4 here. 
Level 1 - Subsystem - pass data from level 4
Level 2 - Internal dependency for company - Do some work and pass data to level 1
Level 3 - Framework on top of which created Internal dependency.
Level 4 - Dependency of the framework. Objects from here still should be available.
Level 5 - Dependency of the framework provide some usefull primitive which would be used very often on the level of whole system. Hmmm. 

## Example 2 (reason why I could increase to 5)
Level 0 - Large system. Some code should use properties of the object on level 4 here. 
Level 1 - Subsystem - pass data from level 4
Level 2 - Internal dependency for company - Wrap access to internal library and perform business logic. Do some work and pass data to level 1
Level 3 - Internal library which wrap access to API.
Level 4 - Http library.
Level 5 - Promise Polyfill. 

All these examples are highly speculative and I specifically try to create such deep level. If you develop on windows with such deep level you already close to your limit, your maximum is probably 10. 
If you have such deep structure already, chances that you have more then one such deep dependency is very high, and thus we just increase time for loading solution, loading analysis file and increase memory consumption without any practical benefits.

From my point of view practical benefits much better, and better have good and usable product with 90% working intellisense.
Or we just have 2 options: Medium and None without Full being usable by anyone on the large projects.

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.cs
kant2002: Should I create something like this?

```
if (_fullIntelliSenseRadioButton.Checked) {
    return AnalysisLevel.High;
}

if (_mediumIntelliSenseRadioButton.Checked) {
    return AnalysisLevel.Medium;
} else {
    return AnalysisLevel.None;
}
```

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.resx
```diff
@@ -322,6 +322,51 @@
...
+  </data>
+  <data name="_mediumIntelliSenseRadioButton.Text" xml:space="preserve">
+    <value>&amp;Medium IntelliSense</value>
+  </data>
+  <data name="_mediumIntelliSenseRadioButton.TextAlign" type="System.Drawing.ContentAlignment, System.Drawing">
+    <value>TopLeft</value>
+  </data>
+  <data name="_mediumIntelliSenseRadioButton.ToolTip" xml:space="preserve">
+    <value>Medium completions (less memory/CPU usage)</value>
+  </data>
```
mousetraps: Minor language suggestions, since I'm just now realizing that our label for "high" isn't actually "high", and we don't have a "low", so "medium" doesn't work as well. 
* Name: Quick IntelliSense
* Tooltip: Limited completions (reduced memory/CPU usage)

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.cs
mousetraps: Yeah, and you can even go one step further with if, else if, else for best readability.

mousetraps:
Left some comments, minor tweaks for the most part. @DinoV, it would be great to get your eyes on this as well.

Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
mousetraps: Awesome, thanks for the detailed explanation. Logically it makes sense to me, though Dino has the  most familiarity on this and what the edge cases might be, so it'll be great if he can chime in here too.

Nodejs/Tests/AnalysisTests/AnalysisFile.cs
kant2002: If I push another commit with changes, would our discussion would be preserved?

mousetraps: Yep, and that's preferred so we can have the full discussion in one place. you won't see it inline with the current code, but the old diff will still be available by each comment.

Nodejs/Tests/AnalysisTests/AnalysisFile.cs
```diff
@@ -131,6 +131,20 @@ static class Analysis {
...
             foreach (var file in Directory.GetFiles(directory, "*", SearchOption.AllDirectories)) {
                 if (String.Equals(Path.GetExtension(file), ".js", StringComparison.OrdinalIgnoreCase)) {
+                    var relativeFile = file.Substring(directory.Length);
+                    int nestedModulesCount = 0;
+                    int startIndex = 0;
+                    int index = relativeFile.IndexOf("node_modules", startIndex, StringComparison.OrdinalIgnoreCase);
+                    while (index != -1) {
+                        nestedModulesCount++;
+                        if (nestedModulesCount > limits.NestedModulesLimit) {
+                            continue;
```
mousetraps: the check is nested now, so continue is going to continue the while loop, not the outer for loop, so the while loop just became infinite after the limit is exceeded. 

One solution is the refactor the while loop out into its own method, say "NestedModulesExceedsLimit", and return true instead of continuing, false otherwise. Then...
```csharp
if (!NestedModulesExceedsLimit(...) { 
    files.Add(new AnalysisFile(file, File.ReadAllText(file)));
}
```

Nodejs/Tests/AnalysisTests/AnalysisFile.cs
```diff
@@ -131,6 +131,20 @@ static class Analysis {
             List<AnalysisFile> files = new List<AnalysisFile>();
             foreach (var file in Directory.GetFiles(directory, "*", SearchOption.AllDirectories)) {
                 if (String.Equals(Path.GetExtension(file), ".js", StringComparison.OrdinalIgnoreCase)) {
+                    var relativeFile = file.Substring(directory.Length);
```
mousetraps: Because you are using the position of the string, you can get rid of this substring completely, and just use file instead of relativeFile.

Nodejs/Tests/AnalysisTests/AnalysisFile.cs
```diff
@@ -131,6 +131,20 @@ static class Analysis {
...
+                    int nestedModulesCount = 0;
+                    int startIndex = 0;
+                    int index = relativeFile.IndexOf("node_modules", startIndex, StringComparison.OrdinalIgnoreCase);
+                    while (index != -1) {
+                        nestedModulesCount++;
+                        if (nestedModulesCount > limits.NestedModulesLimit) {
+                            continue;
+                        }
+
+                        startIndex = index + 1;
```
mousetraps: this can be ```index + "node_modules".Length```

Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
DinoV: I'm not seeing the change that actually limits this in the product?  AnalysisTests\AnalysisFile will limit it when we run the tests, but won't do anything when running inside of VS.

Personally I would suggest not putting this in AnalysisLimits, and not having the analyzer actually be aware of this at all.  The AnalysisLimits class is really used for limits on things when we're analyzing.  But in this case we actually just don't want to analyze these files at all.  Which means we also want to avoid parsing them entirely too.  

Therefore I think the correct spot to put this limit in would be in NodejsFileNode.ShouldAnalyze.  If you return false from there then we won't even bother parsing the file which should give the best performance gain.

mousetraps:
Hey there, just checking in to see if you have any questions. It would be great to get this change into the next release. :smiley:

kant2002:
Hi, was side-tracked on the main work. Will take a look at the comments from @DinoV later today, or tomorrow

mousetraps:
Yeah, no worries - it's totally okay to have a life outside NTVS :wink:. This feature is super impactful, though, so unlike cookie dough... the more bake time, the better.

objelke:
>> This feature is super impactful, though

oh yes, we have a good amount of developers waiting... (the homebrew patch is fine but it would be so much nicer to get rid of it)
If I can assist, most happy to do so. Let me know.

kant2002:
Thanks all for waiting, I send additional changes which should address @DinoV comment that I don't place fix in the correct place.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -490,6 +490,26 @@ public NodejsProjectNode(NodejsProjectPackage package)
...
+            int nestedModulesCount = 0;
+            int startIndex = 0;
+            int index = relativeFile.IndexOf(AnalysisConstants.NodeModulesFolder, startIndex, StringComparison.OrdinalIgnoreCase);
+            while (index != -1) {
+                nestedModulesCount++;
+                if (nestedModulesCount > maxModulesDepth){
+                    return false;
+                }
+
+                startIndex = index + 1;
```
mousetraps: this can be startIndex = ```index + "node_modules".Length``` since you already know you've seen node_modules before.

Nodejs/Tests/AnalysisTests/AnalysisFile.cs
```diff
@@ -131,6 +131,20 @@ static class Analysis {
             List<AnalysisFile> files = new List<AnalysisFile>();
             foreach (var file in Directory.GetFiles(directory, "*", SearchOption.AllDirectories)) {
```
mousetraps: You can remove the code from here now, right?

Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
```diff
@@ -34,6 +34,17 @@ internal class AnalysisLimits {
             MaxObjectLiteralProperties = 50;
             MaxObjectKeysTypes = 5;
             MaxMergeTypes = 5;
+            NestedModulesLimit = AnalysisConstants.MaxAnalysisDepthQualty;
```
mousetraps: For the sake of readability and consistency, let's not add this to AnalysisConstants, and instead just include the number. The comment about reasoning is very helpful, though, and should be included where the NestedModulesLimit property is defined.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -490,6 +490,26 @@ public NodejsProjectNode(NodejsProjectPackage package)
             if (CommonUtils.IsSubpathOf(_intermediateOutputPath, fileNode.Url)) {
                 return false;
             }
+
+            int maxModulesDepth = AnalysisConstants.MaxAnalysisDepthQualty;
```
mousetraps: Better yet, use ```Analyzer.Project.Limits.NestedModulesLimit``` to get the current limit.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -490,6 +490,26 @@ public NodejsProjectNode(NodejsProjectPackage package)
             if (CommonUtils.IsSubpathOf(_intermediateOutputPath, fileNode.Url)) {
                 return false;
             }
+
```
mousetraps: This entire block should be placed below the checks for _maxFileSize and _analysisIgnoredDirs, not above.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -490,6 +490,26 @@ public NodejsProjectNode(NodejsProjectPackage package)
...
+            while (index != -1) {
+                nestedModulesCount++;
+                if (nestedModulesCount > maxModulesDepth){
+                    return false;
+                }
+
+                startIndex = index + 1;
+                index = relativeFile.IndexOf(AnalysisConstants.NodeModulesFolder, startIndex, StringComparison.OrdinalIgnoreCase);
+            }
+
```
mousetraps: This entire block should be placed below the checks for _analysisIgnoredDirs and _maxFileSize, rather than above because those should take precedent.

mousetraps:
Awesome, thanks! Added some more comments, so lgtm after that. DinoV and @objelke, feel free to jump in if you have any more comments.

Nodejs/Tests/AnalysisTests/AnalysisFile.cs
kant2002: I think this code should still be there for testing reason. 
Without that code we could measure just raw JsAnalyzer performance without any limitations, but there should be way to measure analysis speed with depth limit. If needed I could add command line switch to be able distinguish between desired behaviors.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
kant2002: Thanks!! I was looking for something like that.

Nodejs/Tests/AnalysisTests/AnalysisFile.cs
mousetraps: How about a new unit test?

mousetraps: The issue with the test as its written now is that it's not actually testing the code that's in the product itself.

kant2002: From what I understand about this tool, it is purpose is kind of developer tool which run analysis on the common npm modules and verifies how analysis goes on then. This is not actual automated testing, but rather tool for capturing feedback. I maybe wrong about that, but I use it that way, trying figure out what's wrong. Again @DinoV maybe give clearer explanation.

mousetraps: Ahh, I think I understand where you're coming from now...

It's not "automated testing" per se, but it is still important that we are testing the code that is in the product. Otherwise if we break something in the product related to this, we won't figure it out. The analysis limits are getting set from the following location.
https://github.com/Microsoft/nodejstools/blob/master/Nodejs/Tests/AnalysisDriver/AnalysisDriver.cs#L170

So we should provide an option to also run with medium analysis set (instead of the boolean flag we have right now to distinguish between low/high)

mousetraps: Essentially we use it to make sure analysis continues performing as expected as we change things and as the packages themselves change.

DinoV: Really the tool exists to do comparisons between different versions of the analyzer.  It certainly doesn't run anywhere in an automated way because you really need to get your baseline, make the changes, and then select what you want to compare it to.

For keeping the product code and test code in sync here what I'd do is move the file check into a common helper where both NTVS and AnalysisDriver can both call into it (e.g. JsAnalyzer).  Then I'd have them both manage the nested modules limit independently - with AnalysisDriver updated to take it from the command line, and NTVS taking it from user settings somewhere.

kant2002: I try to find place where this could be placed, but did not find any public helper class which accessible from AnalysisDryver.

@DinoV could you show me 
1. in which public helper could I place this code 
2. or alternatively I could add InternalsVisibleTo("AnalysisDriver") to Analyisys assembly.

DinoV: I would add it to AnalysisLimits, that seems like a reasonable spot for it to go

kant2002: Reasonable place. Already done that. @DinoV and @mousetraps give last look, hopefully this is last comments :smile:

Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
```diff
@@ -34,6 +34,44 @@ internal class AnalysisLimits {
...
+            // Level 0 - Large system. Some code should use properties of the object on level 4 here. 
+            // Level 1 - Subsystem - pass data from level 4
+            // Level 2 - Internal dependency for company - Wrap access to internal library and perform business logic. Do some work and pass data to level 1
+            // Level 3 - Internal library which wrap access to API.
+            // Level 4 - Http library.
+            // Level 5 - Promise Polyfill.
+            //
+            // All these examples are highly speculative and I specifically try to create such deep level. 
+            // If you develop on windows with such deep level you already close to your limit, your maximum is probably 10. 
+            // </remarks>
```
mousetraps: nit: there's an end tag, but no opening tag here

mousetraps:
Awesome, looks great! super nitpicky comment, but overall :+1:

Nodejs/Product/Analysis/Analysis/AnalysisLimits.cs
kant2002: Just remove it.

Issue #79, opened 2015-04-24
VS 2015 RC: Unit test debugging broken


jinglou-:
- VS 2015 22822.01
- Node.js 0.12.2
- NTVS 30324.01

1. Create an NTVS app
2. Add JavaScript Unit test
3. Set a breakpoint in the test
4. In Test Explorer, select to debug the test
5. Breakpoint not hit

PTVS also has the same issue: https://github.com/Microsoft/PTVS/issues/23
This is a very recent regression.

toddpi314:
+1 Seeing this issue on VS 2015 RC and x64 Windows 10 RC

ant-fx:
+1 - Same with VS2015 RC Win8.1

mousetraps:
Hey all, fix is available in the [latest dev build](https://github.com/Microsoft/nodejstools/releases)

Issue #140, opened 2015-05-28
Sync common code with PTVS


mousetraps:
We have not done this in over 6 months...

PR #150, opened 2015-05-31
Split out common code into subtree and update with latest changes

mousetraps:
See https://github.com/Microsoft/VisualStudio-SharedProject/pull/1


mousetraps:
(will fix issues #79 and #140)

mousetraps:
To make this a little easier to wade through... the most relevant commit is:
https://github.com/mousetraps/nodejstools/commit/10c50489b852502918d1ae94fd3a8435fd019b15

PR #161, opened 2015-06-02
Fix #124 - include error details in mocha unit test results

### LINKED ISSUE #124: Include error details in mocha unit test results
When an error occurs in a mocha unit test, the details of the error are not displayed in the output page. This is fairly critical information and makes running tests through the IDE unviable.

For example, for the following test:

    describe('Test Suite 1', function() {
        it('Test', function () {
            throw new Error('this should show in test explorer');
        })
    })

the output page looks something like:

screenshot

mousetraps:
Fix #124 - include error details in mocha unit test results
- use the 'tap' error reporter instead of xunit because 'xunit' does not
  print the trace.

### DIFF
--- a/Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
+++ b/Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
@@ -61,10 +61,12 @@ var run_tests = function (testName, testFile, workingFolder, projectFolder) {
         mocha.grep(testName);
     }
     mocha.addFile(testFile);
-    // Choose 'tap' rather 'min'. The reason is when under piped/redirect,
-    // mocha produces undisplayable text to stdout and stderr. Using xunit works fine 
-    // And 'xunit' does not print the stack trace from the test.
+
+    // Choose 'tap' rather than 'min' or 'xunit'. The reason is that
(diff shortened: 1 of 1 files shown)


mousetraps:
We should also consider allowing people to switch use their reporter-of-choice in the future.

Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
```diff
@@ -61,9 +61,10 @@ var run_tests = function (testName, testFile, workingFolder, projectFolder) {
         mocha.grep(testName);
     }
     mocha.addFile(testFile);
-    //Choose 'xunit' rather 'min'. The reason is when under piped/redirect,
-    //mocha produces undisplayable text to stdout and stderr. Using xunit works fine 
-    mocha.reporter('xunit');
+    // Choose 'tap' rather 'min'. The reason is when under piped/redirect,
+    // mocha produces undisplayable text to stdout and stderr. Using xunit works fine
```
gilbertw: Comment doesn't make sense that xunit works fine and xunit does not print.

mousetraps: Thanks! Should make more sense now :smiley:

PR #163, opened 2015-06-02
Fix #162 Copying files in sln explorer does not work after latest updates when PTVS is installed

### LINKED ISSUE #162: Copying files in sln explorer does not work after latest updates when PTVS is installed
Returns error about being unable to cast to IClipboardService (assembly is probably resolving to PythonTools...)

mousetraps:
Fix #162 
IClipboardService should be an abstract class rather than an interface because the CLR doesn't take into account assembly names when generating and interfaces GUID.


DinoV:
LGTM!

PR #168, opened 2015-06-03
Fix #216 and add additional mocha configuration options

### LINKED ISSUE #216: Add additional mocha configuration options
timeout, reporter, etc.

lukedgr:
#216

This change adds the ability to configure mocha by adding a config file to the project located at $(ProjectDirectory)\test\mocha.json.  The relative path under the test directory was chosen because this is where mocha looks for a configuration by default when run from the command line (looks for mocha.opts).  Adding the mocha.json file will result in the mocha tests running with the options specified.  

An example of a valid mocha.json file is: 

{
	"ui": "tdd",
	"timeout": 300000,
	"reporter": "xunit"
}

Also, note that if the config file does not specify a ui, timeout or reporter the default will be chosen automatically.

### DIFF
--- a/Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
+++ b/Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
@@ -1,4 +1,5 @@
 ﻿var fs = require('fs');
+var path = require('path');
 
 // Choose 'tap' rather than 'min' or 'xunit'. The reason is that
 // 'min' produces undisplayable text to stdout and stderr under piped/redirect, 
@@ -41,7 +42,7 @@ var find_tests = function (testFileList, discoverResultFile, projectFolder) {
             getTestList(mocha.suite, testFile);
         } catch (e) {
             //we would like continue discover other files, so swallow, log and continue;
-            logError('An error occurred during mocha discovery:' + e);
+            logError('NTVS_ERROR: An error occurred during mocha test discovery in file: ' + testFile, e);
         }
     });
 
@@ -69,13 +70,18 @@ var run_tests = function (testName, testFile, workingFolder, projectFolder) {
     });
 };
 
-function logError(errorMessage) {
-    console.log("NTVS_ERROR: " + errorMessage);
+function logError(errorMessage, error) {
+    if (typeof error === 'undefined') {
+        console.error("NTVS_ERROR: " + errorMessage);
+    } else {
+        console.error("NTVS_ERROR: " + errorMessage, error);
+    }
 }
 
 function detectMocha(projectFolder) {
     try {
-        var Mocha = new require(projectFolder + '\\node_modules\\mocha');
+        var mochaPath = path.join(projectFolder, 'node_modules', 'mocha');
+        var Mocha = new require(mochaPath);
         return Mocha;
     } catch (ex) {
         logError("Failed to find Mocha package.  Mocha must be installed in the project locally.  Mocha can be installed locally with the npm manager via solution explorer or with \".npm install mocha\" via the Node.js interactive window.");
@@ -95,11 +101,11 @@ function applyMochaOptions(mocha, options) {
             var mochaOpt = mocha[opt];
             var optValue = options[opt];
 
-            if (typeof mochaOpt == 'function') {
+            if (typeof mochaOpt === 'function') {
                 try {
                     mochaOpt.call(mocha, optValue);
                 } catch (e) {
-                    logError("Could not set mocha option '" + opt + "' with value '" + optValue + "' due to error: " + e);
+                    logError("Could not set mocha option '" + opt + "' with value '" + optValue + "' due to error: ", e);
                 }
             }
         }
@@ -109,7 +115,8 @@ function applyMochaOptions(mocha, options) {
 function getMochaOptions(projectFolder) {
     var mochaOptions = defaultMochaOptions;
     try {
-        var options = require(projectFolder + '\\test\\mocha.json');
+        var optionsPath = path.join(projectFolder, 'test', 'mocha.json');
+        var options = require(optionsPath);
         options = options || {};
         for (var opt in options) {
             mochaOptions[opt] = options[opt];
--- a/Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
+++ b/Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
@@ -120,7 +120,7 @@ function getMochaOptions(projectFolder) {
             mochaOptions[opt] = options[opt];
         }
     } catch (ex) {
-        console.log("NTVS: mocha.json options file not found. Using default values.");
+        logError("mocha.json options file not found. Using default values.");
     }
 
     return mochaOptions;
--- a/Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
+++ b/Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
@@ -1,5 +1,10 @@
 ﻿var fs = require('fs');
 
+// Choose 'tap' rather than 'min' or 'xunit'. The reason is that
+// 'min' produces undisplayable text to stdout and stderr under piped/redirect, 
+// and 'xunit' does not print the stack trace from the test.
+var defaultMochaOptions = { ui: 'tdd', reporter: 'tap', timeout: 600000 };
+
 var find_tests = function (testFileList, discoverResultFile, projectFolder) {
     var Mocha = detectMocha(projectFolder);
     if (!Mocha) {
@@ -11,7 +16,7 @@ var find_tests = function (testFileList, discoverResultFile, projectFolder) {
             if (suite.tests && suite.tests.length !== 0) {
                 suite.tests.forEach(function (t, i, testArray) {
                     testList.push({
-                        test: t.fullTitle(),
+                        test: t.title,
                         suite: suite.fullTitle(),
                         file: testFile,
                         line: 0,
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
```diff
@@ -53,33 +57,68 @@ var run_tests = function (testName, testFile, workingFolder, projectFolder) {
...
+function getMochaOptions(projectFolder) {
+    var mochaOptions = defaultMochaOptions;
+    try {
+        var options = require(projectFolder + '\\test\\mocha.json');
+        options = options || {};
+        for (var opt in options) {
+            mochaOptions[opt] = options[opt];
+        }
+    } catch (ex) {
+        console.log("NTVS: mocha.json options file not found. Using default values.");
```
mousetraps: This should be console.error("NTVS_ERROR:", blah), otherwise the test reporter won't pick it up (it's a bit confusing because you see it in debug mode.)

mousetraps:
Updated the comment to point to the right issue (just created a feature for mocha config options)

Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
```diff
@@ -53,33 +57,68 @@ var run_tests = function (testName, testFile, workingFolder, projectFolder) {
...
-    // and 'xunit' does not print the stack trace from the test.
-    mocha.reporter('tap');
-
     mocha.run(function (code) {
         process.exit(code);
     });
 };
 
+function logError(errorMessage) {
+    console.log("NTVS_ERROR: " + errorMessage);
```
mousetraps: console.error

Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
```diff
@@ -53,33 +57,68 @@ var run_tests = function (testName, testFile, workingFolder, projectFolder) {
...
+    return mocha;
+}
+
+function applyMochaOptions(mocha, options) {
+    if (options) {
+        for (var opt in options) {
+            var mochaOpt = mocha[opt];
+            var optValue = options[opt];
+
+            if (typeof mochaOpt == 'function') {
```
mousetraps: ```===``` instead of `==`

Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
```diff
@@ -53,33 +57,68 @@ var run_tests = function (testName, testFile, workingFolder, projectFolder) {
...
+                try {
+                    mochaOpt.call(mocha, optValue);
+                } catch (e) {
+                    logError("Could not set mocha option '" + opt + "' with value '" + optValue + "' due to error: " + e);
```
mousetraps: What do you think of just warning here and fall back to the defaults in this case rather than blocking?

Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
```diff
@@ -53,33 +57,68 @@ var run_tests = function (testName, testFile, workingFolder, projectFolder) {
...
+                }
+            }
+        }
+    }
+}
+
+function getMochaOptions(projectFolder) {
+    var mochaOptions = defaultMochaOptions;
+    try {
+        var options = require(projectFolder + '\\test\\mocha.json');
```
mousetraps: better to use ```path.join``` here

Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
```diff
@@ -53,33 +57,68 @@ var run_tests = function (testName, testFile, workingFolder, projectFolder) {
...
     });
 };
 
+function logError(errorMessage) {
+    console.log("NTVS_ERROR: " + errorMessage);
+}
+
 function detectMocha(projectFolder) {
     try {
         var Mocha = new require(projectFolder + '\\node_modules\\mocha');
```
mousetraps: better to use ```path.join```

mousetraps:
Hey, thanks so much for the contribution! :smiley: I left some more specific notes on the code, but some more general comments below.

I've actually been toying around with this change in my mind, and it would be great to hear your thoughts.

I'm not entirely sure about the configuration discovery aspect (projectFolder\tests), but I also think it's a big step in the right direction :smiley:. So I'm in favor of taking this change, and then adapting it over time as we get feedback on it and understand people's preferences. Ultimately if people like this approach, and they start requesting it for more features, we might want to start including many more settings in a single file (sort of like VSCode's settings.json). But this is a good litmus test.

The other interesting note is that... when we get the chance, we'll want to consider moving to the Common Project System (rather than maintaining VisualStudio-SharedProject, as we do now), and CPS is a more folder/file based system, so we'll probably have to revisit this issue then.- and not just go for mocha tests, but really settings in general.

On the technicalities end...
* legal stuff - your profile says MSFT org, but for some reason it's marked as CLA-required. Not sure what's up there... did you join the github msft org after you submitted the PR? 
* as an FYI, there will be some merge conflicts with recent changes / PRs (we were both working on test stuff at the same time and I didn't get a chance to look through this PR until now). They're pretty minor, and I'll resolve them at merge-time, but worth taking a look at.

Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
lukedgr: To me if this would happen as a result of developer error.  For example, if the settings json file contained this:
{
    "reporter": "InvalidReporter"
}

Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
mousetraps: I saw your comment fly by in email, but I can't see it on github for some reason. Anyways to elaborate on this... as the code is written now, the user will not see it. "NTVS_ERROR" is a string that we search for and strip out of the final output when we display it in the output window. 
``` csharp
IEnumerable<String> stdoutLines = stdout.Split(new string[] {Environment.NewLine},
    StringSplitOptions.RemoveEmptyEntries).Where(s => s.StartsWith("NTVS_ERROR:")).Select(s => s.Trim().Remove(0,11));
```
https://github.com/Microsoft/nodejstools/blob/35817ffdaa1797f95628a22f588fc3ed885f884b/Nodejs/Product/TestAdapter/TestFrameworks/TestFramework.cs#L53-L54

You'll only see the full output printed out when the product is built with the "Debug" configuration, but not the "Release" config. So using ```logError``` is the way to go here.

Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
```diff
@@ -29,15 +35,14 @@ var find_tests = function (testFileList, discoverResultFile, projectFolder) {
...
+        var mocha = initializeMocha(Mocha, projectFolder);
         try {
-            mocha.ui('tdd');
             mocha.addFile(testFile);
             mocha.loadFiles();
             getTestList(mocha.suite, testFile);
         } catch (e) {
             //we would like continue discover other files, so swallow, log and continue;
-            console.error('catch discover error:' + e);
+            logError('NTVS_ERROR: An error occurred during mocha test discovery in file: ' + testFile, e);
```
mousetraps: with logError, there is no need to prepend ````NTVS_ERROR``` here anymore.

Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
```diff
@@ -53,33 +58,74 @@ var run_tests = function (testName, testFile, workingFolder, projectFolder) {
...
-    // 'min' produces undisplayable text to stdout and stderr under piped/redirect, 
-    // and 'xunit' does not print the stack trace from the test.
-    mocha.reporter('tap');
-
     mocha.run(function (code) {
         process.exit(code);
     });
 };
 
+function logError(errorMessage, error) {
```
mousetraps: Looking at this again... it looks this function is just trying to emulate console.error, but prepend an argument. So a more flexible way to do this would be something like:
``` javascript
var errorArgs = Array.prototype.slice.call(arguments).unshift("NTVS_ERROR:");
console.error.apply(errorArgs);
```

Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
```diff
@@ -53,33 +58,74 @@ var run_tests = function (testName, testFile, workingFolder, projectFolder) {
...
+    if (options) {
+        for (var opt in options) {
+            var mochaOpt = mocha[opt];
+            var optValue = options[opt];
+
+            if (typeof mochaOpt === 'function') {
+                try {
+                    mochaOpt.call(mocha, optValue);
+                } catch (e) {
+                    logError("Could not set mocha option '" + opt + "' with value '" + optValue + "' due to error: ", e);
```
mousetraps: You don't need spaces as separator characters when you are appending using console.log arguments, so it would be ```"due to error:", e```, not ```"due to error: ", e``` (which would insert an extra space)

mousetraps:
Thx  for the new iteration. Added some super minor comments, and lgtm after that

lukedgr:
I am happy to help improve NTVS!

... On the mocha options json path, I agree that it is not ideal to have the relative path hard-coded, and I think it would be great if each .njsproj had it's own settings file.  The reason I decided to use projectFolder\test\mocha.json is that when running from the command line mocha will look for options at .\test\mocha.opts (http://mochajs.org/#mocha.opts).

mousetraps:
Ahh, that makes much more sense now, and thanks for the link - I wasn't familiar with mocha.opts

So eventually... what if instead of specifying a separate json file, we could just pick up the options from the mocha.opts? That way it "just works" with an existing setup.

lukedgr:
I see your point of it just working with the existing set up, however the downside to that is you are unable to specify different options for the CL and VS.

For example, my project uses a different reporter on the CL than when running inside VS.

mousetraps:
Gotcha, good call

mousetraps:
:+1: with the new changes btw. I'll merge it in when I get the chance.

landonpoch:
Any plans to support more options (like the --require flag)? Also, it might be good to document which flags don't currently work. I'm tempted to hack around in the mocha.js file to get things working but I don't want to break my upgrade path.

mjbvz:
@landonpoch I don't think we have any plans to do this at the moment, but feel free to look into adding additional options and submit a PR (see #940.)

However, I don't think we want to encourage the use of the `--require` flag. In my understanding, it messes up intellisense since an analyzer cannot know about modules listed in `--require`.

Issue #61, opened 2015-04-11
icon in solution explorer for Typescript files in VS2013 is incorrect


jinglou-:
VS 2015 22803.01
NTVS 30324.01
![typescript_iconissue](https://cloud.githubusercontent.com/assets/10081483/7099403/f1410506-dfa6-11e4-995a-c3cefbd947ee.png)

heroboy:
In my VS2013 solution explorer the .js and .ts files' icons are the same. 
Is it a bug?

mousetraps:
Yeah, it should be different (see this HTML App project)
![image](https://cloud.githubusercontent.com/assets/762848/7245059/db737876-e799-11e4-948f-afdb3c6be261.png)

mousetraps:
Fixed this issue for VS2015 using the new project system improvements, but VS2013 is still broken.

billti:
Fixed for VS2015. Not a major issue for VS2013. Closing.

PR #170, opened 2015-06-05
Start using image monikers in project system, fix typescript file issue, fix misc. build issues

### LINKED ISSUE #169: VS2015 build broken
merge issue, working on fixing it...

mousetraps:
* Start using ImageMonikers in project system instead of image lists
* update common code for back compat with image lists
* fix part of #61 typescript icon incorrect (fixed icon in VS2015)
* Fix #169 (misc. build issues)


mousetraps:
without whitespace changes https://github.com/Microsoft/nodejstools/pull/170/files?w=1

We should also normalize the repo and add a .gitattributes file so this doesn't happen in the future.

mousetraps:
still fixing this up...

PR #172, opened 2015-06-05
Fix exception in classifier

### LINKED ISSUE #130: Exception when scrolling text editor with node.js code
Got a random exception while scrolling the text editor with node.js code. I really can't think of any special circumstances that caused this nor can I reproduce. Perhaps the only non-standard condition is that I'm working in a dual monitor setup where my text editor tabs are in a separate window than the IDE main frame.

Here is my [AcitivityLog.xml](https://www.dropbox.com/s/d75dl04u5dhi8cv/ActivityLog.xml?dl=0)

Running latest build version v1.1.Dev-5.13.2015 (44bd6d6962ab7acb49d1bbe538af1f350edc928f)
VS 2013 Ultimate, win7 x64

jaredpar:
This appears to be the root cause of issue #130.  The use of `<=` allows
for `currentLineTemp` to be equal to the number of lines in the
snapshot.  When that occurs and the value is passed to `TokenizeLine` it
will result in an exception because it will be passed to
`GetLineFromLineNumber` which throws when the line requested is `>=` to
the line count.

closes #130


mousetraps:
:+1:

PR #175, opened 2015-06-05
Ignore templates that are autogenerated

paulvanbrenk:
Add 2 entries to the ignore file for templates that are autogenerated and remove them from the repro, since they're autogenerated on build.


mousetraps:
Lgtm

PR #176, opened 2015-06-05
 Make the various Guids used in the debug engine unique from PTVS  

### LINKED ISSUE #147: Need to use a unique GUID for the DebugEngineAttribute
Product\SharedProjectProject\ProvideDebugEngineAttribute.cs.  The CLSID is the same as PTVS and is colliding in registry.  We need to use a unique GUID for this value.

paulvanbrenk:
Adds the debug guids to the central guid file, and ensures they're unique from the PTVS guids. 

Fixes #147

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ TestResults
 *.suo
 *.user
 *.sln.docstates
+.vs
 
 # Build results
 [Dd]ebug/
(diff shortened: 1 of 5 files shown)


mousetraps:
LGTM. Based on the change to CRLF, I assume you intentionally change the line endings on this one? 

https://github.com/Microsoft/nodejstools/pull/176/files?w=1 without line endings

mousetraps:
Actually, one of the GUIDs is still the same: 
```
A241707C-7DB3-464F-8D3E-F3D33E86AE99
```

paulvanbrenk:
Yes, the RemoteDebugPortSupplier is the same, that wasn't changed in PTVS either. and they don't collide. So I kep that the same.

paulvanbrenk:
And VS changed the line endings for me..

PR #177, opened 2015-06-05
Replaced a bunch of double white lines with single whitelines... and …

paulvanbrenk:
…normalized line ending for the files I touched.


mousetraps:
forgoing this change because it'll make merging a pain

Issue #40, opened 2015-04-07
NPM Install dialog should not appear on Project create


AndreSayreMSFT:
On project creation, we expect npm to automatically installed the requisite packages required by the new template.

Issue #39, opened 2015-04-07
Status bar from Analyzer incorrect with only 1 function


AndreSayreMSFT:
When the analyzer only analyzes one function, the message contains `analyzed 1 functions`.  Should read `function`.

PR #190, opened 2015-06-08
Npm install dialog & one function analysis

paulvanbrenk:
This removes the dialog to confirm the npm package installation after creating a new project as per #40 and fixes up a string when the analyzer only analyses a single function as per #39


mousetraps:
:shipit:

PR #214, opened 2015-06-17
Fix #96 when throwing strings

### LINKED ISSUE #96: Please print the stack trace when an error is encountered when parsing tests
Hi,

When there is an error in a test file NTVS current prints out the following on the "Tests" output window:

```
------ Discover test started ------
Processing: D:\foo\bar\bar.test.js;<plus some 100 file names>;D:\foo\bar\baz.test.js
TypeError: Cannot read property 'foo' of undefined
Processing finished for framework of ExportRunner
Discovered 0 testcases.
========== Discover test finished: 0 found (0:00:02.0461167) ==========
```

From the above, it's impossible to tell which file has the actual error.

Please make this more user-friendly.

* The long list of files is not very useful.
* Print a stack trace with the error.

Thanks.

mousetraps:
Rather than printing the stack, which is not available, print the testFile path


gilbertw:
:+1:

PR #217, opened 2015-06-17
Fix #154 Mocha unit tests should timeout @ 2sec

### LINKED ISSUE #154: Mocha unit tests should time out
Mocha includes a built in test time out of 2 seconds. The integrated test runner should share this behaviour.

mousetraps:
- set default timeout to 2 seconds when the debugger is not attached


gilbertw:
:+1:

PR #232, opened 2015-06-26
Fix #148 Avoid running tf command if it is not tf environment

### LINKED ISSUE #148: BuildRelease.ps1 require tf command to build
BuildRelease.ps1 file requires that `tf` command should be in the path. This is no longer needed since NodeJS tools now on the GitHub instead of CodePlex.

zhengbli:
On the new MicroBuild server there is no TFS environment. Remove the unnecessary commands in the build process.

### DIFF
--- a/Nodejs/Setup/BuildRelease.ps1
+++ b/Nodejs/Setup/BuildRelease.ps1
@@ -325,8 +325,8 @@ if ($name) {
 $version_file_backed_up = 0
 # Force use of a backup if there are pending changes to $version_file
 $version_file_force_backup = 0
-$has_no_tf_workspace = (tf workspaces | Select-String -pattern "No workspace", "Unable to determine the workspace")
-if (-not $has_no_tf_workspace) {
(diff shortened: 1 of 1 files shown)


Nodejs/Setup/BuildRelease.ps1
```diff
@@ -325,9 +325,12 @@ if ($name) {
 $version_file_backed_up = 0
 # Force use of a backup if there are pending changes to $version_file
 $version_file_force_backup = 0
-if (-not (tf status $version_file /format:detailed | Select-String "There are no pending changes.")) {
-    Write-Output "$version_file has pending changes. Using backup instead of tf undo."
-    $version_file_force_backup = 1
+$has_no_tf_workspace = (tf workspaces | Select-String -pattern "No workspace", "Unable to determine the workspace")
```
mousetraps: The double negative is confusing from a readability standpoint, and we include the negation everywhere we use it, so it's better to negate it on assignment, so it can be ```if ($has_tf_workspace)``` everywhere else

mousetraps:
:+1:

Issue #234, opened 2015-06-26
A lot of System.IO.PathTooLongException when hitting MAX_PATH limit


kant2002:
When opening package has large dependency tree, the system produce a ton of `System.IO.PathTooLongException` (more then 100 definitely) in the log. Project opens just fine, bit CPU is spiking during opening.

Instead of just concatenating paths opening files, files should be opened when their path does not exceed MAX_PATH constant. Since we know that we could not open file before hand, then no need to wait for `System.IO.PathTooLongException`

PR #236, opened 2015-06-26
Avoid PathTooLongExceptions when possible

kant2002:
When opening NPM modules with large depth tree, a lot of exceptions silently produced. This lead to CPU spikes as described in #234

### DIFF
--- a/Common/Product/SharedProject/CommonProjectNode.cs
+++ b/Common/Product/SharedProject/CommonProjectNode.cs
@@ -819,8 +819,8 @@ namespace Microsoft.VisualStudioTools.Project {
                 return true;
             }
 
-            if ((path.Length >= 248 || !Directory.Exists(path) 
-                && (path.Length >= 260 || !File.Exists(path)))) {
+            if ((path.Length >= NativeMethods.MAX_FOLDER_PATH || !Directory.Exists(path) 
+                && (path.Length >= NativeMethods.MAX_PATH || !File.Exists(path)))) {
                 // if the file has disappeared avoid the exception...
                 // same if file is outside of path limits.
                 return true; // Files/directories that don't exist should be hidden. This also fix DiskMerger when adds files that were already deleted
--- a/Common/Product/SharedProject/CommonProjectNode.cs
+++ b/Common/Product/SharedProject/CommonProjectNode.cs
@@ -819,8 +819,10 @@ namespace Microsoft.VisualStudioTools.Project {
                 return true;
             }
 
-            if (!File.Exists(path) && !Directory.Exists(path)) {
+            if ((path.Length >= 248 || !Directory.Exists(path) 
+                && (path.Length >= 260 || !File.Exists(path)))) {
                 // if the file has disappeared avoid the exception...
+                // same if file is outside of path limits.
                 return true; // Files/directories that don't exist should be hidden. This also fix DiskMerger when adds files that were already deleted
             }
 
--- a/Common/Tests/Utilities/FileUtils.cs
+++ b/Common/Tests/Utilities/FileUtils.cs
@@ -23,16 +23,6 @@ using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace TestUtilities {
     public static class FileUtils {
-        /// <summary>
-        /// Maximum directory length in Windows.
-        /// </summary>
-        internal const int MaxDirectoryLength = 248;
-
-        /// <summary>
-        /// Maximum file name length in Windows.
-        /// </summary>
-        internal const int MaxFileNameLength = 260;
-
         public static void CopyDirectory(string sourceDir, string destDir) {
             sourceDir = sourceDir.TrimEnd('\\');
             destDir = destDir.TrimEnd('\\');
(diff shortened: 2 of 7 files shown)


kant2002:
Please @mousetraps look at this too.

kant2002:
Please take a look at this, it is really helps during debugging. With that improvements, I could run VS with NodeJS tools under debugger and take a look at the errors immediately. Right now it is very slow to launch any sizable project in second VS

Common/Product/SharedProject/CommonProjectNode.cs
```diff
@@ -819,8 +819,10 @@ class DirState {
                 return true;
             }
 
-            if (!File.Exists(path) && !Directory.Exists(path)) {
+            if ((path.Length >= 248 || !Directory.Exists(path) 
+                && (path.Length >= 260 || !File.Exists(path)))) {
```
mousetraps: You can use NativeMethods.MAX_PATH and NativeMethods.MAX_FOLDER_PATH here

Common/Tests/Utilities/FileUtils.cs
```diff
@@ -23,6 +23,16 @@
...
     public static class FileUtils {
+        /// <summary>
+        /// Maximum directory length in Windows.
+        /// </summary>
+        internal const int MaxDirectoryLength = 248;
+
+        /// <summary>
+        /// Maximum file name length in Windows.
+        /// </summary>
+        internal const int MaxFileNameLength = 260;
```
mousetraps: NativeMethods here too

Nodejs/Product/Npm/SPI/NodeModules.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.NodejsTools.Npm.SPI {
     internal class NodeModules : AbstractNodeModules {
         public NodeModules(IRootPackage parent, bool showMissingDevOptionalSubPackages) {
             var modulesBase = Path.Combine(parent.Path, "node_modules");
-            if (Directory.Exists(modulesBase)) {
+            if (modulesBase.Length <= 224 && Directory.Exists(modulesBase)) {
```
mousetraps: Why 224?

Nodejs/Product/Npm/SPI/NodeModules.cs
```diff
@@ -34,12 +34,15 @@ internal class NodeModules : AbstractNodeModules {
...
                     Package module = null;
                     if (!Contains(dependency.Name)) {
-                        module = new Package(
-                            parent,
-                            Path.Combine(modulesBase, dependency.Name),
-                            showMissingDevOptionalSubPackages);
-                        if (parent as IPackage == null || !module.IsMissing || showMissingDevOptionalSubPackages) {
-                            AddModule(module);
+                        var dependencyPath = Path.Combine(modulesBase, dependency.Name);
+                        if (dependencyPath.Length <= 224) {
```
mousetraps: This check should be moved to to RootPackage instead for maximum impact

Common/Tests/Utilities/FileUtils.cs
```diff
@@ -57,8 +69,13 @@ where f.StartsWith(destDir)
...
-                    File.SetAttributes(copyTo, FileAttributes.Normal);
+                    if (copyTo.Length < MaxFileNameLength && copyFrom.Length < MaxFileNameLength) {
+                        var copyToDir = Path.GetDirectoryName(copyTo);
+                        if (copyToDir.Length < MaxDirectoryLength) {
+                            File.Copy(copyFrom, copyTo);
+                            File.SetAttributes(copyTo, FileAttributes.Normal);
+                        }
+                    }
                 } catch {
                     Debug.WriteLine("Failed to copy " + copyFrom + " to " + copyTo);
```
mousetraps: We should still print the debug message (here and everywhere else we used to print) if we hit max_path since the exception is longer getting caught

mousetraps:
Thanks for the PR! Left a few comments.

Sorry for the delay, things have been crazy on our end (aiming for a release on July 20 in line with VS RTM).

Nodejs/Product/Npm/SPI/NodeModules.cs
kant2002: 224 because it is 248 - length of "node_modules\.bin" (?) and some room for the nested dependency name. If I would more precise, I have to refactor much more and I under assumption that if you on Windows and hit MAX_PATH limit you are already in world of pain, and if I miss some edge package, nothing wrong happens. I believe that easier to make this work 90% of time, and then workout edge cases for others, since this is really helpful during debugging.

Nodejs/Product/Npm/SPI/NodeModules.cs
kant2002: Right now I recall why I do that, I probably don't want to dig deeper in the Package put checks in this RootPackage, since I not 100% understand how this works. But if needed i could try.

mousetraps: Yeah, digging into it would be best. For instance, right now this fix won't cover the case where you have a global package exceeding max_path.

Essentially you'l just want to perform the check on fullPathToRootDirectory

Nodejs/Product/Npm/SPI/NodeModules.cs
mousetraps: Don't you also do the check below though with dependencyPath.Length < 224?

Nodejs/Product/Npm/SPI/NodeModules.cs
mousetraps: Actually, nevermind, just looked at it again - this is the correct spot for it.

kant2002:
@mousetraps This should be very close to finish

Issue #230, opened 2015-06-23
Can't debug unit tests with quotes in the name


beautifulcoder:
This one had me spinning wheels until I realized it was a bug somewhere. While hacking around the [Express](https://github.com/strongloop/express) code base, I encountered these unit tests:

    it('should work "view engine" with leading "."', function(done){
      var app = express();

      app.set('views', __dirname + '/fixtures');
      app.engine('.html', render);
      app.set('view engine', '.html');
      app.locals.user = { name: 'tobi' };

      app.render('user', function(err, str){
        if (err) return done(err);
        str.should.equal('<p>tobi</p>');
        done();
      })
    })

Test explorer seems to work when I right-click "Run Selected Test." But, when try to step into the test, Visual Studio freezes or just ignores my break point. There are no errors and nothing tells me something went wrong. The only thing I see is this in the Output > Debug window:

    The thread 'main thread' (0x1) has exited with code 0 (0x0).
    The program '[1] node.exe @ tcp://localhost:52589/#ping=0' has exited with code -1 (0xffffffff).

What is even more bizarre, when I change the unit test to this:

      it('should work view engine with leading .', function(done){

Or this:

      it("should work 'view engine' with leading '.'", function(done){

Then everything magically works. :\

mousetraps:
Wow, good job on narrowing that one down! This issue most likely occurs with the way we are appending arguments and wrapping with quotes when we pass it into the process. 
https://github.com/Microsoft/nodejstools/blob/35817ffdaa1797f95628a22f588fc3ed885f884b/Nodejs/Product/TestAdapter/TestExecutor.cs#L172-L187

mousetraps:
Available in the [latest dev build](https://github.com/Microsoft/nodejstools/releases)

PR #237, opened 2015-06-26
Add escaping of the test names for passing as command line parameters

kant2002:
See #230 for details.

### DIFF
--- a/Nodejs/Product/TestAdapter/TestFrameworks/TestFramework.cs
+++ b/Nodejs/Product/TestAdapter/TestFrameworks/TestFramework.cs
@@ -95,34 +95,38 @@ namespace Microsoft.NodejsTools.TestAdapter.TestFrameworks {
             workingDirectory = workingDirectory.TrimEnd(new char['\\']);
             projectRootDir = projectRootDir.TrimEnd(new char['\\']);
             return new string[] {
-                WrapWithQuot(_runTestsScriptFile),
+                WrapWithQuotes(_runTestsScriptFile),
                 Name,
-                WrapTestNameWithQuot(testName),
-                WrapWithQuot(testFile),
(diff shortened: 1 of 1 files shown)


kant2002:
Please @mousetraps look at this too.

Nodejs/Product/TestAdapter/TestFrameworks/TestFramework.cs
```diff
@@ -111,6 +111,11 @@ class TestFramework {
             return path;
         }
 
+        private string WrapTestNameWithQuot(string path)
```
mousetraps: Minor spelling/grammar error - WrapTestNameWithQuotes 

We should also correct the other method name as well while we're at it

Nodejs/Product/TestAdapter/TestFrameworks/TestFramework.cs
```diff
@@ -111,6 +111,11 @@ class TestFramework {
...
+        {
```
mousetraps: Brace on same line

Nodejs/Product/TestAdapter/TestFrameworks/TestFramework.cs
mousetraps: Also the name of the argument going in should just be testName rather than path.

mousetraps: Lastly including this near the WrapWithQuotes method will make the code a tad more organized

mousetraps:
Minor nitpicky comments and lgtm after that!

kant2002:
I made change @mousetraps. Please review.

mousetraps:
Awesome Thx and merged!

PR #238, opened 2015-06-26
Add more detailed explanation in case of missing startup file.

kant2002:
See #6 and #5 for background.


kant2002:
@mousetraps Could you take a look?

mousetraps:
Thx for the PR! Much better than what we had before. A couple comments:
* Pointing people to project properties too would be helpful to those who find it difficult to map Node-isms to VS-isms because it more closely mirrors to the node cmd line args.
* example filenames aren't necessary, and we can also remove the 'the' before sln explorer to keep the message short and simple. 

So how about something like this? 
> Please select a startup file to launch by right-clicking the file in Solution Explorer and selecting 'Set as Node.js Startup File' or by modifying your configuration in project properties.

kant2002:
Agree. That's even better.

kant2002:
@mousetraps please take a look

PR #242, opened 2015-07-01
Adding wizard to allow Node.js UWP template to add Windows 10 SDK version automatically

munyirik:
This is needed to fix a build bug with NTVS UWP projects and the RTM version of Visual Studio. This is the same fix going into PTVS: https://github.com/Microsoft/PTVS/pull/530


mousetraps:
The code itself lgtm, but is there any reason this isn't going into the extension bits, since its not actually being referenced by any templates shipped with ntvs?

munyirik:
@mousetraps I was thinking that as well but wanted to make minimal changes as well as match what I was doing on the PTVS side. If future the code could be used in other templates.

munyirik:
@mousetraps on second thoughts for this extension it's probably better to create a separate wizard since it will be easier to track issues and fix.

PR #243, opened 2015-07-01
#142, #143, #241 - make it easier to manage browser/node.js-side IntelliSense modes and fix tests

### LINKED ISSUE #142: Make it easier to switch between client-side and server-side IntelliSense modes
The current process is cumbersome, especially if you are dealing with full folders.
https://github.com/Microsoft/nodejstools/wiki/Working-with-Browser-(Clientside)-JavaScript-in-NTVS

### LINKED ISSUE #143: Increase visibility of client-side/server-side modes
We get a lot of questions on this, and people are *generally* okay with the solution, but it would be great to get rid of that initial frustration. 
https://github.com/Microsoft/nodejstools/wiki/Working-with-Browser-(Clientside)-JavaScript-in-NTVS

### LINKED ISSUE #241: Unit tests using AutomationWrapper are not working on VS2015
The issue is specifically with selecting nodes. It looks like we'll need to use OpenSolutionExplorer() rather than going to the SolutionExplorerTree

mousetraps:
Fix #142 - Make it easier to switch between client-side and server-side IntelliSense modes
- Add a context menu to folders that makes it easy to recursively switch between node.js and browser code.
- Choose good defaults for new files based on the current IntelliSense mode
- add unit test

Fix #143 - Increase visibility of client-side/server-side modes
- Recursively traverse the files under a directory to determine the current setting based on whether the js files are set to "Content" or  "Compile", and append a label denoting it as such
- Add option to disable these labels
- Update other UI tests so that they don't break
- add unit test

Fix #241 - unit tests using AutomationWrapper are not working on VS2015
- use OpenSolutionExplorer() before using the SolutionExplorerTree


mousetraps:
without whitespace changes - https://github.com/Microsoft/nodejstools/pull/243/files?w=1

mousetraps:
We should also have an option to toggle this behavior off so that it does not break the other UI tests.

Issue #254, opened 2015-07-06
Exception when switching from Full IntelliSense to Quick IntelliSense


mousetraps:
```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.IO.FileNotFoundException
Stack:
   at System.IO.__Error.WinIOError(Int32, System.String)
   at System.IO.FileInfo.get_Length()
   at Microsoft.NodejsTools.Project.NodejsProjectNode.IncludeNodejsFile(Microsoft.NodejsTools.Project.NodejsFileNode)
   at Microsoft.NodejsTools.Project.NodejsFileNode.get_ShouldAnalyze()
   at Microsoft.NodejsTools.Project.NodejsProjectNode.Reanalyze(Microsoft.VisualStudioTools.Project.HierarchyNode, Microsoft.NodejsTools.Intellisense.VsProjectAnalyzer)
   at Microsoft.NodejsTools.Project.NodejsProjectNode.Reanalyze(Microsoft.VisualStudioTools.Project.HierarchyNode, Microsoft.NodejsTools.Intellisense.VsProjectAnalyzer)
   at Microsoft.NodejsTools.Project.NodejsProjectNode.Reanalyze(Microsoft.VisualStudioTools.Project.HierarchyNode, Microsoft.NodejsTools.Intellisense.VsProjectAnalyzer)
   at Microsoft.NodejsTools.Project.NodejsProjectNode.Reanalyze(Microsoft.VisualStudioTools.Project.HierarchyNode, Microsoft.NodejsTools.Intellisense.VsProjectAnalyzer)
   at Microsoft.NodejsTools.Project.NodejsProjectNode.IntellisenseOptionsPageAnalysisLevelChanged(System.Object, System.EventArgs)
   at System.EventHandler`1[[System.__Canon, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]].Invoke(System.Object, System.__Canon)
   at Microsoft.NodejsTools.Options.NodejsIntellisenseOptionsPage.set_AnalysisLevel(Microsoft.NodejsTools.Options.AnalysisLevel)
   at Microsoft.NodejsTools.Options.NodejsIntellisenseOptionsControl.SyncPageWithControlSettings(Microsoft.NodejsTools.Options.NodejsIntellisenseOptionsPage)
[...]

PR #267, opened 2015-07-09
Add the check for file existence

kant2002:
Also as part of commit found another issue with race conditions.

This is for #254

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
@@ -552,13 +552,16 @@ namespace Microsoft.NodejsTools.Project {
                     return false;
                 }
             }
-            if (new FileInfo(fileNode.Url).Length > _maxFileSize) {
-                // skip obviously generated files...
+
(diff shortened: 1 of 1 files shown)


mousetraps:
Could you elaborate a little more on what you discovered with the race condition and whether we might be able to do something better about it rather than just patching it up here?

Also, the fix looks fine, but any idea why the fileInfo exception occurred? Were you able to reproduce it? I thought it might have had to do with me switching from npm v2 to v3, but never looked into it any further...

kant2002:
When switch Intellisense mode, I discover that `_analyzer` at that time was null, which reasonable, since we recreate analyzer to have new settings. But for me not clear why there is check for `ShouldAnalyze` run before we create new analyzer. Technically we have similar issue when `_analyzer.Project` was null and we have `ShouldAnalyze` called.

kant2002:
Actully I was not able reproduce issue with IOException, so maybe you are right, it could be NPM3 issue. So we could have two issues here. One on NPM3 and one on NPM2 and they exhibit differrnt behavior.

Issue #38, opened 2015-04-07
Make the error message clearer in the REPL window for .npm command 


jinglou-:
When you have a solution with multiple projects and you try to install npm packages from the REPL window, the error prompt for the .npm command wrt the poject name part is miss leading.  We should change it from: 

```Please specify a valid Node.js project or project directory in solution. If solution contains multiple projects, specify target project using .npm [ProjectName or ProjectDir] <npm arguments>```

to something like:

```Please specify a valid Node.js project or project directory in solution. If solution contains multiple projects, specify target project using .npm [ProjectName or ProjectDir] (with the brackets) <npm arguments>```

kant2002:
This text is also a bit misleading.
Maybe we put some example too, since even after reading the issue description don't immediately understand what I have to write.

`Please specify a valid Node.js project or project directory in solution. If solution contains multiple projects, specify target project using .npm [ProjectName or ProjectDir] <npm arguments>
For example: .npm [Lib] list`

or alternatively why the brackets even needed? Maybe fix the parsing logic so it will work without brackets?

kant2002:
Any feedback from @mousetraps or other team?

mousetraps:
I like the approach of using a simple example. While we're at it, fixed up the grammar a bit below and changed Lib to MyApp
```Please specify a valid Node.js project or project directory. If your solution contains multiple projects, specify a target project using .npm [ProjectName or ProjectDir] <npm arguments> For example: .npm [MyApp] list```

We use brackets because we don't want our parsing logic to be too complex or conflict with npm commands (remember, it's not required unless you're using multiple projects.) It's confusing because of the connotations that brackets typically have with cmd line arguments, though :-/

Every so often, we discuss adding a command to let you switch project context altogether, but haven't ever settled on a good design.

kant2002:
Ok, let's change text and we are all set. Will send PR soon.

mousetraps:
Available in the latest dev build: 
github.com/Microsoft/nodejstools/releases

PR #269, opened 2015-07-10
Changed error text in NPM Repl window

kant2002:
Text added to include example, so user easier understand how command should works.

See #38 for background


paulvanbrenk:
:+1:

mousetraps:
:+1:

Issue #253, opened 2015-07-06
"Install New NPM Packages" Refresh Error


fatihyucelsen:
New >> Basic Nodejs Express 4 Application project 
Right Click "npm" in Solution Explorer 
Click "Install New npm packages..." 
New window opens and downloads package list which is about 90MB 
Shows "Refresh Failed" error 

Error Details in Output Window: 
Microsoft.NodejsTools.Npm.SemverVersionFormatException: Invalid semantic version: '0.1.4-DEPRECATED-USE-cfenv-INSTEAD'. The version number must consist of three non-negative numeric parts of the form MAJOR.MINOR.PATCH, with optional pre-release and/or build metadata. The optional parts may only contain characters in the set [0-9A-Za-z-]. 

Platform: 
Windows 8.1
Visual Studio 2015 Community Edition 
Nodejs Tools v1.1 Beta

mousetraps:
Unfortunately I can't repro this issue. Can you try again after clearing your npm cache?
![image](https://cloud.githubusercontent.com/assets/762848/8519632/f91840e8-2388-11e5-8a36-d7cbfdf737f2.png)

fatihyucelsen:
I try it. useless.

Detail:
http://txs.io/xm6b

mousetraps:
hmm you shouldn't be seeing the upgrade message at the beginning if the db was deleted. You can double check (and delete it manually) in "%localappdata%\microsoft\node.js tools". Go ahead and delete that entire folder and restart VS. 

If that doesn't work, I can send you a copy of my db sometime tomorrow.

fatihyucelsen:
Unfortunately this did not work either.

http://txs.io/1m6b

billti:
I tried with the latest release and got no issue either. I also extracted the [code that throws that exception](https://github.com/Microsoft/nodejstools/blob/35817ffdaa1797f95628a22f588fc3ed885f884b/Nodejs/Product/Npm/SemverVersion.cs#L47) into a test app and fed it the semver it is complaining about (i.e. '0.1.4-DEPRECATED-USE-cfenv-INSTEAD') and it parses it fine. There must be something else weird going on here, as the RegEx that parses this looks pretty straight forward and hasn't changed in a while.

What language OS/VS are you running?

kant2002:
I also has no such issues on the latest bits.
Could you @fatihyucelsen make screenshot of the Help->About with information about NodeJS version ?

kant2002:
[...]

PR #270, opened 2015-07-10
Simplify regex and test on Turkish locale

kant2002:
I remove RegexOptions.IgnoreCase since somehow this break the whole recognition on Turkish locale and I don't see any point have it. We already define both uppercase and lower case letters in the regex. Alternative for A-Za-z is to use \w and for 0-9 use \d, so maybe this will lead to more reliable version recognizer. If strict fragment rules needed they should be placed to RegexOptionalFragment, but I was unable to make this work reliable so far.

See #253

### DIFF
--- a/Nodejs/Product/Npm/SPI/NpmGetCatalogCommand.cs
+++ b/Nodejs/Product/Npm/SPI/NpmGetCatalogCommand.cs
@@ -122,7 +122,8 @@ namespace Microsoft.NodejsTools.Npm.SPI {
                                         if (!string.IsNullOrEmpty(latestVersion)) {
                                             try {
                                                 builder.LatestVersion = SemverVersion.Parse(latestVersion);
-                                            } catch (SemverVersionFormatException) { 
+                                            } catch (SemverVersionFormatException) {
+                                                OnOutputLogged(String.Format(Resources.InvalidPackageSemVersion, latestVersion, builder.Name));
                                             }
                                         }
                                     }
--- a/Nodejs/Product/Npm/SPI/NpmGetCatalogCommand.cs
+++ b/Nodejs/Product/Npm/SPI/NpmGetCatalogCommand.cs
@@ -120,7 +120,10 @@ namespace Microsoft.NodejsTools.Npm.SPI {
                                             .FirstOrDefault();
 
                                         if (!string.IsNullOrEmpty(latestVersion)) {
-                                            builder.LatestVersion = SemverVersion.Parse(latestVersion);
+                                            try {
+                                                builder.LatestVersion = SemverVersion.Parse(latestVersion);
+                                            } catch (SemverVersionFormatException) { 
+                                            }
                                         }
                                     }
 
--- a/Nodejs/Product/Npm/SemverVersion.cs
+++ b/Nodejs/Product/Npm/SemverVersion.cs
@@ -33,11 +33,11 @@ namespace Microsoft.NodejsTools.Npm {
             + "\\.(?<patch>[…0-9]+)" // The '…' is there to handle the 'classy' library, which has a very long version number - see slightly snarky comment about that unadulterated bag of hilarity below.
             + "(?:-(?<prerelease>[…0-9A-Za-z-]+(\\.[…0-9A-Za-z-]+)*))?"
             + "(?:\\+(?<buildmetadata>[…0-9A-Za-z-]+(\\.[…0-9A-Za-z-]+)*))?$",
-            RegexOptions.IgnoreCase | RegexOptions.Singleline);
+            RegexOptions.Singleline);
 
         private static readonly Regex RegexOptionalFragment = new Regex(
             "^[…0-9A-Za-z-]+(\\.[…0-9A-Za-z-]+)*$",
-            RegexOptions.IgnoreCase | RegexOptions.Singleline);
+            RegexOptions.Singleline);
 
         public static SemverVersion Parse(string versionString) {
             var matches = RegexSemver.Matches(versionString);
--- a/Nodejs/Product/Npm/Resources.Designer.cs
+++ b/Nodejs/Product/Npm/Resources.Designer.cs
@@ -223,6 +223,15 @@ namespace Microsoft.NodejsTools.Npm {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Semversion {0} for package {1} is invalid.
+        /// </summary>
+        internal static string InvalidPackageSemVersion {
+            get {
+                return ResourceManager.GetString("InvalidPackageSemVersion", resourceCulture);
+            }
+        }
+        
+        /// <summary>
         ///   Looks up a localized string similar to npm command cancelled.
         /// </summary>
         internal static string NpmCommandCancelled {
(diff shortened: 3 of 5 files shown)


billti:
Great job figuring this out! The docs do actually call out the ignoreCase can cause issues on Turkish locale (see https://msdn.microsoft.com/en-us/library/yd1hzczs.aspx#Invariant ). Using \w or \d would include non-ASCII characters in a .NET RegEx unless ECMAScript mode is enabled, so using the character ranges explicitly is still the best way to go here I think. Other comments on the code.

Nodejs/Product/Npm/SemverVersion.cs
```diff
@@ -31,13 +31,13 @@ public struct SemverVersion {
             "^(?<major>[0-9]+)"
             + "\\.(?<minor>[0-9]+)"
             + "\\.(?<patch>[…0-9]+)" // The '…' is there to handle the 'classy' library, which has a very long version number - see slightly snarky comment about that unadulterated bag of hilarity below.
-            + "(?:-(?<prerelease>[…0-9A-Za-z-]+(\\.[…0-9A-Za-z-]+)*))?"
-            + "(?:\\+(?<buildmetadata>[…0-9A-Za-z-]+(\\.[…0-9A-Za-z-]+)*))?$",
-            RegexOptions.IgnoreCase | RegexOptions.Singleline);
+            + "(?:-(?<prerelease>[…0-9A-Za-z-.]+))?"
```
billti: I get adding the "." to allow multi-part pre-release tags (i.e. "1.0.1-beta.1"), but doesn't this also allow invalid formats such as "1.0.1-..5.."? Also, I though the "-" had to be immediately after the opening "[" or before the closing "]" to be treated as a regular character rather than the range operator.

Nodejs/Product/Npm/SemverVersion.cs
```diff
@@ -31,13 +31,13 @@ public struct SemverVersion {
             "^(?<major>[0-9]+)"
             + "\\.(?<minor>[0-9]+)"
             + "\\.(?<patch>[…0-9]+)" // The '…' is there to handle the 'classy' library, which has a very long version number - see slightly snarky comment about that unadulterated bag of hilarity below.
```
billti: p.s. I still don't get what the '…' actually does here. It's not valid text in a SemVer version or a RegEx special char (that I can find), and I don't see how it helps with the 'classy' library. If someone can expand on how this actually makes the code behave more correctly, I'd appreciate it.

Nodejs/Product/Npm/SemverVersion.cs
```diff
@@ -60,12 +60,26 @@ public struct SemverVersion {
...
+                    preReleaseFragment = preRelease.Value;
+                }
+
+                string buildMetadataFragment;
+                if (!buildMetadata.Success || !IsValidOptionalFragment(buildMetadata.Value)) {
+                    buildMetadataFragment = null;
+                } else {
+                    buildMetadataFragment = buildMetadata.Value;
+                }
+
```
billti: I don't see what value adding this code gets. The only difference over the original ternary operators is that is also calls IsValidOptionalFragment, but that just checks the fragment matches a RegEx which is identical to the match in the original RegEx that created the "prerelease" capture group. Thus is must be valid, else it wouldn't have been in the capture group and prerelease.Success would be false.

Nodejs/Product/Npm/SemverVersion.cs
```diff
@@ -31,13 +31,13 @@ public struct SemverVersion {
...
-            + "(?:\\+(?<buildmetadata>[…0-9A-Za-z-]+(\\.[…0-9A-Za-z-]+)*))?$",
-            RegexOptions.IgnoreCase | RegexOptions.Singleline);
+            + "(?:-(?<prerelease>[…0-9A-Za-z-.]+))?"
+            + "(?:\\+(?<buildmetadata>[…0-9A-Za-z-.]+))?$",
+            RegexOptions.Singleline);
 
         private static readonly Regex RegexOptionalFragment = new Regex(
-            "^[…0-9A-Za-z-]+(\\.[…0-9A-Za-z-]+)*$",
-            RegexOptions.IgnoreCase | RegexOptions.Singleline);
+            "^[…0-9A-Za-z-.]+$",
```
billti: Same here and on line 35. The prior pattern of `^[…0-9A-Za-z-]+(\\.[…0-9A-Za-z-]+)*` looks right to me in terms of disallowing invalid formats. `^[…0-9A-Za-z-.]+` seems a little too lax. Why is this change needed?

Nodejs/Product/Npm/SemverVersion.cs
mousetraps: Way back when... npm did not have the db stored in a json format, so we used to parse the textual output ([here's an example](https://raw.githubusercontent.com/Microsoft/nodejstools/1bb20aa7ed8b8ace307825ed7072ea54fb7ff40d/Nodejs/Tests/TestData/NpmSearchData/npmsearchfullcat_npm143.txt))

We've since [removed that code](https://github.com/Microsoft/nodejstools/commit/a1cb641ce7a9bdf7fe240d3ae069bde1f9414a2b), but looks like we missed a spot.

In general, this file needs some attention, but it's one of those areas that's been around for a while, and works well enough.

mousetraps: (so this was handling the case where the textual output for the version number was cut off by the next column, and therefore ended with a ...)

billti: Ah, got it. So it's not part of the version number, just part of the textual output the code used to scrape, where it used this char when truncating a column. Makes sense.

It would be good to just delete these if no longer needed. I spent quite a while looking at the SemVer spec, the 'classy' history, and the RegEx operators to trying and figure out what the hell it might be doing :confused:

mousetraps:
Nice job tracking this down! In addition to Bill's comments... let's go ahead and catch this exception during GetCatalog (and print to the output window) so that people aren't left empty handed if parsing fails (as we do with ArgumentException)
https://github.com/Microsoft/nodejstools/blob/master/Nodejs/Product/Npm/SPI/NpmGetCatalogCommand.cs#L143-L148

kant2002:
Probably was too late to debug this issue, now patch is much more clean. I just have to remove `RegexOptions.IgnoreCase`
@mousetraps and @billti please review.

Nodejs/Product/Npm/SPI/NpmGetCatalogCommand.cs
```diff
@@ -120,7 +120,10 @@ internal class NpmGetCatalogCommand : NpmCommand, IPackageCatalog {
                                             .FirstOrDefault();
 
                                         if (!string.IsNullOrEmpty(latestVersion)) {
-                                            builder.LatestVersion = SemverVersion.Parse(latestVersion);
+                                            try {
+                                                builder.LatestVersion = SemverVersion.Parse(latestVersion);
+                                            } catch (SemverVersionFormatException) {
```
mousetraps: Also print out a message here - we want to know about the issue, but don't want it to block the user.

mousetraps:
minor comment, but lgtm other than that!

mousetraps:
@kant2002 I presume you mistakenly closed this since it looks like you restored your branch too, so I went ahead ad re-opened it :smile:

kant2002:
True. I just add reporting of the message in case of invalid package semversion

PR #274, opened 2015-07-12
Added properties for overriding the default browser in debug mode

obeobe:
In VS 2013, when starting to debug, if the "Start Web Browser" property is ticked, the browser selected in the browsers combo is started.

However, any custom arguments associated with that browser seem to be ignored. I like to debug in Chrome using a specific profile, which requires passing custom arguments.

I tried to find a way to look up the arguments of the selected browsers but couldn't. Perhaps someone who is more knowledgeable in the VS SDK would be able to.

Instead, I added the ability to override the default browser with a custom executable and arguments in the project properties sheet.

### DIFF
--- a/Nodejs/Product/Nodejs/Debugger/DebugEngine/AD7Engine.cs
+++ b/Nodejs/Product/Nodejs/Debugger/DebugEngine/AD7Engine.cs
@@ -69,6 +69,8 @@ namespace Microsoft.NodejsTools.Debugger.DebugEngine {
         private Guid _ad7ProgramId;             // A unique identifier for the program being debugged.
         private static readonly HashSet<WeakReference> Engines = new HashSet<WeakReference>();
         private string _webBrowserUrl;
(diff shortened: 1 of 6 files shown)


mousetraps:
Hey there, thanks for the two PRs! Just wanted to apologize in advance for the potential delay in reviewing this (right now we're heads down on a release set for July 20 in line with VS RTM, but we'll be coming up for air shortly after that :smiley:)

kant2002:
Any news on this PR?

kant2002:
@mousetraps Could you look on this PR. It is very useful for debugging

mousetraps:
Thanks again for the contribution :smiley:. Love the idea, but this functionality seems better suited for the VS platform itself than it is for NTVS specifically (ideally you'd be able to customize your browser settings from the launch dropdown itself), so I'll pass it on to that team. Additionally, we're working through some project system changes, so the timing isn't great.

PR #280, opened 2015-07-13
Fix #139 Add docker integration

### LINKED ISSUE #139: Docker integration
Similar to the support for ASP.NET
https://www.microsoft.com/en-us/download/details.aspx?id=46891

mousetraps:
- adds docker item template
- add integration point with Docker Tools

### DIFF
--- a/Nodejs/Product/Nodejs/Nodejs.csproj
+++ b/Nodejs/Product/Nodejs/Nodejs.csproj
@@ -775,7 +775,7 @@
     <ZipProject Include="ProjectTemplates\AzureExpress4App\Web.Debug.config" />
     <ZipItem Include="Templates\Files\DockerfileTemplate\Dockerfile" />
     <VSIXSourceItem Include="Templates\Files\DockerfileTemplate\Dockerfile">
-      <VSIXSubPath>Templates\Files\DockerfileTemplate\</VSIXSubPath>
+      <VSIXSubPath>FileTemplates\ReferenceFiles\Dockerfile</VSIXSubPath>
     </VSIXSourceItem>
     <None Include="Theme\contrast.vstheme">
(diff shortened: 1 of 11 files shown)


billti:
Why are both the XML files showing up as being fully replaced? Is this just a line-ending thing? Makes it hard to see what has changed.

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -1597,6 +1604,7 @@
     </ItemGroup>
   </Target>
   <PropertyGroup>
-    <PostBuildEvent>copy /y $(OutDir)..\WebRole\Microsoft.NodejsTools.WebRole.dll $(OutDir)Debugger\RemoteDebug\</PostBuildEvent>
+    <PostBuildEvent>echo f | xcopy /y $(ProjectDir)Templates\Files\DockerfileTemplate\Dockerfile $(OutDir)Templates\Files\DockerfileTemplate\Dockerfile
```
billti: What is the "echo f" for?

mousetraps: xcopy will create new directories (unlike copy), but also asks whether it's a file or a directory (f or d), and there is no other way to suppress this.

mousetraps:
Yes, it's a line ending thing (we really need to sweep through and fix all the files in one go to prevent this from continuing to happen)
you can use append ?w=1 to the url view it without the line endings
https://github.com/Microsoft/nodejstools/pull/280/files?w=1

mousetraps:
Fixed the installer to add the loose dockerfile as well.

mousetraps:
- adds docker item template
- add integration point with Docker Tools

PR #288, opened 2015-07-14
#287 Add option to skip devenv /setup

### LINKED ISSUE #287: Add option to skip devenv /setup
This is required for the IOT bundle installer, so we need to get this into the next release.

mousetraps:
Fix #287


mousetraps:
@munyirik FYI

With this change you'll want to run with SKIP_DEVENV_SETUP="TRUE" to skip devenv/setup in the msi. This will be available in our next build

PR #303, opened 2015-07-21
String var containing time, not displayed correctly

### LINKED ISSUE #106: String var containing time, not displayed correctly.
Variable time from following sample is in watch window displayed as: "05/07/2015 22:00:00", but of  course logged to console correctly as  "2015-05-07T22:00:00.000Z".

var time= "2015-05-07T22:00:00.000Z";
console.log(time);

kant2002:
This pull request is not fully fix #106, but just to indicate problems which was discovered during the debugging.

### DIFF
--- a/Nodejs/Product/Nodejs/Debugger/Serialization/NodeBacktraceVariable.cs
+++ b/Nodejs/Product/Nodejs/Debugger/Serialization/NodeBacktraceVariable.cs
@@ -30,7 +30,7 @@ namespace Microsoft.NodejsTools.Debugger.Serialization {
             StackFrame = stackFrame;
             Name = (string)parameter["name"] ?? NodeVariableType.AnonymousVariable;
             TypeName = (string)value["type"];
-            Value = (string)value["value"];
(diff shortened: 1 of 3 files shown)


mousetraps:
Thx. This approach makes sense, but there should be a more direct way to get at the the value as a string. I'd need to refresh my memory a bit more on Newtonsoft.Json, but have you tried this method?
http://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_Linq_JToken_Value__1.htm

kant2002:
This method just convert `Value` property to appropriate type. Since that property contains `DateTime` type method is useless.

mousetraps:
Have you gotten a chance to update this in response to https://github.com/Microsoft/nodejstools/issues/106#issuecomment-138608332?

kant2002:
Sorry forget completely about that. Will take a look on the option provided.

mousetraps:
Would *love* to get this fix in, but closing because we haven't heard back in a while. If the proposed methods aren't working, then let us know and we can revisit the less "elegant" solution :smiley:

PR #320, opened 2015-07-27
Fix #319 Failing to parse some PackageJson Person fields

### LINKED ISSUE #319: Failing to parse some Person PackageJson fields
This results in an assertion during debug mode, and also a failure to properly display those fields in the UI

mousetraps:
Fix issues while parsing Person package.json fields
* we were not properly handling cases when the name, email, and url fields
      were out of order.
* fix assertion check when result is empty
* add relevant test cases

### DIFF
--- a/Nodejs/Product/Npm/SPI/Person.cs
+++ b/Nodejs/Product/Npm/SPI/Person.cs
@@ -83,7 +83,8 @@ namespace Microsoft.NodejsTools.Npm.SPI {
             // Verify we are parsing correctly
             try {
                 var jObject = JObject.Parse(source);
-                Debug.Assert((string)jObject["name"] == Name, string.Format("Failed to parse name from {0}", source));
+                var name = (string)jObject["name"];
+                Debug.Assert(name != null ? name == Name : Name == source, string.Format("Failed to parse name from {0}", source));
                 Debug.Assert((string)jObject["email"] == Email, string.Format("Failed to parse email from {0}", source));
                 Debug.Assert((string)jObject["url"] == Url, string.Format("Failed to parse url from {0}", source));
             } catch (Exception) {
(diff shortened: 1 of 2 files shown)


Nodejs/Tests/NpmTests/PackageJsonTests.cs
```diff
@@ -108,6 +113,10 @@ public class PackageJsonTests : AbstractPackageJsonTests {
    ""name"": ""mypackage"",
    ""version"": ""0.7.0"",
    ""description"": ""Sample package for CommonJS. This package demonstrates the required elements of a CommonJS package."",
+   ""author"": {
```
paulvanbrenk: Add more test cases with only a single field and with an empty author element.

Nodejs/Product/Npm/SPI/Person.cs
```diff
@@ -71,9 +83,9 @@ internal class Person : IPerson {
             // Verify we are parsing correctly
             try {
                 var jObject = JObject.Parse(source);
-                Debug.Assert(((string)jObject["name"] ?? string.Empty) == Name, string.Format("Failed to parse name from {0}", source));
-                Debug.Assert(((string)jObject["email"] ?? string.Empty) == Email, string.Format("Failed to parse email from {0}", source));
-                Debug.Assert(((string)jObject["url"] ?? string.Empty) == Url, string.Format("Failed to parse url from {0}", source));
+                Debug.Assert((string)jObject["name"] == Name, string.Format("Failed to parse name from {0}", source));
```
paulvanbrenk: Why bother with this... we're extracting them using the regex, do we still care what the json parser makes of it...

mousetraps: We don't care unless we're doing it wrong. It's nice to keep this codepath around in an #if DEBUG section and assert we're getting the same results as our own parser in case there are any inputs we're not handling properly (which is how we managed to track down this bug in the first place.)

paulvanbrenk:
:+1:

Nodejs/Tests/NpmTests/PackageJsonTests.cs
mousetraps: good call. added a few more.

PR #325, opened 2015-07-27
Fix #297 add support for relative node.exe paths

### LINKED ISSUE #297: Node.exe path should support relative paths
I upgraded from 1.1 Beta to the 7-15-2015 dev build and reopened my solution and was unable to run it.

I had a relative path in the project settings under "Node.exe path". I use the [electron](https://github.com/atom/electron) [pre-built package](https://www.npmjs.com/package/electron-prebuilt) which installs the executable under `node_modules\electron-prebuilt\dist\electron.exe` relative to the project. It used to work but no longer does.

Specifying a relative path is key in allowing us to submit the .njsproj file into source control and having others be able to debug/run the project without having to modify the project file to specify an absolute path on their machines.

mousetraps:
Additionally...
* Consolidated Node exe path resolution to reduce code duplication and fix other issues due to path resolution inconsistencies
* added basic node.exe path resolution unit tests

### DIFF
--- a/Common/Product/SharedProject/CommonProjectNode.cs
+++ b/Common/Product/SharedProject/CommonProjectNode.cs
@@ -1539,8 +1539,7 @@ namespace Microsoft.VisualStudioTools.Project {
         /// Returns resolved value of the current working directory property.
         /// </summary>
         public string GetWorkingDirectory() {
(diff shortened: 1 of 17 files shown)


mousetraps:
Without whitespace: https://github.com/Microsoft/nodejstools/pull/325/files?w=1

Issue #158, opened 2015-06-02
Please add support for snippets


waynebloss:
Snippets currently only work in client side Javascript (when the file is marked with Build Action: Content).

It would be great if snippets also worked in NodeJS Javascript files.

mousetraps:
This has been fixed in #339

mjbvz:
Quick update in case anyone runs across this. I removed the snippets features from NTVS 1.3 with #1263 since they already did not work in NTVS 1.2 when using ES6 IntelliSense. Our editor is now powered by TypeScript, and the TypeScript team is adding support for snippets: Microsoft/TypeScript#312 

In VS 2017 ("VS15"), the new TypeScript snippets are now available:

![image](https://cloud.githubusercontent.com/assets/12821956/19535959/c2c3d310-95fe-11e6-95dd-b9fc055d9662.png)

I'll have to check if this was also ported back to TypeScript 2.x for VS2015.

PR #339, opened 2015-07-30
Fix #317 add snippet support for node.js tools

### LINKED ISSUE #317: use Custom Code Snippets
Hello when working with server files or better with property Build Action = Compile,
there is no chance to use any code snippet at all.

For example right click Insert Code Snippet Ctrl+K, Ctrl+X is not available.

avitalb:


Nodejs/Product/AssemblyVersion.cs
```diff
@@ -1,64 +1,64 @@
...
+    public const string ReleaseVersion = "1.0";
+
+    // This version string (and the comment for Version) should be updated
+    // manually between minor releases (e.g. from 1.0 to 1.1).
+    // Servicing branches and prereleases should retain the value.
+    public const string FileVersion = "1.1";
+
+    // This version should never change from "4100.00"; BuildRelease.ps1
+    // will replace it with a generated value.
+    public const string BuildNumber = "30730.00";
```
mousetraps: This change shouldn't be checked in

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,332 @@
+﻿using System;
```
mousetraps: Be sure to include the copyright header for all the new files

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,332 @@
...
+                            // we didn't actually have a selction, and our negative tracking pushed us
+                            // back to the start of the buffer...
+                            end = start;
+                        }
+                        var selectedSpan = Span.FromBounds(start, end);
+
+                        if (surroundsWithStatement &&
+                            String.IsNullOrWhiteSpace(_textView.TextBuffer.CurrentSnapshot.GetText(selectedSpan))) {
+                            // we require a statement here and the user hasn't selected any code to surround,
+                            // so we insert a pass statement (and we'll select it after the completion is done)
```
mousetraps: Luckily JavaScript is less annoying than python, so this can all be simplified :smiley:

The reason we did this in python was because python requires that code blocks cannot be empty. So the following code isn't syntactically valid. 
```python
while True:
    # nothingness
``` 

Instead, python has a no-op ("do nothing") statement for these cases.
```python
while True:
    pass
````

**Comment by mousetraps** [comment]
GH notifications are annoying so sorry for the sporadic comments - still reviewing - got pulled off to help with a TR demo and I'll be back to this after that :smiley:

File: Nodejs/Product/Nodejs/Guids.cs [hunk 4]
```diff
@@ -33,8 +33,9 @@ static class Guids
         public const string NodejsBaseProjectFactoryString = "9092AA53-FB77-4645-B42D-1CCCA6BD08BD";
         public const string TypeScriptLanguageInfoString = "87bdf188-e6e8-4fcf-a82a-9b8506e01847";
         public const string JadeEditorFactoryString = "6CB69EF8-1329-4DC0-84B4-FA134EA59BE3";
+        public const string DefaultLanguageServiceString = "{8239BEC4-EE87-11D0-8C98-00C04FC2AB22}";
 
-        internal static readonly Guid DefaultLangaugeService = new Guid("{8239BEC4-EE87-11D0-8C98-00C04FC2AB22}");
+        internal static readonly Guid DefaultLanguageService = new Guid(DefaultLanguageServiceString);
```
mousetraps: You refer to this guid in a bunch of places, but I'm not sure that's really the one you want. Essentially the default language service (JSLS) handles javascript completions within VS by default. In order to work around some of the performance issues caused by JSLS, we open javascript files in our own Node editor using our own language service (represented by the ```NodejsLanguageInfoString``` above. So if you want snippets to appear in the Node editor, you should be referencing the Nodejs string rather than the Default.

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,332 @@
...
+            List<string> importList = new List<string>();
+            if (ErrorHandler.Succeeded(hr = _session.GetSnippetNode("CodeSnippet:Imports", out imports))
+                && imports != null) {
+                foreach (IXMLDOMNode import in imports.childNodes) {
+                    var id = import.selectSingleNode("./CodeSnippet:Namespace");
+                    if (id != null) {
+                        importList.Add(id.text);
+                    }
+                }
+            }
```
mousetraps: You're not actually using Imports or importList anywhere. Do you have plans to? For instance, it might be useful with any snippets that use require

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,332 @@
...
+                    // snippet type.
+                    //
+                    // But, to make things even more complicated, we don't have a good indication of what's the 
+                    // template text vs. what's the selected text.  We do have access to the original template,
+                    // but all of the values have been replaced with their default values when we get called
+                    // here.  So we need to go back and re-apply the template, except for the $selected$ part.
+                    //
+                    // Also, the text has \n, but the inserted text has been replaced with the appropriate newline
+                    // character for the buffer.
+                    var templateText = codeNode.text.Replace("\n", _textView.Options.GetNewLineCharacter());
```
mousetraps: The code snippets that you added are all CRLF (```\r\n```). They would need to be ```\n``` in order for this replacement to make sense.

Additionally, use ```VsExtensions.GetNewLineText``` instead of ```GetNewLineCharacter()```, which works around some annoying VS issues...
https://github.com/Microsoft/nodejstools/blob/master/Common/Product/SharedProject/VsExtensions.cs#L162

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,332 @@
...
+                    foreach (var decl in declList) {
+                        string defaultValue;
+                        if (ErrorHandler.Succeeded(_session.GetFieldValue(decl, out defaultValue))) {
+                            templateText = templateText.Replace("$" + decl + "$", defaultValue);
+                        }
+                    }
+                    templateText = templateText.Replace("$end$", "");
```
mousetraps: $end$ isn't used by any of the snippets currently. I recommend using it in instances where we want to ensure the cursor is in a position where a dev can continue typing right away.

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,332 @@
...
+
+                    // we can finally figure out where the selected text began witin the original template...
+                    int selectedIndex = templateText.IndexOf("$selected$");
```
mousetraps: $selected$ isn't used in any of the snippets currently. I recommend using it for snippets that begin with with a console.log that people might want to overwrite with their own app logic.

Nodejs/Product/Nodejs/Intellisense/IntellisenseController.cs
```diff
@@ -1,663 +1,772 @@
...
+                if (_textView.Selection.IsActive && !_textView.Selection.IsEmpty) {
+                    // when deleting a selection don't do anything to pop up signature help again
+                    _sigHelpSession.Dismiss();
+                    return false;
+                }
+
+                SnapshotPoint? caretPoint = _textView.BufferGraph.MapDownToFirstMatch(
+                    _textView.Caret.Position.BufferPosition,
+                    PointTrackingMode.Positive,
+                    EditorExtensions.IsNodeJsContent,
```
mousetraps: If you're going to make this change, you may as well remove the IsNodejsContent method in this file altogether since the two functions are exactly the same.

Nodejs/Product/Nodejs/Intellisense/IntellisenseController.cs
```diff
@@ -1,663 +1,772 @@
...
+                        case VSConstants.VSStd2KCmdID.END:
+                        case VSConstants.VSStd2KCmdID.WORDPREV:
+                        case VSConstants.VSStd2KCmdID.WORDPREV_EXT:
+                        case VSConstants.VSStd2KCmdID.DELETEWORDLEFT:
+                            _sigHelpSession.Dismiss();
+                            _sigHelpSession = null;
+                            break;
+                    }
+                }
+            } else {
```
mousetraps: Is there a reason you added this else statement here? It changes the ctrl flow unnecessarily and has potential to introduce subtle bugs.

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -773,6 +777,102 @@
     <ZipProject Include="ProjectTemplates\AzureExpress4App\setup_web.cmd" />
     <ZipProject Include="ProjectTemplates\AzureExpress4App\Web.config" />
     <ZipProject Include="ProjectTemplates\AzureExpress4App\Web.Debug.config" />
+    <None Include="Snippets\1033\Nodejs\alert.snippet">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
```
mousetraps: All of the snippets must have the IncludeInVSIX property set to true, otherwise they won't be packaged and included in the experimental hive extension folder. Additionally you'll want to set the Build Action to Content rather than None (required to use the IncludeInVSIX property anyways).

Lastly, there's no reason to always copy to the output directory in this scenario - PreserveNewest is sufficient here. Otherwise what happens is that the project will rebuild every time you build.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/alert.snippet
```diff
@@ -0,0 +1,25 @@
...
+    </Header>
+    <Snippet>
+      <Declarations>
+        <Literal>
+          <ID>message</ID>
+          <Default>message</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[alert($message$);]]>
```
mousetraps: alerts are really only useful for browser-side code (there's no such concept in node because there's no front-facing component), so you can probably go ahead and remove this snippet since browser-side code is best dealt with in the  default VS editor. Especially since if we're getting to browsers, then we're opening up a whole new can of worms.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/cl.snippet
```diff
@@ -0,0 +1,19 @@
...
+      <Title>cl</Title>
+      <Description>Code Snippet for a Console message</Description>
+      <Shortcut>cl</Shortcut>
+      <SnippetTypes>
+        <SnippetType>Expansion</SnippetType>
+      </SnippetTypes>
+    </Header>
+    <Snippet>
+      <Code Language="Nodejs">
+        <![CDATA[console.log();]]>
```
mousetraps: console.log($message$) instead. There's no reason to have an empty console.log()

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/class.snippet
```diff
@@ -0,0 +1,31 @@
...
+    </Header>
+    <Snippet>
+      <Declarations>
+        <Literal>
+          <ID>MyClass</ID>
+          <Default>MyClass</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[class $MyClass$ {
```
mousetraps: Let's stick to ES5. Classes are an ES6 language feature that are not currently compatible with the Node editor anyways (you'll get errors if you type this into the editor.)

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/exports.snippet
```diff
@@ -0,0 +1,26 @@
...
+    </Header>
+    <Snippet>
+      <Declarations>
+        <Object>
+          <ID>myModule</ID>
+          <Default>myModule</Default>
+        </Object>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[module.exports.$myModule$;]]>
```
mousetraps: typically you want to assign something to ```module.exports```

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/http_delete.snippet
```diff
@@ -0,0 +1,28 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<CodeSnippets
+    xmlns="http://schemas.microsoft.com/VisualStudio/2005/CodeSnippet">
+  <CodeSnippet Format="1.0.0">
+    <Header>
+      <Title>delete</Title>
+      <Description>Code Snippet for handling an HTTP DELETE request with Express</Description>
+      <Shortcut>http_delete</Shortcut>
```
mousetraps: This actually looks like express delete, not http.delete (which doesn't exist anyways) - same with some of the other snippets.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/http_get.snippet
```diff
@@ -0,0 +1,28 @@
...
+    </Header>
+    <Snippet>
+      <Declarations>
+        <Literal>
+          <ID>Hello World</ID>
+          <Default>Hello World</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[app.get('/', function (request, response) {
```
mousetraps: the routes in this and similar snippets should be editable as well because people typically want to specify more than the root page, so better to use ```'$route$'``` than ```'/'```

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/initConfig.snippet
```diff
@@ -0,0 +1,24 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<CodeSnippets
+    xmlns="http://schemas.microsoft.com/VisualStudio/2005/CodeSnippet">
+  <CodeSnippet Format="1.0.0">
+    <Header>
+      <Title>initConfig</Title>
+      <Description>Code Snippet for intializing a Grunt configuration object</Description>
+      <Shortcut>initConfig</Shortcut>
```
mousetraps: On second thought, let's hold off on grunt for now and see if we can just integrate better with the already built-in support in VS (which includes a snippet)
http://webtooling.visualstudio.com/task-runners/grunt/

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/parse.snippet
```diff
@@ -0,0 +1,25 @@
...
+    </Header>
+    <Snippet>
+      <Declarations>
+        <Literal>
+          <ID>string</ID>
+          <Default>string</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[JSON.parse('$string$');]]>
```
mousetraps: it's rare that someone would want to parse an explicit string rather than a variable here so don't bother with the quotes.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/readfile.snippet
```diff
@@ -0,0 +1,29 @@
...
+    <Snippet>
+      <Declarations>
+        <Literal>
+          <ID>file</ID>
+          <Default>file</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[fs.readFile('$file$', function (error, data) {
+    if (err) throw err;
```
mousetraps: This error should be editable as it is in the other snippets otherwise I'm not convinced it's really worth including since re-throwing the error isn't particularly valuable. As a dev, I'd prefer to just use the if snippet afterwards to add the error handling

Also, the function argument is ```error``` but the if statement checks for ```err``` so it will never throw anyways.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/server.snippet
```diff
@@ -0,0 +1,32 @@
...
+          <Default>message</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[var http = require('http');
+var server = http.createServer(function (request, response) {
+    response.writeHead(200, { "Content-Type": "text/plain" });
+    response.end("$message$");
+});
+server.listen(80);
```
mousetraps: Instead of 80, use process.env.PORT || 3000 to be consistent with our other project templates.

process.env.PORT is also more likely correct and is going to work in deployment scenarios. When you deploy a node app, the PORT environment variable is typically set by the production server (which will forward that port as appropriate.) And in the event that PORT doesn't exist, you want to include a fallback port.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/stringify.snippet
```diff
@@ -0,0 +1,25 @@
...
+    </Header>
+    <Snippet>
+      <Declarations>
+        <Object>
+          <ID>object</ID>
+          <Default>object</Default>
+        </Object>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[JSON.stringify($object$);]]>
```
mousetraps: JSON.stringify and JSON.parse are typically used as a part of a statement rather than the full statement itself so you probably don't want the semicolon.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/switch.snippet
```diff
@@ -0,0 +1,29 @@
...
+    </Header>
+    <Snippet>
+      <Declarations>
+        <Literal>
+          <ID>true</ID>
+          <Default>true</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[switch ($true$) {
```
mousetraps: In this case there's no value to supplying something like true/false, so I'd just go with ```switch_on```, which is consistent with the default javascript editing experience in VS

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/tryc.snippet
```diff
@@ -0,0 +1,24 @@
...
+      <Shortcut>tryc</Shortcut>
+      <SnippetTypes>
+        <SnippetType>Expansion</SnippetType>
+      </SnippetTypes>
+    </Header>
+    <Snippet>
+      <Code Language="Nodejs">
+        <![CDATA[try {
+
+} catch (error) {
```
mousetraps: you probably want to make the error a replaceable parameter here since people are picky about what they name their errors. I believe it's also more common to use ```e``` rather than ```error```

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/tryf.snippet
```diff
@@ -0,0 +1,24 @@
...
+      <Shortcut>tryf</Shortcut>
+      <SnippetTypes>
+        <SnippetType>Expansion</SnippetType>
+      </SnippetTypes>
+    </Header>
+    <Snippet>
+      <Code Language="Nodejs">
+        <![CDATA[try {
+
+} finally {
```
mousetraps: Any reason not to include a try, finally over try, catch, finally?

Nodejs/Setup/NodejsTools/NodejsTools.wxs
```diff
@@ -18,11 +18,17 @@
                 <Directory Id="Dir_$(var.key)TestFrameworksMocha" Name="Mocha" />
             </Directory>
             <Directory Id="Dir_$(var.key)InstallLocationScripts" Name="Scripts">
-                <Directory Id="Dir_$(var.key)InstallLocationScriptsTypings" Name="Typings">
+                <Directory Id="Dir_$(var.key)InstallLocationScriptsTypings" Name="Typings">
```
mousetraps: remove extra whitespace

Nodejs/Setup/NodejsTools/NodejsTools.wxs
```diff
@@ -18,11 +18,17 @@
...
-                <Directory Id="Dir_$(var.key)InstallLocationScriptsTypings" Name="Typings">
+                <Directory Id="Dir_$(var.key)InstallLocationScriptsTypings" Name="Typings">                
                     <Directory Id="Dir_$(var.key)ScriptsTypingsNode" Name="Node" />
                 </Directory>
             </Directory>
             <Directory Id="Dir_$(var.key)RemoteDebug" Name="RemoteDebug"/>
+            <Directory Id="Dir_$(var.key)Snippets" Name="Snippets">
+                <Directory Id="Dir_$(var.key)Snippets_1033" Name="1033">
+                <Directory Id="Dir_$(var.key)Snippets_1033_Nodejs" Name="Nodejs"/>
+                <!--<Directory Id="Dir_$(var.key)Snippets_1033_Test" Name="Test"/>-->
```
mousetraps: For test snippets, it would be useful to take a look at Mocha and ExportRunner which are the two test frameworks we support. 

And speaking of tests, we should add some unit tests to verify all the core parts of this experience are working and also make sure we don't break it going forward. :smiley:

Nodejs/Setup/NodejsTools/NodejsToolsFiles.proj
```diff
@@ -105,6 +105,53 @@
...
+                       Snippets\1033\Nodejs\require.snippet;
+                       Snippets\1033\Nodejs\server.snippet;
+                       Snippets\1033\Nodejs\stringify.snippet;
+                       Snippets\1033\Nodejs\switch.snippet;
+                       Snippets\1033\Nodejs\tryc.snippet;
+                       Snippets\1033\Nodejs\tryf.snippet;
+                       Snippets\1033\Nodejs\while.snippet;
+                       Snippets\1033\Nodejs\with.snippet">
+            <InstallDirectory>Snippets_1033_Nodejs</InstallDirectory>
+        </File>
```
mousetraps: Easier than including each of these files individually is...
```xml
<File Include="Snippets\1033\Nodejs\*.snippet">
    <InstallDirectory>Snippets_1033_Nodejs</InstallDirectory>
</File>
```

mousetraps:
Still needs a bit of work, but looking good so far :smiley:. Super excited to bring this in - we've had a bunch of requests for snippets, so our users will be certainly appreciate it:
* http://nodejstools.codeplex.com/workitem/1301
* https://github.com/Microsoft/nodejstools/issues/317
* https://github.com/Microsoft/nodejstools/issues/158
* https://nodejstools.codeplex.com/workitem/1970

And I'll be extra happy because it'll make demos super slick.

Anywho... left some comments inline. A bunch of them center around the core experience, so it's worth taking a look at the following doc for snippets so you can get a better sense of what users will expect (mainly pay attention to how many ways there are of invoking snippets because ideally we would be consistent with those approaches.)
* https://msdn.microsoft.com/en-US/library/6hf704tz.aspx

Nodejs/Setup/NodejsTools/NodejsToolsFiles.proj
avitalb: Yeah, we tried that, but it didn't seem to work

Nodejs/Product/AssemblyVersion.cs
```diff
@@ -1,64 +1,64 @@
...
+// If you get compiler errors CS0579, "Duplicate '<attributename>' attribute", check your
+// Properties\AssemblyInfo.cs file and remove any lines duplicating the ones below.
+// (See also AssemblyInfoCommon.cs in this same directory.)
+
+#if !SUPPRESS_COMMON_ASSEMBLY_VERSION
+[assembly: AssemblyVersion(AssemblyVersionInfo.StableVersion)]
+#endif
+[assembly: AssemblyFileVersion(AssemblyVersionInfo.Version)]
+
+class AssemblyVersionInfo {
```
mousetraps: Mind reverting the whitespace changes in this file, since no other changes were made? It's CRLF now rather than LF as it was before.

Nodejs/Product/Nodejs/Guids.cs
```diff
@@ -27,14 +27,16 @@ static class Guids
         public const string NodejsDebugLanguageString = "{65791609-BA29-49CF-A214-DBFF8AEC3BC2}";
         public const string NodejsEditorFactoryString = "88941496-93F4-4E37-83AF-AFE087415334";
         public const string NodejsEditorFactoryPromptEncodingString = "C8576E92-EFB6-4414-8F63-C84D474A539E";
-        public const string NodejsLanguageInfoString = "ABD5E8A5-5A35-4BE9-BCAF-E10C1212CB40";
+        public const string NodejsLanguageInfoString = "{ABD5E8A5-5A35-4BE9-BCAF-E10C1212CB40}";
```
mousetraps: Worth including a comment here about why it needs to have braces since it's non-obvious.

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,308 @@
...
+//    distributed under the License is distributed on an "AS IS" BASIS, 
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or 
+//    implied. See the License for the specific language governing 
+//    permissions and limitations under the License.
+//
+//*********************************************************//
+using System;
+using System.Collections.Generic;
+using Microsoft.NodejsTools.Analysis;
+using Microsoft.NodejsTools.Parsing;
```
mousetraps: above two references are unnecessary

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,308 @@
...
+        private readonly IVsTextView _view;
+        private readonly ITextView _textView;
+        private readonly IVsEditorAdaptersFactoryService _adapterFactory;
+        private readonly IServiceProvider _serviceProvider;
+        private IVsExpansionSession _session;
+        private bool _sessionEnded, _selectEndSpan;
+        private ITrackingPoint _selectionStart, _selectionEnd;
+
+        public const string SurroundsWith = "SurroundsWith";
+        public const string SurroundsWithStatement = "SurroundsWithStatement";
```
mousetraps: You shouldn't need to use SurroundsWithStatement in JavaScript - this looks like a remnant of the python codebase. Is there a reason it's still here?

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,308 @@
...
+                    }
+                }
+            }
+            // get the indentation of where we're inserting the code...
+            string baseIndentation = GetBaseIndentation(ts);
+
+            TextSpan? endSpan = null;
+            using (var edit = _textView.TextBuffer.CreateEdit()) {
+                if (surroundsWith || surroundsWithStatement) {
+                    var templateText = codeNode.text.Replace("\r \n", VsExtensions.GetNewLineText(_textView.TextSnapshot));
```
mousetraps: Should be ```"\r\n"```, not ```"\r \n"```

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,308 @@
...
+                            _textView.TextBuffer.CurrentSnapshot.GetLineFromPosition(start).LineNumber + 1, // 1st line is already indented
+                            _textView.TextBuffer.CurrentSnapshot.GetLineFromPosition(end).LineNumber
+                        );
+                    }
+                }
+
+                // we now need to update any code which was not selected  that we just inserted.
+                IndentSpan(edit, baseIndentation, ts[0].iStartLine + 1, ts[0].iEndLine);
+
+                edit.Apply();
```
mousetraps: The snippets won't behave very well when different formatting options are supplied. E.g. tabs vs. spaces, or 2 spaces rather than four spaces. If we're lucky, VS might handle things for us if we supply tabs in the snippets rather than spaces, but otherwise we should add some additional handling logic here.

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,308 @@
...
+            }
+
+            if (endSpan != null) {
+                _session.SetEndSpan(endSpan.Value);
+            }
+
+            return hr;
+        }
+        public int InsertNamedExpansion(string pszTitle, string pszPath, TextSpan textSpan) {
```
mousetraps: missing a line before this method

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,308 @@
...
+
+            if (ErrorHandler.Succeeded(hr)) {
+                if (_sessionEnded) {
+                    _session = null;
+                }
+            }
+            return hr;
+        }
+
+
```
mousetraps: nit: extraneous line

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,308 @@
...
+                    return _session.EndCurrentExpansion(1);
+                }
+            }
+            return _session.EndCurrentExpansion(leaveCaret ? 1 : 0);
+        }
+
+        public int PositionCaretForEditing(IVsTextLines pBuffer, TextSpan[] ts) {
+            return VSConstants.S_OK;
+        }
+        private void GetCaretPosition(out int caretLine, out int caretColumn) {
```
mousetraps: nit: missing line before this method

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -105,6 +104,10 @@
     <Reference Include="Microsoft.Build.Utilities.$(MicrosoftBuildAssemblyVersionSuffix), Version=$(MicrosoftBuildAssemblyVersion), Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="Microsoft.JScript" />
+    <Reference Include="microsoft.msxml, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <EmbedInteropTypes>False</EmbedInteropTypes>
+      <Private>False</Private>
```
mousetraps: <private> is unnecessary here.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/cberr.snippet
```diff
@@ -0,0 +1,30 @@
...
+          <ID>e</ID>
+          <Default>e</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[var hadError = function ($e$, response) {
+    console.error($e$);$end$
+    response.end('Server Error');
+};
+]]>
```
mousetraps: Let's remove this snippet - it covers too specific of a scenario, so I don't think it'll get much use.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/cl.snippet
```diff
@@ -0,0 +1,26 @@
...
+    </Header>
+    <Snippet>
+      <Declarations>
+        <Literal>
+          <ID>message</ID>
+          <Default>message</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[console.log("$message$")$end$]]>
```
mousetraps: Semicolon makes sense at the end of this statement.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/dowhile.snippet
```diff
@@ -0,0 +1,32 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<CodeSnippets
+    xmlns="http://schemas.microsoft.com/VisualStudio/2005/CodeSnippet">
+  <CodeSnippet Format="1.0.0">
+    <Header>
+      <Title>dowhile</Title>
+      <Description>Code Snippet for a Do...While Loop</Description>
+      <Shortcut>dowhile</Shortcut>
```
mousetraps: this snippet can be ```do``` rather than ```dowhile``` - it's both shorter and consistent with the rest of the VS javascript experience.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/exports.snippet
```diff
@@ -0,0 +1,31 @@
...
+          <ID>myModule</ID>
+          <Default>myModule</Default>
+        </Object>
+        <Object>
+          <ID>myVar</ID>
+          <Default>myVar</Default>
+        </Object>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[module.exports.$myModule$ = $myVar$;]]>
```
mousetraps: I think this snippet is a bit too specific of a scenario - people assign all sorts of things to module.exports (not just variables, so we would really just want to end at the equals so people can start typing whatever they want), and I don't think we're saving them much time here with ```module.exports.$myModule$ = ```, so let's just remove this snippet.

(take a look at the following query to see how people use module.exports in real code)
https://github.com/search?p=1&q=module.exports&ref=searchresults&type=Code&utf8=%E2%9C%93

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/express_get.snippet
```diff
@@ -0,0 +1,33 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<CodeSnippets
+    xmlns="http://schemas.microsoft.com/VisualStudio/2005/CodeSnippet">
+  <CodeSnippet Format="1.0.0">
+    <Header>
+      <Title>express_get</Title>
+      <Description>Code Snippet for handling a GET request with Express</Description>
+      <Shortcut>express_get</Shortcut>
```
mousetraps: This shortcut isn't very short or easy to type. how about ```expget```,  ```exppost```, ``expput``, and ```expdelete```?

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/for.snippet
```diff
@@ -0,0 +1,30 @@
...
+    </Header>
+    <Snippet>
+      <Declarations>
+        <Literal>
+          <ID>length</ID>
+          <Default>length</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[for (var i = 0; i < $length$; i++) {
```
mousetraps: We should add ```i``` to the declarations as well.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/forin.snippet
```diff
@@ -0,0 +1,24 @@
...
+      <Shortcut>forin</Shortcut>
+      <Author>Microsoft Corporation</Author>
+      <SnippetTypes>
+        <SnippetType>SurroundsWithStatement</SnippetType>
+        <SnippetType>Expansion</SnippetType>
+      </SnippetTypes>
+    </Header>
+    <Snippet>
+      <Code Language="Nodejs">
+        <![CDATA[for (var i in o) {
```
mousetraps: ```i``` and ```o``` should be added as declarations.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/forr.snippet
```diff
@@ -0,0 +1,30 @@
...
+    </Header>
+    <Snippet>
+      <Declarations>
+        <Literal>
+          <ID>length</ID>
+          <Default>length</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[for (var i = $length$ - 1; i >= 0; i--) {
```
mousetraps: ```i``` should also be added to the declarations here.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/props.snippet
```diff
@@ -0,0 +1,36 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<CodeSnippets
+    xmlns="http://schemas.microsoft.com/VisualStudio/2005/CodeSnippet">
+  <CodeSnippet Format="1.0.0">
+    <Header>
+      <Title>props</Title>
+      <Description>Code Snippet for iterating through the properties of an object</Description>
+      <Shortcut>props</Shortcut>
```
mousetraps: ```forprops``` would be more precise here

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/readfile.snippet
```diff
@@ -0,0 +1,34 @@
...
+          <Default>file</Default>
+        </Literal>
+        <Literal>
+          <ID>error</ID>
+          <Default>error</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[fs.readFile('$file$', function ($error$, data) {
+    if ($error$) throw $error$;
```
mousetraps: Let's keep the error names consistent across snippets.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/readstream.snippet
```diff
@@ -0,0 +1,33 @@
...
+    </Header>
+    <Snippet>
+      <Declarations>
+        <Literal>
+          <ID>resource</ID>
+          <Default>resource</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[var stream = fs.createReadStream('$resource$');
```
mousetraps: use ```var readStream = ``` rather than ```var stream =```, as it's more descriptive. Additionally, use $path$ rather than $resource$ as that is actually how the node docs describe the argument to ```createReadStream```

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/readstream.snippet
```diff
@@ -0,0 +1,33 @@
...
+stream.on('data', function (chunk) {
```
mousetraps: more commonly called ```data``` rather than ```chunk```

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/readstream.snippet
```diff
@@ -0,0 +1,33 @@
...
+    console.log(chunk);
+});
+stream.on('end', function () {
+    console.log('finished');
+});
+]]>
```
mousetraps: make sure the cursor ends up in a reasonable spot here.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/require.snippet
```diff
@@ -0,0 +1,26 @@
...
+    </Header>
+    <Snippet>
+      <Declarations>
+        <Literal>
+          <ID>myModule</ID>
+          <Default>myModule</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[var $myModule$ = require('$myModule$');]]>
```
mousetraps: make sure the cursor ends up at the end of the require statement in this snippet.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/switch.snippet
```diff
@@ -0,0 +1,30 @@
...
+    <Snippet>
+      <Declarations>
+        <Literal>
+          <ID>switch_on</ID>
+          <Default>switch_on</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[switch ($switch_on$) {
+    case:$end$
```
mousetraps: I wouldn't bother including the ```case``` here and just end on the line in ```default```.

The reason for this is that we generally don't want the snippets to produce syntax errors, and in this case it would unless the user navigates to before the colon in case, types in their case statement, etc.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/tryc.snippet
```diff
@@ -0,0 +1,31 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<CodeSnippets
+    xmlns="http://schemas.microsoft.com/VisualStudio/2005/CodeSnippet">
+  <CodeSnippet Format="1.0.0">
+    <Header>
+      <Title>tryc</Title>
```
mousetraps: Rarely do people use try without the catch, so calling it ```try``` should be fine here.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/tryc.snippet
```diff
@@ -0,0 +1,31 @@
...
+<CodeSnippets
+    xmlns="http://schemas.microsoft.com/VisualStudio/2005/CodeSnippet">
+  <CodeSnippet Format="1.0.0">
+    <Header>
+      <Title>tryc</Title>
+      <Description>Code Snippet for a Try...Catch Statement</Description>
+      <Shortcut>tryc</Shortcut>
+      <Author>Microsoft Corporation</Author>
+      <SnippetTypes>
+        <SnippetType>Expansion</SnippetType>
```
mousetraps: Make this a SurroundsWith as well - I anticipate that'll be the most likely scenario anyways.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/tryc.snippet
```diff
@@ -0,0 +1,31 @@
...
+          <ID>e</ID>
+          <Default>e</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[try {
+$end$
+} catch ($e$) {
+
+};
```
mousetraps: Once you make it a SurroundsWith snippet, it'll be better to end in the catch block.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/tryf.snippet
```diff
@@ -0,0 +1,33 @@
...
+          <ID>e</ID>
+          <Default>e</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[try {
+$end$
+}catch ($e$){
+
+}finally {
```
mousetraps: add a space before the catch and finally blocks, and also make this a SurroundsWith snippet too.

Nodejs/Setup/NodejsTools/NodejsTools.wxs
```diff
@@ -23,6 +23,12 @@
                 </Directory>
             </Directory>
             <Directory Id="Dir_$(var.key)RemoteDebug" Name="RemoteDebug"/>
+            <Directory Id="Dir_$(var.key)Snippets" Name="Snippets">
+                <Directory Id="Dir_$(var.key)Snippets_1033" Name="1033">
+                <Directory Id="Dir_$(var.key)Snippets_1033_Nodejs" Name="Nodejs"/>
+                <Directory Id="Dir_$(var.key)Snippets_1033_Test" Name="Test"/>
+                </Directory>
```
mousetraps: Fix indentation

Nodejs/Tests/Core.UI/Nodejs.Tests.UI.csproj
```diff
@@ -1,188 +1,191 @@
...
+    <Reference Include="Microsoft.VisualStudioTools.VSTestHost.$(VSTarget), Version=$(VSTarget).1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
+    <Reference Include="Microsoft.VisualStudio.CoreUtility, Version=$(VSTarget).0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
+    <Reference Include="Microsoft.VisualStudio.Language.Intellisense, Version=$(VSTarget).0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
+    <Reference Include="Microsoft.VisualStudio.OLE.Interop, Version=7.1.40304.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
+    <Reference Include="Microsoft.VisualStudio.QualityTools.ExecutionCommon, Version=$(VSTarget).0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
+    <Reference Include="Microsoft.VisualStudio.QualityTools.UnitTestFramework, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <Private>False</Private>
```
mousetraps: unnecessary

Nodejs/Tests/Core.UI/SnippetsTests.cs
```diff
@@ -0,0 +1,228 @@
...
+
+            foreach (var decl in snippet.Declarations) {
+                Console.WriteLine("Declaration: {0}", decl.Replacement);
+                Keyboard.Type(decl.Replacement);
+                server.WaitForText(decl.Expected.Replace("$body$", body));
+                Keyboard.Type("\t");
+            }
+            //
+            Keyboard.Type("\r");
+            //Keyboard.Type(body);
```
mousetraps: remove commented out code

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/forin.snippet
```diff
@@ -0,0 +1,24 @@
...
+   $selected$$end$
```
mousetraps: The spacing in many of the snippets is inconsistent - in this case, there are three spaces for indents, in other cases, there two or four.

mousetraps:
Almost there... :smiley:

Code as a whole is looking better now. Left a bunch of comments - mostly on the snippets themselves - and lmk if you have any questions. 

As a stretch... it would be nice to add the snippets to the completions list, but I'd rather get the bulk of the code checked in sooner rather than later since we're hoping to ship this in the next release.

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,305 @@
...
+            return _session.GoToPreviousExpansionField();
+        }
+
+        public int EndCurrentExpansion(bool leaveCaret) {
+            if (_selectEndSpan) {
+                TextSpan[] endSpan = new TextSpan[1];
+                if (ErrorHandler.Succeeded(_session.GetEndSpan(endSpan))) {
+                    var snapshot = _textView.TextBuffer.CurrentSnapshot;
+                    var startLine = snapshot.GetLineFromLineNumber(endSpan[0].iStartLine);
+                    var span = new Span(startLine.Start + endSpan[0].iStartIndex, 4);
```
mousetraps: This appears to be the root cause of the selection issue you're running into. Ensure ```_selectEndSpan``` is set to ```true``` in the ```surroundWith && !selection.IsNullOrEmpty``` case -- note that this is different than the current condition, which sets ```_selectEndSpan``` to ```true``` when the selection *is* empty. Then, you can set the ```span``` to something that makes sense (right now this method is still targeting Python pass statements.)

Something like the following should work.
```csharp
var selectionLength = _selectionEnd.GetPosition(_textView.TextBuffer.CurrentSnapshot) - _selectionStart.GetPosition(_textView.TextBuffer.CurrentSnapshot);
var span = new Span(startLine.Start + endSpan[0].iStartIndex + selectionLength, 0);
```

Nodejs/Product/Nodejs/Guids.cs
```diff
@@ -27,14 +27,18 @@ static class Guids
         public const string NodejsDebugLanguageString = "{65791609-BA29-49CF-A214-DBFF8AEC3BC2}";
         public const string NodejsEditorFactoryString = "88941496-93F4-4E37-83AF-AFE087415334";
         public const string NodejsEditorFactoryPromptEncodingString = "C8576E92-EFB6-4414-8F63-C84D474A539E";
-        public const string NodejsLanguageInfoString = "ABD5E8A5-5A35-4BE9-BCAF-E10C1212CB40";
+        //do not remove the curly braces. Without curly braces, the interactive window component will fail to load because
+        //the NodejsLanguageInfoString will be compared to the NodejsLanguageInfoString with curlies and they will not match
+        public const string NodejsLanguageInfoString = "{ABD5E8A5-5A35-4BE9-BCAF-E10C1212CB40}";
```
mousetraps: The interactive window issue was a separate issue, actually (MEF). In this case, the issue was that in certain cases language service features (e.g. snippets) would fail to load because some comparisons in native code surround the guid string with curlies, and they'll fail to match unless we also surround the guid string with curlies.

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,310 @@
...
+                            var endText = endLine.GetText().Substring(end - endLine.Start);
+                            if (!String.IsNullOrWhiteSpace(endText)) {
+                                edit.Insert(end, _textView.Options.GetNewLineCharacter());
+                            }
+
+                        }else if (surroundsWith &&
+                          !string.IsNullOrWhiteSpace(_textView.TextBuffer.CurrentSnapshot.GetText(selectedSpan))) {
+                            _selectEndSpan = true;
+
+                        }
```
mousetraps: You can simplify this to be... 
```csharp
if (surroundsWith) {
    if (string.IsNullOrWhiteSpace(_textView.TextBuffer.CurrentSnapshot.GetText(selectedSpan))) {
        // ...
    } else {
        // ...
    }
}
```

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,310 @@
...
+                // we now need to update any code which was not selected  that we just inserted.
+                IndentSpan(edit, baseIndentation, ts[0].iStartLine + 1, ts[0].iEndLine);
+
+                edit.Apply();
+
+            }
+
+            if (endSpan != null) {
+                _session.SetEndSpan(endSpan.Value);
+            }
```
mousetraps: This is unnecessary now, and all references to endSpan can be removed.

Nodejs/Product/Nodejs/Intellisense/ExpansionClient.cs
```diff
@@ -0,0 +1,310 @@
...
+        public int EndCurrentExpansion(bool leaveCaret) {
+            if (_selectEndSpan) {
+                TextSpan[] endSpan = new TextSpan[1];
+                if (ErrorHandler.Succeeded(_session.GetEndSpan(endSpan))) {
+                    var snapshot = _textView.TextBuffer.CurrentSnapshot;
+                    var startLine = snapshot.GetLineFromLineNumber(endSpan[0].iStartLine);
+                    var selectionLength = _selectionEnd.GetPosition(_textView.TextBuffer.CurrentSnapshot) - _selectionStart.GetPosition(_textView.TextBuffer.CurrentSnapshot);
+                    var span = new Span(startLine.Start + endSpan[0].iStartIndex + selectionLength, 0);
+                    _textView.Caret.MoveTo(new SnapshotPoint(snapshot, span.Start));
+                    _textView.Selection.Select(new SnapshotSpan(_textView.TextBuffer.CurrentSnapshot, span), false);
```
mousetraps: since you're not actually selecting anything here, you should be able to simplify this to the ```_textView.Caret.MoveTo``` call with start being... ```startLine.Start + endSpan[0].iStartIndex + selectionLength```

Nodejs/Product/Nodejs/Snippets/1033/Test/mochatest.snippet
```diff
@@ -0,0 +1,37 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<CodeSnippets
+    xmlns="http://schemas.microsoft.com/VisualStudio/2005/CodeSnippet">
+  <CodeSnippet Format="1.0.0">
+    <Header>
+      <Title>mochatest</Title>
+      <Description>Code Snippet for a Unit Test using Mochajs</Description>
+      <Shortcut>mochatest</Shortcut>
```
mousetraps: Let's split this one out into two snippets - ```describe``` and ```it```, and both should be a SurroundWith + Expansion. That way it's easier for people to drop it into their code. 

Don't bother with the require - this is the sort of snippet people might use multiple times in a file, so we don't want to repeat that code.

Nodejs/Product/Nodejs/Snippets/1033/Test/mochatest.snippet
```diff
@@ -0,0 +1,37 @@
...
+          <Default>myFunction</Default>
+        </Literal>
+        <Literal>
+          <ID>message</ID>
+          <Default>should work</Default>
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[var assert = require("assert");
+describe("myFunction", function () {
```
mousetraps: ```myFunction``` and ```should work``` should be literals.

Nodejs/Product/Nodejs/Snippets/1033/Nodejs/server.snippet
```diff
@@ -0,0 +1,33 @@
...
+        </Literal>
+      </Declarations>
+      <Code Language="Nodejs">
+        <![CDATA[var http = require('http');
+var server = http.createServer(function (request, response) {
+    response.writeHead(200, { "Content-Type": "text/plain" });$end$
+    response.end("$message$");
+});
+server.listen(process.env.PORT || 3000);
+console.log("Server is listening");
```
mousetraps: Make sure to keep the quotation marks consistent throughout all the snippets and our templates (single quotes)

Nodejs/Product/Nodejs/Snippets/1033/Test/mochatest.snippet
```diff
@@ -0,0 +1,37 @@
...
+    it("should work", function () {
```
mousetraps: single quotes, not double quotes for strings to be consistent.

mousetraps:
Left some minor comments, and lgtm after that! :smiley:

mousetraps:
:+1:

mousetraps:
and it's in!

PR #348, opened 2015-08-03
Fix #271 Add optional debugger logging

### LINKED ISSUE #271: Add optional debugger logging
This will help us track down race various debugging oddities that people are running into.

mousetraps:
- Added live diagnostics log pane
- Added new live diagnostic option
- enable live diagnostics on debugger cmds and events

image

### DIFF
--- a/Nodejs/Product/Nodejs/Commands/DiagnosticsForm.Designer.cs
+++ b/Nodejs/Product/Nodejs/Commands/DiagnosticsForm.Designer.cs
@@ -28,6 +28,7 @@
             this._ok = new System.Windows.Forms.Button();
             this.tableLayoutPanel1 = new System.Windows.Forms.TableLayoutPanel();
             this._copy = new System.Windows.Forms.Button();
+            this._diagnosticLoggingCheckbox = new System.Windows.Forms.CheckBox();
             this.tableLayoutPanel1.SuspendLayout();
             this.SuspendLayout();
             // 
@@ -35,13 +36,13 @@
             // 
             this.tableLayoutPanel1.SetColumnSpan(this._textBox, 3);
             this._textBox.Dock = System.Windows.Forms.DockStyle.Fill;
-            this._textBox.Location = new System.Drawing.Point(6, 3);
-            this._textBox.Margin = new System.Windows.Forms.Padding(6, 3, 6, 3);
+            this._textBox.Location = new System.Drawing.Point(8, 4);
+            this._textBox.Margin = new System.Windows.Forms.Padding(8, 4, 8, 4);
             this._textBox.Multiline = true;
             this._textBox.Name = "_textBox";
             this._textBox.ReadOnly = true;
             this._textBox.ScrollBars = System.Windows.Forms.ScrollBars.Vertical;
-            this._textBox.Size = new System.Drawing.Size(812, 464);
+            this._textBox.Size = new System.Drawing.Size(1083, 569);
             this._textBox.TabIndex = 0;
             // 
             // _ok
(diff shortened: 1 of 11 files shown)


Nodejs/Product/Nodejs/Debugger/DebugEngine/AD7Engine.cs
```diff
@@ -482,8 +483,8 @@ public sealed class AD7Engine : IDebugEngine2, IDebugEngineLaunch2, IDebugProgra
         // in which case Visual Studio uses the IDebugEngineLaunch2::LaunchSuspended method
         // The IDebugEngineLaunch2::ResumeProcess method is called to start the process after the process has been successfully launched in a suspended state.
         int IDebugEngineLaunch2.LaunchSuspended(string pszServer, IDebugPort2 port, string exe, string args, string dir, string env, string options, enum_LAUNCH_FLAGS launchFlags, uint hStdInput, uint hStdOutput, uint hStdError, IDebugEventCallback2 ad7Callback, out IDebugProcess2 process) {
-            DebugWriteLine("--------------------------------------------------------------------------------");
-            DebugWriteLine("AD7Engine LaunchSuspended Called with flags '{0}' ({1})", launchFlags, GetHashCode());
+            LiveLogger.WriteLine("--------------------------------------------------------------------------------");
```
billti: Why the row of dashes above this particular emit (and none of the others)? Seems unnecessary.

Nodejs/Product/Nodejs/Logging/LiveLogger.cs
```diff
@@ -0,0 +1,98 @@
...
+                            _instance = new LiveLogger();
+                        }
+                    }
+                }
+                return _instance;
+            }
+        }
+
+        public static void WriteLine(string message, Type category) {
+#if !DEBUG
```
billti: What is the value in all these preprocessor directives? Seems if you remove all of them, it still results in a Debug.WriteLine on DEBUG builds, and message to the Output pane on all builds if IsLiveDiagnosticsEnabled - which is what is desired right?

Nodejs/Product/Nodejs/Commands/DiagnosticsForm.Designer.cs
```diff
@@ -50,12 +51,12 @@ partial class DiagnosticsForm {
...
-            this._ok.Margin = new System.Windows.Forms.Padding(6, 3, 6, 3);
-            this._ok.MinimumSize = new System.Drawing.Size(86, 24);
+            this._ok.Location = new System.Drawing.Point(976, 581);
+            this._ok.Margin = new System.Windows.Forms.Padding(8, 4, 8, 4);
+            this._ok.MinimumSize = new System.Drawing.Size(115, 30);
             this._ok.Name = "_ok";
-            this._ok.Size = new System.Drawing.Size(86, 24);
-            this._ok.TabIndex = 1;
+            this._ok.Size = new System.Drawing.Size(115, 30);
+            this._ok.TabIndex = 3;
```
billti: Do you have a screen shot you can put in the conversation of how the layout appears with these changes?

Nodejs/Product/Nodejs/Logging/LiveLogger.cs
mousetraps: If IsLiveDiagnostics is not enabled, but we are in a DEBUG build we still want Debug.WriteLine to occur.

Nodejs/Product/Nodejs/Debugger/DebugEngine/AD7Engine.cs
mousetraps: I agree it's a little odd to have the lines on this event in particular (we should probably include them at the start of the debug session), but but having lines somewhere does help with readability because it segments debug sessions.

Nodejs/Product/Nodejs/Logging/LiveLogger.cs
billti: All calls end up in LogMessage to write to the diagnostics pan, and this already checks IsLiveDiagnosticsEnabled before doing so, so if you remove all the other `if (_diagnosticsOptions.IsLiveDiagnosticsEnabled)` checks, and all the `#if !DEBUG` directives, I believe it still does exactly what you want (and the code is a lot cleaner).

mousetraps: Yeah, agreed - I was trying to over-optimize to catch the flag as soon as possible

Nodejs/Product/Nodejs/Commands/DiagnosticsForm.Designer.cs
mousetraps: Sure, just added the screenshot to the original description.

mousetraps:
thx for the comments, just updated w/ a new iteration

billti:
:+1:

PR #349, opened 2015-08-03
Fix #334 Test explorer view is messed up in VS 2015

### LINKED ISSUE #334: Test Explorer view is messed up in VS 2015
Not sure if this issue should be addressed to the NodeJS team or not ... The Test Explorer view in VS 2015 is messed up when using mocha.

If you have nested describe() then only top level is shown and the rest gets ignored, which makes it nearly impossible to test in VS in large projects. 

For the example depicted below I was expecting to see the following in Test Explorer: 

spread.dal.mongo
--> index #get() should return an object of DataProvider type

Instead I only see this 

spread.dal.mongo
--> should return an object of DataProvider type

This is only an issue in VS 2015, VS 2013 seems to work fine.

test_code

test_explorer

mousetraps:
- nested tests were not appearing in Test Explorer discovery because we
  were returning title rather than fullTitle()


billti:
:+1:

PR #350, opened 2015-08-03
Fix #228 Analyzer not respecting "None" analysis setting

### LINKED ISSUE #228: Analyzer not respecting "None" analysis setting
Currently investigating the heap dump @NoelAbrahams sent us

mousetraps:
- verify that the analysis setting is not set to None for implicit
  projects before enqueuing

### DIFF
--- a/Nodejs/Product/Nodejs/Intellisense/VsProjectAnalyzer.cs
+++ b/Nodejs/Product/Nodejs/Intellisense/VsProjectAnalyzer.cs
@@ -769,7 +769,7 @@ namespace Microsoft.NodejsTools.Intellisense {
             }
 #endif
 
-            if (_implicitProject) {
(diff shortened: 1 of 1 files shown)


billti:
:+1:

Nodejs/Product/Nodejs/Intellisense/VsProjectAnalyzer.cs
```diff
@@ -769,7 +769,7 @@ sealed partial class VsProjectAnalyzer : IDisposable {
             }
 #endif
 
-            if (_implicitProject) {
+            if (_implicitProject && _analysisLevel != AnalysisLevel.None) {
```
mousetraps: Moving this check closer to the Enqueue below after chatting with Dino, as not executing _projectFiles.TryRemove could potentially keep us from cleaning things up.

PR #360, opened 2015-08-10
First set of change containing telemetry infrastructure for NTVS

nareshjo:
@mousetraps @ryan-dawson


Nodejs/Prerequisites/generate.js
```diff
@@ -39,7 +39,8 @@ var assemblies = [
     'Microsoft.NodejsTools.WebRole',
     'Microsoft.NodejsTools.Npm',
     'Microsoft.NodejsTools.PressAnyKey',
-    'Microsoft.NodejsTools.VsLogger'
+    'Microsoft.NodejsTools.VsLogger',
+	'Microsoft.NodejsTools.Telemetry'
```
paulvanbrenk: nit: these should be alligned

Nodejs/Product/Telemetry/DefaultImpl/DefaultTelemetryLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+//*********************************************************//
+
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.NodejsTools.Telemetry {
+    /// <summary>
+    /// Dummy telemetry logger to be used if no other logger is found.
+    /// </summary>
+    public class DefaultTelemetryLogger : ITelemetryLogger {
```
paulvanbrenk: Can you name this 'DummyTelemetryLogger' to make it clear what it's doing

Nodejs/Product/Telemetry/Shared/ITelemetryLogger.cs
```diff
@@ -0,0 +1,55 @@
...
+        void ReportEvent(string eventName);
+
+        /// <summary>
+        /// Records event with single property
+        /// </summary>
+        /// <param name="eventName">Event Name</param>
+        /// <param name="propertyName">Property Name</param>
+        /// <param name="propertyValue">Property Value</param>
+        /// <param name="isPersonallyIdentifiable">Whether property has personally identifiable information</param>
+        void ReportEvent(string eventName, string propertyName, object propertyValue, bool isPersonallyIdentifiable = false);
```
paulvanbrenk: Default value on an interface method doesn't do anything.

Nodejs/Product/Telemetry/TelemetryEvents.cs
```diff
@@ -0,0 +1,27 @@
...
+//    implied. See the License for the specific language governing 
+//    permissions and limitations under the License.
+//
+//*********************************************************//
+
+namespace Microsoft.NodejsTools.Telemetry {
+    /// <summary>
+    /// Telemetry event names
+    /// </summary>
+    public class TelemetryEvents {
```
paulvanbrenk: make the class static since it only contains static members

Nodejs/Product/Telemetry/TelemetryProperties.cs
```diff
@@ -0,0 +1,28 @@
...
+//    implied. See the License for the specific language governing 
+//    permissions and limitations under the License.
+//
+//*********************************************************//
+
+namespace Microsoft.NodejsTools.Telemetry {
+    /// <summary>
+    /// Property names for telemetry events
+    /// </summary>
+    public class TelemetryProperties {
```
paulvanbrenk: make the class static since it only contains static members

Nodejs/Product/Telemetry/TelemetrySetup.cs
```diff
@@ -0,0 +1,70 @@
...
+//    
+//    Unless required by applicable law or agreed to in writing, software 
+//    distributed under the License is distributed on an "AS IS" BASIS, 
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or 
+//    implied. See the License for the specific language governing 
+//    permissions and limitations under the License.
+//
+//*********************************************************//
+
+using Microsoft.VisualStudio.ComponentModelHost;
```
paulvanbrenk: nit: System.* namespaces should be sorted first

Nodejs/Product/Telemetry/Shared/ITelemetryLogger.cs
nareshjo: This will make it convenient (without adding another override to each implementation of this interface) for callers which use this interface as the compile time type.

paulvanbrenk: I didn't know that.

Common/Product/SharedProject/ProjectNode.cs
```diff
@@ -1575,9 +1577,11 @@ internal enum EventTriggering {
                             }
 
                             FinishProjectCreation(basePath, baseLocation);
+                            this.LogProjectLoadTelemetry(TelemetryEvents.ProjectCreated);
```
mousetraps: You shouldn't be adding anything to Common/*. This code is shared with other projects, like python tools. Instead, add to the classes that extend these ones (e.g. ```NodejsProjectNode```):
https://github.com/microsoft/visualstudio-sharedproject

Common/Product/SharedProject/ProjectNode.cs
```diff
@@ -1593,6 +1597,16 @@ internal enum EventTriggering {
...
         }
 
+        private void LogProjectLoadTelemetry(string eventName) {
+            DataPointCollection properties = new DataPointCollection();
+            // Enclose guid in braces
+            properties.Add(TelemetryProperties.ProjectGuid, this.ProjectIDGuid.ToString("B"));
+            properties.Add(TelemetryProperties.ProjectFlavorGuid, this.ProjectGuid.ToString("B"));
+            properties.Add(TelemetryProperties.ProjectType, this.ProjectType);
+
+            NodejsPackage.Instance.TelemetryLogger.ReportEvent(eventName, properties);
```
mousetraps: The properties set here seem like they should be set pretty much every time we log telemetry involving a project, so there should be a more centralized method for it.

Nodejs/Prerequisites/generate.js
```diff
@@ -39,7 +39,8 @@ var assemblies = [
     'Microsoft.NodejsTools.WebRole',
     'Microsoft.NodejsTools.Npm',
     'Microsoft.NodejsTools.PressAnyKey',
-    'Microsoft.NodejsTools.VsLogger'
+    'Microsoft.NodejsTools.VsLogger',
```
mousetraps: remove VsLogger here too.

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -1471,7 +1470,7 @@
     </None>
   </ItemGroup>
   <ItemGroup>
-    <Content Include="..\Icons\SystemRegisteredICO\NodeJS.ico">
+  <Content Include="..\Icons\SystemRegisteredICO\NodeJS.ico">
```
mousetraps: nit: fix indentation here

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -1531,6 +1530,11 @@
       <IncludeOutputGroupsInVSIX>BuiltProjectOutputGroup%3bBuiltProjectOutputGroupDependencies%3bGetCopyToOutputDirectoryItems%3bSatelliteDllsProjectOutputGroup%3b</IncludeOutputGroupsInVSIX>
       <IncludeOutputGroupsInVSIXLocalOnly>DebugSymbolsProjectOutputGroup%3b</IncludeOutputGroupsInVSIXLocalOnly>
     </ProjectReference>
+    <ProjectReference Include="..\Telemetry\Telemetry.csproj">
+      <Project>{d6fcbc60-f21e-4854-ac51-aed1a0a7c868}</Project>
+      <Name>Telemetry</Name>
+      <IncludeInVSIX>true</IncludeInVSIX>
+    </ProjectReference>
```
mousetraps: Should we consider adding #ifdefs since this dll only works for Dev14+?

Nodejs/Product/Nodejs/NodejsPackage.cs
```diff
@@ -248,6 +252,23 @@ internal sealed partial class NodejsPackage : CommonPackage {
...
+                var versionInfo = FileVersionInfo.GetVersionInfo(typeof(NodejsPackage).Assembly.Location);
+                // NTVS version
+                _telemetryLogger.SetCommonProperty(new DataPoint(TelemetryProperties.NtvsVersion, versionInfo.ProductVersion.ToString()));
+                // VS version
+                _telemetryLogger.SetCommonProperty(new DataPoint(TelemetryProperties.HostVersion, Application.ProductVersion.ToString()));
+
+                // Log package loaded event
+                _telemetryLogger.ReportEvent(TelemetryEvents.PackageLoaded);
+            }
+        }
```
mousetraps: We will also need this code when we are loading the Profiling and InteractiveWindow packages, so we should extract it out into a common method that accepts the assembly name, version info, and vs version.

Nodejs/Product/Telemetry/DummyImpl/DummyTelemetryLogger.cs
```diff
@@ -0,0 +1,48 @@
...
+        public void ReportEvent(string eventName, string propertyName, object propertyValue, bool isPersonallyIdentifiable = false) {
+            Debug.WriteLine("Telemetry event posted: " + eventName);
+        }
+
+        public void ReportEvent(string eventName, DataPointCollection properties) {
+            Debug.WriteLine("Telemetry event posted: " + eventName);
+        }
+
+        public void SetCommonProperty(DataPoint property) {
+            Debug.WriteLine(string.Format("Set common property '{0}' to '{1}'", property.Name, property.Value.ToString()));
```
mousetraps: ```ToString()``` is implicit when you use ```String.Format()```

Nodejs/Product/Telemetry/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Node.js Tools for Visual Studio - Telemetry")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+
```
mousetraps: Also set ComVisible to false here and include the following comment.
```csharp
// Setting ComVisible to false makes the types in this assembly not visible 
// to COM components.  If you need to access a type in this assembly from 
// COM, set the ComVisible attribute to true on that type.
[assembly: ComVisible(false)]
```

Nodejs/Product/Telemetry/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,31 @@
...
+[assembly: InternalsVisibleTo("NodeTests, PublicKey=002400000480000094000000060200000024000052534131000400000100010007d1fa57c4aed9f0a32e84aa0faefd0de9e8fd6aec8f87fb03766c834c99921eb23be79ad9d5dcc1dd9ad236132102900b723cf980957fc4e177108fc607774f29e8320e92ea05ece4e821c0a5efe8f1645c4c0c93c1ab99285d622caa652c1dfad63d745d6f2de5f17e5eaf0fc4963d261c8a12436518206dc093344d5ad293")]
+[assembly: InternalsVisibleTo("Nodejs.Tests.UI, PublicKey=002400000480000094000000060200000024000052534131000400000100010007d1fa57c4aed9f0a32e84aa0faefd0de9e8fd6aec8f87fb03766c834c99921eb23be79ad9d5dcc1dd9ad236132102900b723cf980957fc4e177108fc607774f29e8320e92ea05ece4e821c0a5efe8f1645c4c0c93c1ab99285d622caa652c1dfad63d745d6f2de5f17e5eaf0fc4963d261c8a12436518206dc093344d5ad293")]
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
```
mousetraps: These are all unnecessary.

Nodejs/Product/Telemetry/Shared/DataPointCollection.cs
```diff
@@ -0,0 +1,34 @@
...
+//*********************************************************//
+
+using System.Collections.Generic;
+
+namespace Microsoft.NodejsTools.Telemetry {
+    /// <summary>
+    /// DataPointCollection - Represents a collection of DataPoints to report for telemetry
+    /// </summary>
+    public class DataPointCollection : List<DataPoint> {
+        public DataPointCollection() : base() {
```
mousetraps: It would be ideal to have a constructor that takes in DataPoint params here to get away from the following repetition everywhere upon creation.
```
collection.Add(...)
collection.Add(...)
```

Nodejs/Product/Telemetry/Shared/DataPointCollection.cs
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        public DataPointCollection(int capacity) : base(capacity) {
+        }
+
+        public void Add(string name, object value, bool isPersonallyIdentifiable = false) {
```
mousetraps: ```Add(params DataPoint[] datapoints)``` would also be convenient.

Nodejs/Product/Telemetry/Shared/ITelemetryLogger.cs
```diff
@@ -0,0 +1,55 @@
...
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or 
+//    implied. See the License for the specific language governing 
+//    permissions and limitations under the License.
+//
+//*********************************************************//
+
+using System;
+
+namespace Microsoft.NodejsTools.Telemetry {
+    public interface ITelemetryLogger : IDisposable {
```
mousetraps: Why the IDisposable? The Dispose implementation is empty and this isn't being disposed anywhere...

Nodejs/Product/Telemetry/Telemetry.csproj
```diff
@@ -0,0 +1,106 @@
...
+        <UpgradeBackupLocation>
+        </UpgradeBackupLocation>
+        <OldToolsVersion>4.0</OldToolsVersion>
+      </PropertyGroup>
+    </When>
+  </Choose>
+  <Import Project="..\ProjectBefore.settings" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">x86</Platform>
```
mousetraps: Is there any reason this is targetting x86 rather than AnyCPU? Some of our assemblies that would need to take a dependency on this component are AnyCPU (e.g. TestAdapter) and should stay that way.

You can verify that none of the dependencies are compiled for 32-bit only by running ```CorFlags.exe assembly.dll``` and checking whether the 32bit flag is set to 1.

Nodejs/Product/Telemetry/Telemetry.csproj
```diff
@@ -0,0 +1,106 @@
...
+    <ErrorReport>prompt</ErrorReport>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|x86'">
+    <PlatformTarget>x86</PlatformTarget>
+    <DebugType>pdbonly</DebugType>
+    <DefineConstants>$(DefineConstants);TRACE</DefineConstants>
+    <Optimize>true</Optimize>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <ErrorReport>prompt</ErrorReport>
+  </PropertyGroup>
```
mousetraps: Are the above two property groups necessary?

Nodejs/Product/Telemetry/Telemetry.csproj
```diff
@@ -0,0 +1,106 @@
...
+  <PropertyGroup>
+    <SignAssembly>true</SignAssembly>
+  </PropertyGroup>
```
mousetraps: What's this for?

Nodejs/Product/Telemetry/Telemetry.csproj
```diff
@@ -0,0 +1,106 @@
...
+  <PropertyGroup>
+    <StartupObject />
+  </PropertyGroup>
```
mousetraps: And this?

Nodejs/Product/Telemetry/Telemetry.csproj
```diff
@@ -0,0 +1,106 @@
...
+    <Reference Include="System.Xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="DummyImpl\DummyTelemetryLogger.cs" />
+    <Compile Include="DummyImpl\DummyTelemetryFactory.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="Shared\DataPoint.cs" />
+    <Compile Include="Shared\DataPointCollection.cs" />
+    <Compile Include="Shared\ITelemetryFactory.cs" />
+    <Compile Include="Shared\ITelemetryLogger.cs" />
```
mousetraps: What are these all being "shared" with?

Nodejs/Product/Telemetry/TelemetrySetup.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Linq;
+using Microsoft.VisualStudio.ComponentModelHost;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Utilities;
+
+namespace Microsoft.NodejsTools.Telemetry {
+    /// <summary>
+    /// Chooses the best telemetry provider to use out of available ones
+    /// </summary>
+    public class TelemetrySetup : IPartImportsSatisfiedNotification {
```
mousetraps: This looks like it should also implement ITelemetryFactory

Nodejs/Product/Telemetry/TelemetrySetup.cs
```diff
@@ -0,0 +1,70 @@
...
+                }
+                // Use dummy one if no telemetry provider was found
+                return _telemetryFactory ?? (_telemetryFactory = new DummyTelemetryFactory());
+            }
+        }
+
+        public void OnImportsSatisfied() {
+            try {
+                // Pick the top most one
+                var lazyFActory = Orderer.Order(_unOrderedTelemetryFactoryExports).FirstOrDefault();
```
mousetraps: ```lazyFactory```, not ```lazyFActory```

Nodejs/Setup/BuildRelease.ps1
```diff
@@ -243,7 +243,8 @@ $managed_files = (
     "Microsoft.NodejsTools.Npm.dll",
     "Microsoft.NodejsTools.TestAdapter.dll",
     "Microsoft.NodejsTools.PressAnyKey.exe",
-    "Microsoft.NodejsTools.VsLogger.dll"
+    "Microsoft.NodejsTools.VsLogger.dll",
```
mousetraps: Let's go ahead and remove Microsoft.Nodejstools.VsLogger references while we're at it.

Nodejs/Setup/NodejsTools/NodejsToolsFiles.proj
```diff
@@ -17,6 +17,7 @@
         <File Include="Microsoft.NodejsTools.ProjectWizard.dll" />
         <File Include="Microsoft.NodejsTools.Analysis.dll" />
         <File Include="Microsoft.NodejsTools.TestAdapter.dll" />
+        <File Include="Microsoft.NodejsTools.Telemetry.dll" />
```
mousetraps: Is there anything else that needs to be added to the installer here? (Aren't we dependent on an external dll in the event that we don't fall back to Dummy?)

nareshjo: That will be part of closed repo. None of the closed piece will show in open repo. So if someone builds a setup out of open repo it will end up using DummyTelemetryLogger and won't log anything.

Nodejs/Product/Telemetry/DummyImpl/DummyTelemetryFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System.ComponentModel.Composition;
+using Microsoft.VisualStudio.Utilities;
+
+namespace Microsoft.NodejsTools.Telemetry {
+    /// <summary>
+    /// Default, dummy <see cref="ITelemetryFactory"/> implementation.
+    /// It's intended to be used when no other real implementation of ITelemetryFactory is found
+    /// </summary>
+    [Export(typeof(ITelemetryFactory))]
```
mousetraps: Why are we exporting this here? Won't it be picked up on the import even if the real one is there?

Nodejs/Product/Nodejs/Nodejs.csproj
nareshjo: This is a common dll independent o VS version. The dlls which log actual telemetry by implementing interfaces from this one are (will be) part of closed repo. That one will have #ifdef checks

Nodejs/Product/Telemetry/DummyImpl/DummyTelemetryFactory.cs
nareshjo: It won't cause any issues as the concrete implementations will be ordered in a way that they get picked up first. But I agree this export is still redundant. I'll remove it.

    [Export(typeof(ITelemetryFactory))]
    [Name("NodeJs AppInsights Telemetry Factory")]
    [Order(Before = "NodeJs Dummy Telemetry Factory")]

Common/Product/SharedProject/ProjectNode.cs
nareshjo: Removing this telemetry point altogether as existing VS telemetry already has it covered at a more general level and with more rich data.

Nodejs/Setup/NodejsTools/NodejsToolsFiles.proj
mousetraps: The setup code itself should remain the same in both repos, though. Otherwise we have to deal with keeping the installers in sync.

Common/Product/SharedProject/ProjectNode.cs
nareshjo: Removed this part as mentioned in above comment. Also, existing VS telemetry has ActiveProject property to track projectid of current project when event was sent. From projectGuid we can get these extra properties on our data backend by simple joins.

Nodejs/Product/Telemetry/Shared/ITelemetryLogger.cs
nareshjo: I want the actual implementations of ITelemetryLogger to be notified when app exits. I have to find right place to call it from but didn't want to block checkin on it. I'll add that in a later checkin.
Dispose implementation is empty right now in this PR as it is a dummy logger. Other actual logger in internal code might do some cleanup like flushing data at end. If it makes more sense I add it later when it is really needed by some logger implementation then I can do that but I thought it would be better not to change shared interfaces much.

Nodejs/Product/Telemetry/Telemetry.csproj
nareshjo: good catch. changing to anycpu

Nodejs/Product/Telemetry/Telemetry.csproj
nareshjo: They are present by default when project is created by VS. To be consistent with projects in NTVS I am removing them.

Nodejs/Product/Telemetry/Telemetry.csproj
nareshjo: not needed. removing.

Nodejs/Product/Telemetry/Telemetry.csproj
nareshjo: removed. VS added them somehow

Nodejs/Product/Telemetry/Telemetry.csproj
nareshjo: The items under shared folder will be referenced in closed component. These are either the interfaces which will be implemented by closed code or classes which will be used as parameters in these interfaces.

Nodejs/Setup/BuildRelease.ps1
nareshjo: removed vslogger references everywhere in repo

nareshjo:
After some discussions we are proceeding with another approach significantly different from this one so closing this out and will send new PR with it

PR #366, opened 2015-08-13
ES6 Intellisense preview

jramsay:
- Changes to add an "ES6 IntelliSense Preview" option under:
  Tools->Options->Text Editor->Node.js->IntelliSense
- The option is only enabled if the TS version is at least 1.6
- If the intellisense preview option is enabled the TypeScript editor is used for JS files (instead of the NodeJS editor)


Nodejs/Product/Nodejs/NodejsProject.cs
```diff
@@ -574,11 +572,16 @@ out frame
...
+            return NodejsPackage.Instance.IntellisenseOptionsPage.AnalysisLevel == Options.AnalysisLevel.Preview;
+        }
+
         #region IVsProject Members
 
         public int AddItem(uint itemidLoc, VSADDITEMOPERATION dwAddItemOperation, string pszItemName, uint cFilesToOpen, string[] rgpszFilesToOpen, IntPtr hwndDlgOwner, VSADDRESULT[] pResult) {
-            if (_innerProject3 != null && IsJavaScriptFile(pszItemName)) {
-                Guid ourEditor = typeof(NodejsEditorFactory).GUID;
+            if (_innerProject3 != null && IsJavaScriptFile(pszItemName)) {
+                Guid ourEditor = this.IsES6IntellisensePreview() ? Guid.Empty : typeof(NodejsEditorFactory).GUID;
```
billti: How does returning Guid.Empty tell it to use the TypeScript language service for JavaScript files (as opposed to just the regular JSLS editor for JavaScript files)?

Nodejs/Product/Nodejs/Guids.cs
```diff
@@ -57,6 +58,7 @@ static class Guids
         public static readonly Guid NodejsDebugLanguage = new Guid(NodejsDebugLanguageString);
         public static readonly Guid NodejsNpmCmdSet = new Guid(NodejsNpmCmdSetString);
         public static readonly Guid TypeScriptDebugLanguage = new Guid(TypeScriptLanguageInfoString);
+        public static readonly Guid TypeScriptEditorFactory = new Guid(TypeScriptEditorFactoryString);
```
billti: I see TypeScriptEditorFactoryString and TypeScriptEditorFactory defined here, but what for? I don't see anywhere else in this change they are referenced.

Nodejs/Product/Nodejs/NodejsProject.cs
jramsay: The TSLS editor registers the js file extension at a higher priority than the JSLS editor so it will be the first to process the request.

Nodejs/Product/Nodejs/Guids.cs
jramsay: Thanks - will remove - these are not required

Nodejs/Product/Nodejs/Intellisense/VsProjectAnalyzer.cs
```diff
@@ -150,12 +150,16 @@ sealed partial class VsProjectAnalyzer : IDisposable {
 
         private void CreateNewAnalyzer(AnalysisLimits limits) {
             _jsAnalyzer = new JsAnalyzer(limits);
-            if (_analysisLevel != AnalysisLevel.None) {
+            if (this.ShouldAnalyze()) {
```
mousetraps: Rather than ShouldAnalyze(), use ShouldEnqueue(), which is more precise. Also don't use ```this.``` to be consistent with the rest of the code here.

Nodejs/Product/Nodejs/NodejsProject.cs
```diff
@@ -545,8 +543,8 @@ out project
             Guid ourEditor;
             var properties = GetExtensionObject(_innerVsHierarchy, selectionItemId).Properties;
             try {
-                var itemType = properties.Item("ItemType").Value;
-                ourEditor = (string)itemType == ProjectFileConstants.Compile ? Guids.NodejsEditorFactory : Guid.Empty;
+                var itemType = properties.Item("ItemType").Value;
+                ourEditor = (string)itemType == ProjectFileConstants.Compile && !this.IsES6IntellisensePreview() ? Guids.NodejsEditorFactory : Guid.Empty;
```
mousetraps: remove```this``` here too (and everywhere else)

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.resx
```diff
@@ -1,865 +1,884 @@
...
+  <data name="_analysisLogMax.Items4" xml:space="preserve">
+    <value>100000</value>
+  </data>
+  <data name="_analysisLogMax.Items5" xml:space="preserve">
+    <value>1000000</value>
+  </data>
+  <data name="_analysisLogMax.Items6" xml:space="preserve">
+    <value>Max</value>
+  </data>
+  <data name="
```
mousetraps: Any reason the margins on all the tableLayoutPanels changed here? Is it still consistent with the locations of elements on other pages?

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsPage.cs
```diff
@@ -148,6 +169,41 @@ public NodejsIntellisenseOptionsPage()
             SaveString(CompletionCommittedBySetting, CompletionCommittedBy);
             SaveBool(SaveToDiskSetting, SaveToDisk);
 
+        }
+
+        private string _toolsVersion;
```
mousetraps: Put this at the top of the file with the other private members

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsPage.cs
```diff
@@ -148,6 +169,41 @@ public NodejsIntellisenseOptionsPage()
...
+                    // Start with the Nodejs tools path to retrieve the IDE path
+                    var idePath = Assembly.GetExecutingAssembly().Location;
+                    while (!string.IsNullOrEmpty(idePath) && !StringComparer.OrdinalIgnoreCase.Equals(Path.GetFileName(idePath), "IDE")) {
+                        var parentInfo = Directory.GetParent(idePath);
+                        idePath = parentInfo != null ? parentInfo.FullName : string.Empty;
+                    }
+
+                    if (string.IsNullOrEmpty(idePath)) {
+                        return _toolsVersion;
+                    }
```
mousetraps: This method won't work in the experimental hive because the assembly will be in a different location. Instead use ```VSSPROPID_InstallDirectory``` like the following to get the VS installation directory.
```cs
object installDirAsObject = null;
IVsShell shell = NodejsPackage.Instance.GetService(typeof(SVsShell)) as IVsShell;
if (shell != null) {
    shell.GetProperty((int)__VSSPROPID.VSSPROPID_InstallDirectory, out installDirAsObject);
}
idePath = CommonUtils.NormalizeDirectoryPath((string)objInstallationDir) ?? string.Empty;
```

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsPage.cs
```diff
@@ -148,6 +169,41 @@ public NodejsIntellisenseOptionsPage()
...
+
+                    var regex = new Regex(@"toolsVersion = ""(?<version>\d.\d?)"";");
+                    var fileText = File.ReadAllText(typeScriptServicesPath);
+                    var match = regex.Match(fileText);
+
+                    var version = match.Groups["version"].Value;
+                    if (!string.IsNullOrWhiteSpace(version)) {
+                        _toolsVersion = version;
+                    }
+                } catch (Exception) { }
```
mousetraps: What exception are you expecting here? Debug.WriteLine might be better here rather than completely swallowing the exception.

Also don't swallow any critical exceptions like OutOfMemory, etc. You can use ```ExceptionExtensions.IsCriticalException``` here

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -540,7 +540,7 @@ public NodejsProjectNode(NodejsProjectPackage package)
             }
 
             var relativeFile = CommonUtils.GetRelativeFilePath(this.FullPathToChildren, fileNode.Url);
-            if (this._analyzer.Project != null && this._analyzer.Project.Limits.IsPathExceedNestingLimit(relativeFile)) {
+            if (this._analyzer != null && this._analyzer.Project != null && this._analyzer.Project.Limits.IsPathExceedNestingLimit(relativeFile)) {
```
mousetraps: We've actually already taken this change, so I presume this is why we're seeing merge conflicts. Things should work a little better if you remove this file from the commit. 
https://github.com/Microsoft/nodejstools/blob/master/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs#L581-L584

Nodejs/Product/Nodejs/Intellisense/VsProjectAnalyzer.cs
jramsay: Sounds good - done

Nodejs/Product/Nodejs/NodejsProject.cs
jramsay: done

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsPage.cs
jramsay: done

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsPage.cs
jramsay: Thanks - will use this method instead.

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsPage.cs
jramsay: File ReadAllText could throw a number of exceptions (DirectoryNotFound, IOException, UnauthorizedAccess, etc).  I like your ideas above - will add IsCriticalException and Debug.WriteLine.

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.resx
jramsay: This changed was a result of using the designer.  I've ghosted the before and after screenshots and the only spacing difference I'm noticing is due to adding the new radio button.  The rest looks consistent with the before image.

Nodejs/Product/Nodejs/NodejsProject.cs
```diff
@@ -1,1011 +1,1014 @@
...
+                            prgCmds[i].cmdf = (uint)(OLECMDF.OLECMDF_INVISIBLE | OLECMDF.OLECMDF_SUPPORTED | OLECMDF.OLECMDF_DEFHIDEONCTXTMENU | OLECMDF.OLECMDF_ENABLED);
+                            return VSConstants.S_OK;
+                    }
+                }
+            } else if (pguidCmdGroup == VSConstants.VSStd2K) {
+                for (int i = 0; i < prgCmds.Length; i++) {
+                    switch ((VSConstants.VSStd2KCmdID)prgCmds[i].cmdID) {
+                        case VSConstants.VSStd2KCmdID.SETASSTARTPAGE:
+                        case VSConstants.VSStd2KCmdID.CHECK_ACCESSIBILITY:
+                            prgCmds[i].cmdf = (uint)(OLECMDF.OLECMDF_INVISIBLE | OLECMDF.OLECMDF_SUPPORTED | OLECMDF.OLECMDF_DEFHIDEONCTXTMENU | OLECMDF.
```
mousetraps: Additionally you'll won't want to display the "node" and "browser" labels on folders when ES6IntelliSensePreview is enabled since it doesn't actually change anything. Disabling the option is  a good choice as well. 
https://github.com/Microsoft/nodejstools/blob/53c962502871176172ecdd2e489c715893128fa2/Nodejs/Product/Nodejs/Options/NodejsGeneralOptionsPage.cs#L89

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.resx
mousetraps: Check the tab indices - this should be right after the radio button - WinForms can be annoying...

mousetraps: How about "Learn more and send feedback" so it's more of an action?

Also now that this link is there, it's a bit difficult to group the radio buttons and link together, so adding additional margin between the radio buttons and the intellisense log entry limit would be useful.  

@billti thoughts since you commented on this earlier?

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -577,9 +577,9 @@ public NodejsProjectNode(NodejsProjectPackage package)
                 return false;
             }
 
-            var relativeFile = CommonUtils.GetRelativeFilePath(this.FullPathToChildren, fileNode.Url);
+            var relativeFile = CommonUtils.GetRelativeFilePath(this.FullPathToChildren, fileNode.Url);
             if (this._analyzer != null && this._analyzer.Project != null 
-                && this._analyzer.Project.Limits.IsPathExceedNestingLimit(relativeFile)) {
+                && this._analyzer.Project.Limits.IsPathExceedNestingLimit(relativeFile)) {
```
mousetraps: can you verify that the line endings didn't change in this file? (I suspect that's why they're still showing up in the diff)

mousetraps:
Left some minor comments and lgtm after that!

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.resx
jramsay: Works for me. @bill: holler if you prefer the other wording

billti: I like it! Sounds good.

Nodejs/Product/Nodejs/NodejsProject.cs
jramsay: Thanks!  Made this change

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
jramsay: it looks like there is an extra space after the null - let me try to delete it and see if that does the trick.

mousetraps:
:+1: 

Merging this in now. Too bad about the top margin, but that's not a blocking issue

PR #379, opened 2015-08-21
IntelliSense options page save settings fix

jramsay:
Issue: the intellisense options page AnalysisLevel fallback saves the new setting back to storage. There are cases where the call to SaveSettingsStorage() can throw an exception if it is called before some of the other settings have been retrieved. 

Ex: An attempt to SaveSettingsStorage() when the CompletionsComittedBy string is null
SaveString(CompletionCommittedBySetting, CompletionCommittedBy);

Fix:  Move the call to SaveSettingsStorage to the end of LoadSettingsFromStorage() to guarantee that all settings have been retrieved and the final settings with default fallbacks are stored.


mousetraps:
:+1:

PR #381, opened 2015-08-24
Telemetry

nareshjo:
Open piece of NTVS telemetry which consumes private telemetry loggers


Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -613,6 +615,8 @@
     <Compile Include="SmartIndent.cs" />
     <Compile Include="SmartIndentProvider.cs" />
     <Compile Include="SourceMapping\SourceMap.cs" />
+    <Compile Include="Telemetry\TelemetryEvents.cs" />
+    <Compile Include="Telemetry\TelemetryProperties.cs" />
```
mousetraps: If these files are going to be shared across projects, they should be included alongside the Telemetry dlls in Nodejs/Common, and then linked to from the individual projects.

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -809,6 +813,23 @@
     <ZipProject Include="ProjectTemplates\AddWebSiteAzureExpressApp\Preview.png" />
     <ZipProject Include="ProjectTemplates\AzureNodejsApp\Preview.png" />
     <ZipProject Include="ProjectTemplates\NodejsConsoleApp\Preview.png" />
+    <Content Include="..\..\Common\Telemetry\Microsoft.NodejsTools.Telemetry.$(VSTarget).dll">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <IncludeInVSIX>true</IncludeInVSIX>
+	  <VSIXSubPath>.</VSIXSubPath>
```
mousetraps: nit: fix the indentation here and below too (spaces, not tabs)

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -809,6 +813,23 @@
     <ZipProject Include="ProjectTemplates\AddWebSiteAzureExpressApp\Preview.png" />
     <ZipProject Include="ProjectTemplates\AzureNodejsApp\Preview.png" />
     <ZipProject Include="ProjectTemplates\NodejsConsoleApp\Preview.png" />
+    <Content Include="..\..\Common\Telemetry\Microsoft.NodejsTools.Telemetry.$(VSTarget).dll">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <IncludeInVSIX>true</IncludeInVSIX>
+	  <VSIXSubPath>.</VSIXSubPath>
+    </Content>
+    <Content Include="..\..\packages\Microsoft.ApplicationInsights.1.2.0\lib\net45\Microsoft.ApplicationInsights.dll">
```
mousetraps: No better place to put this - why is .nuget/nuget.exe being checked into the project?

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -809,6 +813,23 @@
...
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <IncludeInVSIX>true</IncludeInVSIX>
+	  <VSIXSubPath>.</VSIXSubPath>
+    </Content>
+    <Content Include="..\..\packages\Microsoft.ApplicationInsights.PersistenceChannel.1.2.0\lib\net45\Microsoft.ApplicationInsights.PersistenceChannel.dll">
+      <Link>Microsoft.ApplicationInsights.PersistenceChannel.dll</Link>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <IncludeInVSIX>true</IncludeInVSIX>
+	  <VSIXSubPath>.</VSIXSubPath>
+    </Content>
```
mousetraps: Should this be conditional for VS versions?

Nodejs/Product/Nodejs/NodejsPackage.cs
```diff
@@ -248,6 +252,17 @@ internal sealed partial class NodejsPackage : CommonPackage {
             _logger.LogEvent(NodejsToolsLogEvent.AnalysisLevel, IntellisenseOptionsPage.AnalysisLevel);
         }
 
+        private void InitializeTelemetry() {
+            if (_telemetryLogger == null) {
```
mousetraps: null check is unnecessary because it's always null here.

Nodejs/Setup/BuildRelease.ps1
```diff
@@ -243,7 +239,9 @@ $managed_files = (
     "Microsoft.NodejsTools.Npm.dll",
     "Microsoft.NodejsTools.TestAdapter.dll",
     "Microsoft.NodejsTools.PressAnyKey.exe",
-    "Microsoft.NodejsTools.VsLogger.dll"
+    "Microsoft.NodejsTools.Telemetry.11.0.dll",
+    "Microsoft.NodejsTools.Telemetry.12.0.dll",
+    "Microsoft.NodejsTools.Telemetry.14.0.dll"
```
mousetraps: Be sure to replace these files with the updated versions.

Nodejs/Setup/NodejsTools/NodejsToolsFiles.proj
```diff
@@ -17,10 +17,13 @@
         <File Include="Microsoft.NodejsTools.ProjectWizard.dll" />
         <File Include="Microsoft.NodejsTools.Analysis.dll" />
         <File Include="Microsoft.NodejsTools.TestAdapter.dll" />
+        <File Include="Microsoft.NodejsTools.Telemetry.$(VSTarget).dll" />
 
         <!-- Support Files -->
         <File Include="all.json" />
         <File Include="Newtonsoft.Json.dll"/>
+        <File Include="Microsoft.ApplicationInsights.dll"/>
+        <File Include="Microsoft.ApplicationInsights.PersistenceChannel.dll"/>
```
mousetraps: Are these necessary to package with VS2015+? Can we make it conditional?

Nodejs/Product/Nodejs/Telemetry/TelemetryEvents.cs
```diff
@@ -0,0 +1,24 @@
...
+//    implied. See the License for the specific language governing 
+//    permissions and limitations under the License.
+//
+//*********************************************************//
+
+namespace Microsoft.NodejsTools.Telemetry {
+    /// <summary>
+    /// Telemetry event names
+    /// </summary>
+    public static class TelemetryEvents {
```
mousetraps: should be internal

Nodejs/Product/Nodejs/Telemetry/TelemetryProperties.cs
```diff
@@ -0,0 +1,24 @@
...
+//    implied. See the License for the specific language governing 
+//    permissions and limitations under the License.
+//
+//*********************************************************//
+
+namespace Microsoft.NodejsTools.Telemetry {
+    /// <summary>
+    /// Property names for telemetry events
+    /// </summary>
+    public static class TelemetryProperties {
```
mousetraps: internal

mousetraps:
Left some comments and lgtm after that

Nodejs/Product/Nodejs/Nodejs.csproj
mousetraps: nvm, as we discussed - didn't realize it was replacing with a new version

Common/Product/SharedProject/Telemetry/TelemetryEvents.cs
```diff
@@ -0,0 +1,24 @@
...
+//*********************************************************//
+
+namespace Microsoft.VisualStudioTools.Telemetry {
+    /// <summary>
+    /// Telemetry event names
+    /// </summary>
+    internal static class TelemetryEvents {
+        public const string ProjectImported = "ProjectImported";
+    }
+}
```
mousetraps: I meant to move this to Nodejs/Common, not Common/* alongside the rest of the dlls. 

Again, ```Common/*``` are shared components with Python Tools, and pretty much anyone who's forked us.

Nodejs/Common are shared components within the Nodejs product that we can reference from different NTVS projects.

mousetraps:
lgtm after the above comment about location is addressed.

mousetraps:
:+1:

mousetraps:
Hey, I spent some time debugging the signed binary issue... The problem is that the new telemetry binary were not getting copied to ```$unsigned_msi```s because the ```_CopyOutputsToPath``` target only copies files based on the project being built. In order to fix this, we can add the following copy task to the [_CopyOutputsToPath target in Build/Common.Build.targets](https://github.com/Microsoft/nodejstools/blob/master/Build/Common.Build.targets#L130-L147
)

```xml
<Copy SourceFiles="@(OutputBinaries -> '$(OutputPath)%(filename)%(extension)')" DestinationFolder="$(CopyOutputsToPath)UnsignedBinaries" Condition="'@(OutputBinaries)' != ''"/>
```

Then in the project file, you can define the following item collection to ensure the dll is copied to the right location for signing. 
```xml
<OutputBinaries Include="Microsoft.NodejsTools.Telemetry.$(VSTarget).dll;SomeOtherAssembly.dll" />
```

cc @sanyamc-msft since this'll also apply to your PR.

sanyamc-msft:
Thanks! I added the above copy steps to my PR and pushed the changes for review. Need to test it again on build machine.

mousetraps:
:+1:

Does everything appear to be functional on the build machine now? If so, this is ready to be merged in.

nareshjo:
Yes everything is working as expected in my tests using signed installer. Mergining in

nareshjo:
It seems I don't have permission to merge. Could you please do that.

Issue #308, opened 2015-07-21
VS2015 RTM Exceptions on Project Load


atom0s:
```
System.AggregateException: One or more errors occurred. ---> System.ArgumentNullException: Value cannot be null.
Parameter name: key
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Collections.Generic.Dictionary`2.set_Item(TKey key, TValue value)
   at Microsoft.NodejsTools.Npm.SPI.AbstractNodeModules.AddModule(IPackage package)
   at Microsoft.NodejsTools.Npm.SPI.NodeModules..ctor(IRootPackage parent, Boolean showMissingDevOptionalSubPackages)
   at Microsoft.NodejsTools.Npm.SPI.RootPackage..ctor(String fullPathToRootDirectory, Boolean showMissingDevOptionalSubPackages)
   at Microsoft.NodejsTools.Npm.SPI.NpmController.<RefreshAsync>d__32.MoveNext()
   --- End of inner exception stack trace ---
---> (Inner Exception #0) System.ArgumentNullException: Value cannot be null.
Parameter name: key
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Collections.Generic.Dictionary`2.set_Item(TKey key, TValue value)
   at Microsoft.NodejsTools.Npm.SPI.AbstractNodeModules.AddModule(IPackage package)
   at Microsoft.NodejsTools.Npm.SPI.NodeModules..ctor(IRootPackage parent, Boolean showMissingDevOptionalSubPackages)
   at Microsoft.NodejsTools.Npm.SPI.RootPackage..ctor(String fullPathToRootDirectory, Boolean showMissingDevOptionalSubPackages)
   at Microsoft.NodejsTools.Npm.SPI.NpmController.<RefreshAsync>d__32.MoveNext()<---
```
After loading a nodejs project these present in the output window. These are listed as output from 'Bower/npm'.

VS2015 RTM
NodeJS Tools 1.1 RC

mousetraps:
Thx for reporting this. Would you mind copying the full output from the output window? Does the npm node appear correctly in solution explorer? (for both global and local packages).

atom0s:
The above is the only output at all. Nothing else is in the log but that. 

The npm node is completely empty. It shows the default branches for 'dev', 'global', and 'optional'. But it does not show any of the packages at all.

[...]

Issue #233, opened 2015-06-26
NPM3 is coming


kant2002:
With upcoming release of NPM3 and it's flat dependency tree we have to find another way to fix performance problems with large applications. Nested depth limit will no longer working.

Does you have any plans and ideas how this could be improved further?

Potentially affected areas:
- IntelliSense
  * General analysis logic - this needs the most investigation
  * nested modules limit
- npm integration in solution explorer
- We should verify this doesn't break require intellisense by displaying too many packages to choose from.

Potential actions to take:
- We'll probably end up using `npm ls --json` to get the tree. (This command was not available when NTVS was started)

mousetraps:
Yeah, we've been discussing this lately - npm v3 is one big core issue to handle - more so than the that one feature in analysis.

So the first step is to investigate places were we rely on the nested dependency structure, so that we have a thorough understand all of the affected areas. And then we can discuss the appropriate fix.

The potentially affected feature areas include:
* IntelliSense
  * General analysis logic - this needs the most investigation
  * nested modules limit
* npm integration in solution explorer

And we should verify everything else.

mousetraps:
And we'll probably end up using ```npm ls --json``` to get the tree (I don't believe this command was available when we started on NTVS)

kant2002:
One thing which I think of when try to improve overall speed of analysis on modules with deep structure is that possible we could have global cache of all analysed modules. That way analysis could be done just on user code, since other modules would be pre-cached, We will still have large initial startup time, but data for Intellisense would be already available on second run and for other projects  hopefully for free. 

I will take a look at other bugs which I could fix, to familiarize myself with code base, then could come back to discussion in details. The more I work with Node.JS the more I need VS. Thanks for the product :+1:

mousetraps:
Adding things as they come to me. We should verify this doesn't break require intellisense by displaying too many packages to choose from.

mousetraps:
Okay, I'm starting on this one now. Some working notes so far:
* got the npm node working-ish (need to fix a few quirks, but otherwise all is well and good.)
* discluding all files in the .staging directory that npm v3 creates as it runs
* working on fixing the nested modules limit. We can't quite combine it with the npm node logic because the npm nodes will update slower than files are getting included in the project. So for this I plan to do something similar to ```npm ls --depth=<nestedModulesLimit> | findstr <dirname>```. The cool thing is that npm v3 creates a staging directory when it is working, so npm ls will always reflect the current state.
[...]

PR #386, opened 2015-08-26
#233 npm v3 support, #308 VS2015 RTM exceptions on project load

mousetraps:
#233 npm v3 support
- support solution explorer tree view
- analysis depth optimizations
- add .staging to ignored directories
- ensure global scope visibility is not limited during analysis

#308 VS2015 RTM exceptions on project load
- ensure key is not null before adding package

The bulk of the changes are there, but note that this is not a comprehensive npm v3 changeset - still need to...
- write unit tests to verify npm tree
- fix require intellisense to show fewer completions.

These changes are not as risky, though, so I'll file RTM issues for these items if I can't get to them today.


Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -402,9 +477,7 @@ public NodejsProjectNode(NodejsProjectPackage package)
                 var ignoredPaths = GetProjectProperty(NodejsConstants.AnalysisIgnoredDirectories);
 
                 if (!string.IsNullOrWhiteSpace(ignoredPaths)) {
-                    _analysisIgnoredDirs = ignoredPaths.Split(';').Select(x => '\\' + x + '\\').ToArray();
-                } else {
-                    _analysisIgnoredDirs = new string[0];
+                    _analysisIgnoredDirs.Append(ignoredPaths.Split(';').Select(x => '\\' + x + '\\').ToArray());
```
billti: Is _analysisIgnoredDirs just a list? How often is this function called? Is there a risk here of repeatedly appending to the list over time? (If it's a set, this seems safe).

Nodejs/Product/Npm/SPI/NodeModules.cs
```diff
@@ -14,47 +14,129 @@
...
+
+                    // Go through every directory in node_modules, and see if it's required as a top-level dependency
+                    // We can check this by looking at whether it begins with a hash tag. 
+                    foreach (var moduleDir in directories) {
+                        if (moduleDir.Length < NativeMethods.MAX_FOLDER_PATH && !moduleDir.EndsWith(bin) && !moduleDir.EndsWith(staging)) {
+                            var packageJson = PackageJsonFactory.Create(new DirectoryPackageJsonSource(moduleDir));
+
+                            if (packageJson != null) {
+                                var enumerator = packageJson.RequiredBy.GetEnumerator();
+                                if (packageJson.RequiredBy.Count() > 0) {
```
billti: Could this or the line before throw an exception if the requiredby field is not present (i.e. on NPM < v3)?

Nodejs/Product/Analysis/Analysis/Analyzer/RequireAnalysisUnit.cs
```diff
@@ -0,0 +1,57 @@
...
+            UpdateVisibilities(_tree, _table, _entry, _dependency);
+        }
+
+        internal static void UpdateVisibilities(ModuleTree _tree, ModuleTable _table, ProjectEntry entry, string dependency) {
+            ModuleTree module = _table.RequireModule(new RequireAnalysisUnit(_tree, _table, entry, dependency), dependency, _tree);
+            if (module == null) {
+                return;
+            }
+
+            var enumerator = module.Children.Values.GetEnumerator();
```
DinoV: Is there a reason this isn't just a foreach?

Nodejs/Product/Analysis/Analysis/Analyzer/RequireAnalysisUnit.cs
```diff
@@ -0,0 +1,57 @@
...
+            while (enumerator.MoveNext()) {
+                var childTree = enumerator.Current;
+                if (childTree.ProjectEntry != null) {
+                    _table.AddVisibility(_tree, childTree.ProjectEntry);
+                }
+            }
+        }
+
+        internal override ModuleEnvironmentRecord GetDeclaringModuleEnvironment() {
```
DinoV: This override doesn't look like it's needed

Nodejs/Product/Npm/SPI/PackageJson.cs
```diff
@@ -153,6 +155,17 @@ internal class PackageJson : IPackageJson {
             }
         }
 
+        private void InitRequiredBy() {
+            try {
+                RequiredBy = (_package["_requiredBy"] as IEnumerable<JToken> ?? Enumerable.Empty<JToken>()).Values<string>();
```
billti: OK. Looks like this line addresses the above comment (and the RequiredBy enumerator shouldn't be null).

Nodejs/Product/Analysis/Analysis/Analyzer/RequireAnalysisUnit.cs
DinoV: I think you want to recursively add all children as long as you don't go into a node_modules folder.  Otherwise subdirectories that are required using the path won't be picked up.

Nodejs/Product/Analysis/Analysis/Analyzer/RequireAnalysisUnit.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public RequireAnalysisUnit(ModuleTree tree, ModuleTable table, ProjectEntry entry, string dependency) : base (entry.Tree, entry.EnvironmentRecord) {
+            _tree = tree;
+            _table = table;
+            _entry = entry;
+            _dependency = dependency;
+        }
+
+        internal override void AnalyzeWorker(DDG ddg, CancellationToken cancel) {
+            UpdateVisibilities(_tree, _table, _entry, _dependency);
```
DinoV: This is going to create a new AnalysisUnit each time we re-analyze this AnalysisUnit.  I'm not sure that'll break anything necessarily but it might lead to a memory leak.  I would suggest moving the logic of UpdateVisibilities into here, and make UpdateVisibilitites do new RequireAnalysisUnit(...).AnalyzeWorker() instead or something like that.

Nodejs/Product/Analysis/Analysis/JsAnalyzer.cs
```diff
@@ -140,8 +140,13 @@ internal partial class JsAnalyzer : IGroupableAnalysisProject, IDeserializeIniti
             }
 
             var tree = Modules.GetModuleTree(Path.GetDirectoryName(filePath));
-            
+
             tree.DefaultPackage = entryPoint;
+            if (dependencies != null) {
+                foreach (var dependency in dependencies) {
+                    RequireAnalysisUnit.UpdateVisibilities(tree, Modules, new ProjectEntry(this, filePath, null), dependency);
```
DinoV: the "new ProjectEntry(" should at least be moved out of the loop...  I wonder if you could get rid of it entirely by using a different ProjectEntry but it may not be worth it.

Nodejs/Product/Nodejs/Extensions.cs
```diff
@@ -121,10 +121,12 @@ out project
             }
         }
 
-        internal static T[] Append<T>(this T[] list, T item) {
-            T[] res = new T[list.Length + 1];
+        internal static T[] Append<T>(this T[] list, params T[] items) {
```
DinoV: I'm not quite sure where this is used, but it might be worth adding a new overload rather then just changing this one.  If it's perf critical you're now creating an extra new array whenever you need to append a single item.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
mousetraps: It gets called on project load, and the ProjectNodes don't get reused on Reload, so it's effectively set.

Nodejs/Product/Analysis/Analysis/Analyzer/RequireAnalysisUnit.cs
mousetraps: yep - ended up finding another place in the code we did this, so extracted out to extension method

Nodejs/Product/Analysis/Analysis/JsAnalyzer.cs
mousetraps: Yeah, I a tried to reuse a project entry, but it was way too much hassle. Moving it out of the loop.

Nodejs/Product/Nodejs/Extensions.cs
mousetraps: There's only one other usage of it now, and it's not perf-critical, but it doesn't hurt to make it a new one anyways.

billti:
Code looks good. I'll merge it in and test it. Thanks!

Issue #367, opened 2015-08-14
add privacy checkbox and update text for NTVS installer


mousetraps:
This needs to go in with the new telemetry changes

PR #389, opened 2015-08-27
Adding privacy statement link to installer

munyirik:
- The reason for the separate privacy link control is that I cannot add a hyperlink to the checkbox control text. 
- Assumption made with the new text is that only en-us language is supported.
- Change in License.rtf is to center the apache site text.


billti:
This is for issue https://github.com/Microsoft/nodejstools/issues/367 .  Reviewing now...

PR #390, opened 2015-08-28
Fix build error

zhengbli:


zhengbli:
No longer needed. Closed.

PR #395, opened 2015-08-28
Fix #384 Node.js projects slow to load when npm dependency tree is large

### LINKED ISSUE #384: Node.js projects slow to load when npm dependency tree is large
This is a regression as a result of some project system changes - working on a fix for rc2

mousetraps:
- no need to iterate through file nodes while assigning labels when the
  folder itself is excluded or contains node_modules

To give you an idea for the impact of this change, the project load time goes from ~45s -> ~7s for a project with the following dependencies.
````
├── a@0.5.0 extraneous
├── azure@0.10.6 extraneous
├── azure-storage@0.5.0
├── b@2.0.1 extraneous
├── bindings@1.2.1 extraneous
├── blah@2.1.0 extraneous
├── express@4.13.3
├── g@2.0.0 extraneous
├── ga@0.0.2
├── ha@0.0.17 extraneous
├── jade@1.11.0
├── nan@1.8.4 extraneous
└── socket.io@1.3.6 extraneous
```

We still have general issues in the project system dealing with the node_modules folder and all the resulting Hierarchy nodes, which we should investigate (hence why we include the "don't include node_modules folder" warning when users attempt to include it in the project.)

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsFolderNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsFolderNode.cs
@@ -44,50 +44,57 @@ namespace Microsoft.NodejsTools.Project {
         }
 
         public void UpdateContentType() {
+            var oldContentType = _contentType;
+            _contentType = FolderContentType.None;
+            var parent = Parent as NodejsFolderNode;
             _containsNodeOrBrowserFiles = false;
-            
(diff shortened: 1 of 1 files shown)


billti:
Adding reference to #384

billti:
:+1:

I assume this doesn't affect files included by analysis in any way? I'll merge it in, build, and give it a spin. Thanks!

PR #396, opened 2015-08-28
Change the command to raise when adding new project from existing code

### LINKED ISSUE #133: Adding new Node.js project from existing code removes all other projects from solution
Existing solution with projects, add a new project from existing code, all other projects in the solution are removed.

zhengbli:
This PR fixes #133.

Previously the raised command when adding new project from existing code is ``VSConstants.VSStd97CmdID.OpenProject``, which actually corresponds to the ``Open Project`` menu item, that will always create a new solution. Changing it to ``VSConstants.VSStd97CmdID.AddExistingProject`` fixed the problem.

### DIFF
--- a/Nodejs/Product/Nodejs/Commands/ImportWizardCommand.cs
+++ b/Nodejs/Product/Nodejs/Commands/ImportWizardCommand.cs
@@ -33,7 +33,7 @@ namespace Microsoft.NodejsTools.Commands {
             statusBar.SetText("Importing project...");
 
             var dlg = new Microsoft.NodejsTools.Project.ImportWizard.ImportWizard();
-            String context = "";
+            int commandIdToRaise = (int)VSConstants.VSStd97CmdID.OpenProject;
             
             Microsoft.VisualStudio.Shell.OleMenuCmdEventArgs oleArgs = args as Microsoft.VisualStudio.Shell.OleMenuCmdEventArgs;
             if (oleArgs != null) {
@@ -47,7 +47,7 @@ namespace Microsoft.NodejsTools.Commands {
                             ".njsproj"
                         );
                         dlg.ImportSettings.SourcePath = argItems[1];
-                        context = argItems[2];
+                        commandIdToRaise = int.Parse(argItems[2]);
                     }
                 }
             }
@@ -84,12 +84,7 @@ namespace Microsoft.NodejsTools.Commands {
                         }
                         if (File.Exists(path)) {
                             object outRef = null, pathRef = "\"" + path + "\"";
-                            if (context == "AddNewProject") {
-                                NodejsPackage.Instance.DTE.Commands.Raise(VSConstants.GUID_VSStandardCommandSet97.ToString("B"), (int)VSConstants.VSStd97CmdID.AddExistingProject, ref pathRef, ref outRef);
-                            }
-                            else {
-                                NodejsPackage.Instance.DTE.Commands.Raise(VSConstants.GUID_VSStandardCommandSet97.ToString("B"), (int)VSConstants.VSStd97CmdID.OpenProject, ref pathRef, ref outRef);
-                            }
+                            NodejsPackage.Instance.DTE.Commands.Raise(VSConstants.GUID_VSStandardCommandSet97.ToString("B"), commandIdToRaise, ref pathRef, ref outRef);
                             statusBar.SetText(String.Empty);
                         } else {
                             statusBar.SetText("An error occurred and your project was not created.");
--- a/Nodejs/Product/Nodejs/Commands/ImportWizardCommand.cs
+++ b/Nodejs/Product/Nodejs/Commands/ImportWizardCommand.cs
@@ -81,7 +81,15 @@ namespace Microsoft.NodejsTools.Commands {
                         }
                         if (File.Exists(path)) {
                             object outRef = null, pathRef = "\"" + path + "\"";
-                            NodejsPackage.Instance.DTE.Commands.Raise(VSConstants.GUID_VSStandardCommandSet97.ToString("B"), (int)VSConstants.VSStd97CmdID.AddExistingProject, ref pathRef, ref outRef);
+                            var addExistingProjectCmd = NodejsPackage.Instance.DTE.Commands.Item("File.AddExistingProject");
+                            if (addExistingProjectCmd.IsAvailable)
+                            {
+                                NodejsPackage.Instance.DTE.Commands.Raise(VSConstants.GUID_VSStandardCommandSet97.ToString("B"), (int)VSConstants.VSStd97CmdID.AddExistingProject, ref pathRef, ref outRef);
+                            }
+                            else
+                            {
+                                NodejsPackage.Instance.DTE.Commands.Raise(VSConstants.GUID_VSStandardCommandSet97.ToString("B"), (int)VSConstants.VSStd97CmdID.OpenProject, ref pathRef, ref outRef);
+                            }
                             statusBar.SetText(String.Empty);
                         } else {
                             statusBar.SetText("An error occurred and your project was not created.");
--- a/Nodejs/Product/Nodejs/Commands/ImportWizardCommand.cs
+++ b/Nodejs/Product/Nodejs/Commands/ImportWizardCommand.cs
@@ -81,7 +81,7 @@ namespace Microsoft.NodejsTools.Commands {
                         }
                         if (File.Exists(path)) {
                             object outRef = null, pathRef = "\"" + path + "\"";
-                            NodejsPackage.Instance.DTE.Commands.Raise(VSConstants.GUID_VSStandardCommandSet97.ToString("B"), (int)VSConstants.VSStd97CmdID.OpenProject, ref pathRef, ref outRef);
+                            NodejsPackage.Instance.DTE.Commands.Raise(VSConstants.GUID_VSStandardCommandSet97.ToString("B"), (int)VSConstants.VSStd97CmdID.AddExistingProject, ref pathRef, ref outRef);
                             statusBar.SetText(String.Empty);
                         } else {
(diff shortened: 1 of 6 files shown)


paulvanbrenk:
:+1:

billti:
Thanks for figuring this out!

Now you are calling AddProject instead of OpenProject, I can see how this fixes the issue referenced. Did you also check that running the "Import from existing code" scenario with no existing project/sln open still continues to work also?

zhengbli:
Right, after some tests it didn't work for the no-existing project/sln scenario. I need to fix that.

zhengbli:
@billti Added code to detect the availability of the "Add Existing Project", now it works as expected in both cases.

Nodejs/Product/Nodejs/Commands/ImportWizardCommand.cs
```diff
@@ -81,7 +81,15 @@ class ImportWizardCommand : Command {
                         }
                         if (File.Exists(path)) {
                             object outRef = null, pathRef = "\"" + path + "\"";
-                            NodejsPackage.Instance.DTE.Commands.Raise(VSConstants.GUID_VSStandardCommandSet97.ToString("B"), (int)VSConstants.VSStd97CmdID.OpenProject, ref pathRef, ref outRef);
+                            var addExistingProjectCmd = NodejsPackage.Instance.DTE.Commands.Item("File.AddExistingProject");
```
mousetraps: Does this work in the instance where a project is open (so the AddExistingProject command is available), but someone is nonetheless creating a new project?

Perhaps a better solution is to intercept the command in the NodejsPackage.Initialize() step by handling BeforeExecute and AfterExecute events (set a boolean flag that we check in ImportWizardCommand). 
E.g. http://stackoverflow.com/questions/15406523/visual-studio-extensibility-intercepting-file-paste-event

Nodejs/Product/Nodejs/Commands/ImportWizardCommand.cs
```diff
@@ -81,7 +81,15 @@ class ImportWizardCommand : Command {
...
+                            if (addExistingProjectCmd.IsAvailable)
```
mousetraps: might need a null check here to handle cases that the command doesn't exist altogether.

Nodejs/Product/Nodejs/Commands/ImportWizardCommand.cs
```diff
@@ -81,7 +81,15 @@ class ImportWizardCommand : Command {
                         }
                         if (File.Exists(path)) {
                             object outRef = null, pathRef = "\"" + path + "\"";
-                            NodejsPackage.Instance.DTE.Commands.Raise(VSConstants.GUID_VSStandardCommandSet97.ToString("B"), (int)VSConstants.VSStd97CmdID.OpenProject, ref pathRef, ref outRef);
+                            var addExistingProjectCmd = NodejsPackage.Instance.DTE.Commands.Item("File.AddExistingProject");
+                            if (addExistingProjectCmd.IsAvailable)
+                            {
```
mousetraps: braces on same line here and everywhere else

Nodejs/Product/Nodejs/Commands/ImportWizardCommand.cs
```diff
@@ -81,7 +84,12 @@ class ImportWizardCommand : Command {
                         }
                         if (File.Exists(path)) {
                             object outRef = null, pathRef = "\"" + path + "\"";
-                            NodejsPackage.Instance.DTE.Commands.Raise(VSConstants.GUID_VSStandardCommandSet97.ToString("B"), (int)VSConstants.VSStd97CmdID.OpenProject, ref pathRef, ref outRef);
+                            if (context == "AddNewProject") {
+                                NodejsPackage.Instance.DTE.Commands.Raise(VSConstants.GUID_VSStandardCommandSet97.ToString("B"), (int)VSConstants.VSStd97CmdID.AddExistingProject, ref pathRef, ref outRef);
+                            }
+                            else {
+                                NodejsPackage.Instance.DTE.Commands.Raise(VSConstants.GUID_VSStandardCommandSet97.ToString("B"), (int)VSConstants.VSStd97CmdID.OpenProject, ref pathRef, ref outRef);
+                            }
```
mousetraps: Alternatively, just pass in the cmdIdToRaise rather than effectively checking it twice.

Then, you can can simplify all this to:
```csharp
NodejsPackage.Instance.DTE.Commands.Raise(
VSConstants.GUID_VSStandardCommandSet97.ToString("B"), cmdIdToRaise, ref pathRef, ref outRef);
```

Nodejs/Product/Nodejs/NodejsPackage.cs
```diff
@@ -38,6 +38,7 @@
 using Microsoft.NodejsTools.Project;
 using Microsoft.NodejsTools.Repl;
 using Microsoft.NodejsTools.Telemetry;
+using Microsoft.NodejsTools.ProjectWizard;
```
mousetraps: sort usings - alphabetical order (except System.*, which belongs at the top)

Nodejs/Product/Nodejs/NodejsPackage.cs
```diff
@@ -182,6 +183,10 @@ internal sealed partial class NodejsPackage : CommonPackage {
         // Overridden Package Implementation
         #region Package Members
 
+        // Hold references for the subscribed events. Otherwise the callbacks will be garbage collected
+        // after the initialization
+        private List<EnvDTE.CommandEvents> subscribedCommandEvents = new List<EnvDTE.CommandEvents>();
```
mousetraps: private members should begin with underscores. Also move this to the top with the rest.

Nodejs/Product/Nodejs/NodejsPackage.cs
```diff
@@ -190,6 +195,17 @@ internal sealed partial class NodejsPackage : CommonPackage {
...
+            var addNewProjectId = (int)VSConstants.VSStd97CmdID.AddNewProject;
+            var addNewProjectEvent = DTE.Events.CommandEvents[addNewProjectGuid, addNewProjectId];
+            addNewProjectEvent.BeforeExecute += delegate {
+                NewProjectFromExistingWizard.AddingNewProject = true;
+            };
+            addNewProjectEvent.AfterExecute += delegate {
+                NewProjectFromExistingWizard.AddingNewProject = false;
+            };
+            subscribedCommandEvents.Add(addNewProjectEvent);
+
```
mousetraps: worth splitting this logic out into a private method, ```SubscribeToVsCommandEvents```

Nodejs/Product/ProjectWizard/ImportWizard.cs
```diff
@@ -29,6 +29,7 @@ public sealed class NewProjectFromExistingWizard : IWizard {
         public void ProjectFinishedGenerating(EnvDTE.Project project) { }
         public void ProjectItemFinishedGenerating(EnvDTE.ProjectItem projectItem) { }
         public void RunFinished() { }
+        public static Boolean AddingNewProject { get; set; } = false;
```
mousetraps: * Make this internal
* include this as a property at the top of the file
* no need to set the default value to false because that's the assumed default.

Nodejs/Product/ProjectWizard/ImportWizard.cs
```diff
@@ -45,6 +46,8 @@ public sealed class NewProjectFromExistingWizard : IWizard {
                     dte = new ServiceProvider(provider).GetService(typeof(DTE)) as DTE;
                 }
             }
+
+            bool addingNewProject = false;
```
mousetraps: What's this for? Shouldn't we be using AddingNewProject here?

Nodejs/Product/ProjectWizard/ImportWizard.cs
```diff
@@ -73,10 +76,12 @@ out package
                         directory = Path.GetDirectoryName(Path.GetDirectoryName(replacementsDictionary["$destinationdirectory$"]));
                     }
 
-                    object inObj = projName + "|" + directory, outObj = null;
+                    var context = addingNewProject ? "AddNewProject" : "OpenProject";
```
mousetraps: pass in cmdId here rather than a string

Nodejs/Product/ProjectWizard/ImportWizard.cs
zhengbli: This is to record the ``AddingNewProject`` before it gets reset. The ``AddingNewProject`` will be set in the ``AfterExecute`` before the ``ImportWizardCommand`` gets raised, so I need to record the value.

mousetraps:
Overall implementation looks good now - left several minor comments and lgtm after that.

Nodejs/Product/ProjectWizard/ImportWizard.cs
mousetraps: It only gets reset after WizardCancelledException.

mousetraps: You're also not actually recording or using the value of AddingNew project anywhere.

zhengbli: No I used it at line 79. The one gets reset is the property, not the local variable.

mousetraps: ```dte.Commands.Raise(...)``` gets called before line 84 and ```throw NewWizardCancelledException```, correct?

Nodejs/Product/ProjectWizard/ImportWizard.cs
zhengbli: > Make this internal

This cannot be internal, as it is referenced in ``NodejsPackage.Initialize()`` which is not in the same  namespace.

zhengbli:
@mousetraps can you take a look at my updates? Thanks!

mousetraps:
:+1:

PR #397, opened 2015-08-28
Fix the message string in file override window

### LINKED ISSUE #184: VS2015 add new item duplicate-item naming fail
When attempting to add a package.json file to a folder that already contains a package.json file, through the Add New Item dialog, the following prompt appears, referencing EmptyJson.json, but that was not the name of the file we were trying to add.
image

This may be related to #183

zhengbli:
Fixed #184


paulvanbrenk:
:+1:

PR #373, opened 2015-08-19
Adding Installer Telemetry

sanyamc-msft:
Adding a Custom Action to send back installer telemetry using http POST.

PR #399, opened 2015-08-28
Revert "Adding Installer Telemetry"

billti:
Reverts Microsoft/nodejstools#373


billti:
Builds/installs/runs fine on my machine, and the installer telemetry custom actions are gone from the MSI. Merging...

Issue #246, opened 2015-07-03
Debugger Doesn't Work 75% Of The Time


atom0s:
- Visual Studio 2013 (Update 4)
 - NTVS.Dev.2015-06-14.VS.2013.msi
 - Node v0.12.5

Most of the time, it seems that the debugger refuses to work properly. If I launch my application through Visual Studio, I would say about 75% of the time, it will show the debugger port in the console being 5858 and then nothing else happens. Typically if this occurs it requires me to restart Visual Studio entirely to get it to work again. And that point, it usually only works once, then dies again from there on til the next restart.

This issue happens with both the default startup path being set to the default:
C:\Program Files\nodejs\node.exe

As well as being set to something custom, such as nodemon:
C:\Users\atom0s\AppData\Roaming\npm\nodemon.cmd

Either or has the same issue.

alejandro5042:
This may be related to this issue I reported yesterday #245. Can you confirm?

atom0s:
Hello @alejandro5042 currently I am not using io.js, just using normal node with some packages such as:
 - express
 - passport
 - sequelize
 - vash

All of which I am pretty sure are the 32bit versions. Given that the debugging works sometimes but not always, I don't think that it is an issue with the version types though.

mousetraps:
Haven't gotten the chance to look into it, but this sounds like a race condition, and we can't rule out the possibility it's related to #245 just yet... It could be that the recent x64 versions of iojs happen to exacerbate it.

atom0s:
Thanks for the response @mousetraps - just to mention it again, I do not use io.js in my scenario. But I do seem to have a similar issue to what alejandro5042 is mentioning in his ticket.

alejandro5042:
Actually, you bring up a good point with the race condition. Even with the io.js x86 which actually works vs. x64, I still have an issue with setting breakpoints. I cannot seem to hit a breakpoint in the first couple of lines of code, yet I can set and hit breakpoints in callbacks that occur after. I just did a little testing and observed that any breakpoint in the first 2 seconds of execution are not hit... anything after works. Thanks for attempting to investigate this @mousetraps!

ant-fx:
I have a similar issue but not so frequent, maybe 50% of the time. I get a message saying debugger listening on port 5858 and then nothing.

I've tried uninstalling 64 bit and installing 32 bit but the problem is still there.

[...]

PR #406, opened 2015-09-02
Debugger connection retry logic

### LINKED ISSUE #245: Running/debugging does not work with x64 version of io.js (but does with x86)
I spent a few hours trying to figure out why debugging doesn't work with io.js despite the [wiki documentation](https://github.com/Microsoft/nodejstools/wiki/io.js-and-ES6-support) saying that it should. Turns out you can't use the x64 version. It *might* launch the first time and run. But you'll never hit a breakpoint or anything. The x86 version works just as you would imagine.

Is this a bug? Expected behavior? Should we update the documentation?

Using:
* io.js 2.3.2 (x86 and x64): https://iojs.org/dist/v2.3.2/
* NTVS 1.1 Beta: https://nodejstools.codeplex.com/releases/view/614706
* Visual Studio 2013 Ultimate - Update 4 (non-elevated)

Thank you for this awesome tool!

billti:
This adds retry logic when attempting to connect to the debug port on Node, in an attempt to resolve #245 and #246.

In some scenarios, especially when using Debug builds of Node/io.js, the Node process doesn't open the debug port (5858 by default) in time. At the network level, this results in the connection being refused. This change adds retry logic if the connection fails (and additional logging).

On my machine, using the 'Debug' build of the latest io.js branch (as using 'Release' shows no issue at all), I can now see the connection retry and succeed, e.g.

```
[17:20:09.0841895] AD7Engine Created (63760528)
[17:20:09.0907085] AD7Engine Called SetRegistryRoot
[17:20:09.0912082] AD7Engine Called SetLocale
[17:20:09.0947085] AD7Engine LaunchSuspended Called with flags 'LAUNCH_ENABLE_ENC' (63760528)
[17:20:09.1791871] Debugger connecting to URI: tcp://localhost:5858/
[17:20:11.2412522] Connection attempt 1 failed with: System.Net.Sockets.SocketException (0x80004005): No connection could be made because the target machine actively refused it 127.0.0.1:5858
   at System.Net.Sockets.TcpClient..ctor(String hostname, Int32 port)
   at Microsoft.NodejsTools.Debugger.Communication.NetworkClientFactory.CreateNetworkClient(Uri uri) in C:\src\nodejstools\Nodejs\Product\Nodejs\Debugger\Communication\NetworkClientFactory.cs:line 32
   at Microsoft.NodejsTools.Debugger.Communication.DebuggerConnection.Connect(Uri uri) in C:\src\nodejstools\Nodejs\Product\Nodejs\Debugger\Communication\DebuggerConnection.cs:line 130
[17:20:12.9912323] Debugger connected successfully
[17:20:12.9912323] DebuggerConnection: Request: {"command":"scripts","seq":1,"type":"request","arguments":{"includeSource":false}}
```

Note that on a retry connect, sometime the debugger enters a state where it doesn't appear to be connected, but using "Debug / Break all" will then break on the first line and debugging can then continue. This needs to be looked at separately.

### DIFF
--- a/Nodejs/Product/Nodejs/Debugger/Communication/DebuggerConnection.cs
+++ b/Nodejs/Product/Nodejs/Debugger/Communication/DebuggerConnection.cs
@@ -22,6 +22,7 @@ using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.NodejsTools.Logging;
+using Microsoft.VisualStudioTools;
 using Microsoft.VisualStudioTools.Project;
 using Newtonsoft.Json;
 
@@ -118,11 +119,9 @@ namespace Microsoft.NodejsTools.Debugger.Communication {
             lock (_networkClientLock) {
                 int connection_attempts = 0;
                 const int MAX_ATTEMPTS = 5;
-                while (true)
-                {
+                while (true) {
                     connection_attempts++;
-                    try
-                    {
+                    try {
                         // TODO: This currently results in a call to the synchronous TcpClient
                         // constructor, which is a blocking call, and can take a couple of seconds
                         // to connect (with timeouts and retries). This code is running on the UI
@@ -132,25 +131,23 @@ namespace Microsoft.NodejsTools.Debugger.Communication {
                         // Unclear if the above can succeed and not be connected, but check for safety.
                         // The code needs to either break out the while loop, or hit the retry logic
                         // in the exception handler.
-                        if (_networkClient.Connected)
-                        {
+                        if (_networkClient.Connected) {
                             LiveLogger.WriteLine("Debugger connected successfully");
                             break;
                         }
-                        else
-                        {
+                        else {
                             throw new SocketException();
                         }
                     }
-                    catch (Exception ex)
-                    {
-                        LiveLogger.WriteLine("Connection attempt {0} failed with: {1}", connection_attempts, ex);
-                        if (connection_attempts >= MAX_ATTEMPTS)
-                        {
+                    catch (Exception ex) {
+                        if (ex.IsCriticalException()) {
                             throw;
                         }
-                        else
-                        {
+                        LiveLogger.WriteLine("Connection attempt {0} failed with: {1}", connection_attempts, ex);
+                        if (connection_attempts >= MAX_ATTEMPTS) {
+                            throw;
+                        }
+                        else {
                             // See above TODO. This should be moved off the UI thread or posted to retry
                             // without blocking in the meantime. For now, this seems the lesser of two
                             // evils. (The other being the debugger failing to attach on launch if the
--- a/Nodejs/Product/Nodejs/Debugger/Communication/DebuggerConnection.cs
+++ b/Nodejs/Product/Nodejs/Debugger/Communication/DebuggerConnection.cs
@@ -112,10 +112,53 @@ namespace Microsoft.NodejsTools.Debugger.Communication {
         /// <param name="uri">URI identifying the endpoint to connect to.</param>
         public void Connect(Uri uri) {
             Utilities.ArgumentNotNull("uri", uri);
+            LiveLogger.WriteLine("Debugger connecting to URI: {0}", uri);
 
             Close();
             lock (_networkClientLock) {
-                _networkClient = _networkClientFactory.CreateNetworkClient(uri);
+                int connection_attempts = 0;
+                const int MAX_ATTEMPTS = 5;
+                while (true)
+                {
+                    connection_attempts++;
+                    try
+                    {
+                        // TODO: This currently results in a call to the synchronous TcpClient
+                        // constructor, which is a blocking call, and can take a couple of seconds
+                        // to connect (with timeouts and retries). This code is running on the UI
+                        // thread. Ideally this should be connecting async, or moved off the UI thread.
+                        _networkClient = _networkClientFactory.CreateNetworkClient(uri);
+
+                        // Unclear if the above can succeed and not be connected, but check for safety.
+                        // The code needs to either break out the while loop, or hit the retry logic
+                        // in the exception handler.
+                        if (_networkClient.Connected)
+                        {
+                            LiveLogger.WriteLine("Debugger connected successfully");
+                            break;
+                        }
+                        else
+                        {
+                            throw new SocketException();
+                        }
+                    }
+                    catch (Exception ex)
+                    {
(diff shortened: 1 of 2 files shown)


Nodejs/Product/Nodejs/Debugger/Communication/DebuggerConnection.cs
```diff
@@ -112,10 +112,53 @@ sealed class DebuggerConnection : IDebuggerConnection {
...
+                        {
+                            LiveLogger.WriteLine("Debugger connected successfully");
+                            break;
+                        }
+                        else
+                        {
+                            throw new SocketException();
+                        }
+                    }
+                    catch (Exception ex)
```
paulvanbrenk: Do you expect any other exceptions than the SocketException you throw? If not, just filter this on that type and let the other exceptions naturally bubble up, this will preserve the stack better.

paulvanbrenk: Since you're not using the exception you should remove the 'ex' variable.

billti: I am using it in the WriteLine call immediately below.

billti: And yes, I do expect other exceptions. In fact, per the comment above the condition, I don't expect to have to throw the exception myself, that's just for safety. In general any exception trying to connect should cause either a retry or a rethrow after the max retries.

paulvanbrenk: In that case, I would add a message to the SocketException you throw to indicate that a networkclient was successfully created but not connected. That way we can investigate further if that happens.

billti: Yup, was going to do that, but there is no SocketException constructor that takes a string, and the Message property is read-only. ( https://msdn.microsoft.com/en-us/library/system.net.sockets.socketexception.socketexception(v=vs.110).aspx )

Nodejs/Product/Nodejs/Debugger/Communication/DebuggerConnection.cs
```diff
@@ -112,10 +112,53 @@ sealed class DebuggerConnection : IDebuggerConnection {
...
             Utilities.ArgumentNotNull("uri", uri);
+            LiveLogger.WriteLine("Debugger connecting to URI: {0}", uri);
 
             Close();
             lock (_networkClientLock) {
-                _networkClient = _networkClientFactory.CreateNetworkClient(uri);
+                int connection_attempts = 0;
+                const int MAX_ATTEMPTS = 5;
+                while (true)
+                {
```
mousetraps: braces on same line here and everywhere else.

Nodejs/Product/Nodejs/Debugger/Communication/DebuggerConnection.cs
mousetraps: I don't expect it to be an issue in ths case, but in general we don't want to swallow critical exceptions. 
https://github.com/Microsoft/nodejstools/blob/35817ffdaa1797f95628a22f588fc3ed885f884b/Common/Product/SharedProject/ExceptionExtensions.cs#L23-L29

Nodejs/Product/Nodejs/Debugger/Communication/DebuggerConnection.cs
```diff
@@ -112,10 +112,53 @@ sealed class DebuggerConnection : IDebuggerConnection {
...
+                        {
+                            throw;
+                        }
+                        else
+                        {
+                            // See above TODO. This should be moved off the UI thread or posted to retry
+                            // without blocking in the meantime. For now, this seems the lesser of two
+                            // evils. (The other being the debugger failing to attach on launch if the
+                            // debuggee socket wasn't open quickly enough).
+                            System.Threading.Thread.Sleep(200);
```
mousetraps: I'm fine taking this for RC2, but let's file an issue on this since we really shouldn't be blocking.

Nodejs/Product/Nodejs/Debugger/DebugEngine/Remote/NodeRemoteEnumDebugProcesses.cs
```diff
@@ -103,18 +103,24 @@ public NodeRemoteEnumDebugProcesses(NodeRemoteEnumDebugProcesses processes)
...
+                    LiveLogger.WriteLine("IOException connecting to remote debugger");
                     exception = ex;
                 } catch (SocketException ex) {
+                    LiveLogger.WriteLine("SocketException connecting to remote debugger");
                     exception = ex;
                 } catch (WebSocketException ex) {
+                    LiveLogger.WriteLine("WebSocketException connecting to remote debugger");
                     exception = ex;
                 } catch (PlatformNotSupportedException) {
+                    LiveLogger.WriteLine("NodeRemoteEnumDebugProcesses ping timed out.");
```
mousetraps: PlatformNotSupportedException doesn't have anything to do with the ping timing out, does it?

Nodejs/Product/Nodejs/Debugger/DebugEngine/Remote/NodeRemoteEnumDebugProcesses.cs
```diff
@@ -103,18 +103,24 @@ public NodeRemoteEnumDebugProcesses(NodeRemoteEnumDebugProcesses processes)
                         break;
                     }
                 } catch (OperationCanceledException) {
-                    LiveLogger.WriteLine("NodeRemoteEnumDebugProcesses ping timed out.");
+                    LiveLogger.WriteLine("OperationCanceledException connecting to remote debugger");
```
mousetraps: We included the "ping timed out" before to make it clearer what the core issue actually was. Don't mind including the exception as well, but we should also include the ping timeout bit.

mousetraps:
some minor comments and lgtm after that

Nodejs/Product/Nodejs/Debugger/Communication/DebuggerConnection.cs
billti: Yeah, but as noted above, just the existing TCP connection call blocks for 2 to 3 seconds already trying to connect. Getting this whole logic off the UI thread, or moving to async, seems desirable, but that's a much more invasive change at this point.

mousetraps: gotcha, makes sense

mousetraps:
:+1:

Issue #393, opened 2015-08-28
NullReferenceException while unloading project


mousetraps:
Perhaps ProjectMgr doesn't exist at this time?
```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.NullReferenceException
Stack:
   at Microsoft.NodejsTools.Project.NodejsFileNode.Analyze()
   at Microsoft.NodejsTools.Project.NodejsProjectNode.OnIdleNodeModules(System.Object)
   at System.Threading.TimerQueueTimer.CallCallbackInContext(System.Object)
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.TimerQueueTimer.CallCallback()
   at System.Threading.TimerQueueTimer.Fire()
   at System.Threading.TimerQueue.FireNextTimers()
   at System.Threading.TimerQueue.AppDomainTimerCallback()
```

Unfortunately I don't have a repro on this. Ran into it while unloading and loading the project several times over.

PR #407, opened 2015-09-02
Add null check for NodejsFileNode.Analyze

zhengbli:
Fix for #393

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
@@ -41,7 +41,7 @@ namespace Microsoft.NodejsTools.Project {
         }
 
         internal void Analyze() {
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
```diff
@@ -41,7 +41,7 @@ public NodejsFileNode(NodejsProjectNode root, ProjectElement e)
         }
 
         internal void Analyze() {
-            if (ProjectMgr.Analyzer != null && ShouldAnalyze) {
+            if (ProjectMgr?.Analyzer != null && ShouldAnalyze) {
```
paulvanbrenk: This won't compile on Dev 12

mousetraps:
lgtm, though did you actually manage to reproduce this issue? Or is this just based on the comment in #393?

zhengbli:
I don't have a repro. It is mainly based on how null reference exception could possibly happen within that function.

Issue #409, opened 2015-09-02
Failed test case: CannonicalHelloWorldTest


hoanhtien:
Many tests in the test suites are failing due to different reasons. I am examining them one by one starting from this test.

![image](https://cloud.githubusercontent.com/assets/1707813/9646869/78c16f00-518c-11e5-978d-e8ed8edd6b24.png)

There is a NullReferenceException with the following call stack: 
```
   at Microsoft.NodejsTools.Logging.LiveLogger..ctor() in D:\mygit\nodejstools\Nodejs\Product\Nodejs\Logging\LiveLogger.cs:line 38
   at Microsoft.NodejsTools.Logging.LiveLogger.get_Instance() in D:\mygit\nodejstools\Nodejs\Product\Nodejs\Logging\LiveLogger.cs:line 46
   at Microsoft.NodejsTools.Logging.LiveLogger.WriteLine(String message) in D:\mygit\nodejstools\Nodejs\Product\Nodejs\Logging\LiveLogger.cs:line 60
   at Microsoft.NodejsTools.Logging.LiveLogger.WriteLine(String format, Object[] args) in D:\mygit\nodejstools\Nodejs\Product\Nodejs\Logging\LiveLogger.cs:line 65
   at Microsoft.NodejsTools.Logging.LiveLogger.WriteLine(String message, Type category) in D:\mygit\nodejstools\Nodejs\Product\Nodejs\Logging\LiveLogger.cs:line 55
   at Microsoft.NodejsTools.Debugger.Communication.DebuggerConnection.<ReceiveAndDispatchMessagesWorker>d__24.MoveNext() in D:\mygit\nodejstools\Nodejs\Product\Nodejs\Debugger\Communication\DebuggerConnection.cs:line 157
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.<>c__DisplayClass2.<ThrowAsync>b__5(Object state)
   at System.Threading.QueueUserWorkItemCallback.WaitCallback_Context(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[...]

PR #410, opened 2015-09-03
Add null check for NodejsPackage.Instance.

hoanhtien:
NodejsPackage.Instance is not available in unit testing. This fixes https://github.com/Microsoft/nodejstools/issues/409.


mousetraps:
:+1:

Issue #411, opened 2015-09-03
*MultilineComment* tests are failing


hoanhtien:
![image](https://cloud.githubusercontent.com/assets/1707813/9667913/7f71ff20-5233-11e5-9207-702813938f0c.png)

Many tests involving multiline comments are failling because `buffer.GetNodejsClassifier()` has the value `null` in https://github.com/Microsoft/nodejstools/blob/master/Nodejs/Product/Nodejs/Editor/EditorExtensions.cs#L175.

![image](https://cloud.githubusercontent.com/assets/1707813/9667937/9d950ea2-5233-11e5-814d-64e08b903afb.png)

PR #412, opened 2015-09-03
Fix multiline comment tests

hoanhtien:
` new MockTextBuffer(startingText, "C:\\app.js", NodejsConstants.Nodejs));` 
should be 
`new MockTextBuffer(startingText, NodejsConstants.Nodejs, "C:\\app.js"));` 
as the constructor of `MockTestBuffer` is:
`public MockTextBuffer(string content, string contentType, string filename = null)`

This fixes https://github.com/Microsoft/nodejstools/issues/411.

### DIFF
--- a/Nodejs/Tests/Core/MultiLineCommentTests.cs
+++ b/Nodejs/Tests/Core/MultiLineCommentTests.cs
@@ -142,7 +142,7 @@ namespace NodejsTests {
 
             // create the view and request a multiline comment format
             var view = new MockTextView(
-                new MockTextBuffer(startingText, "C:\\app.js", NodejsConstants.Nodejs));
+                new MockTextBuffer(startingText, NodejsConstants.Nodejs, "C:\\app.js"));
             var insertionPoint = new SnapshotPoint(view.TextSnapshot, insertionPosition);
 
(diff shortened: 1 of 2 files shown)


Nodejs/Tests/Core/MultiLineCommentTests.cs
```diff
@@ -142,7 +142,7 @@ public class MultilineCommentTests {
 
             // create the view and request a multiline comment format
             var view = new MockTextView(
-                new MockTextBuffer(startingText, "C:\\app.js", NodejsConstants.Nodejs));
+                new MockTextBuffer(startingText, NodejsConstants.Nodejs, "C:\\app.js"));
```
paulvanbrenk: Can you make this a call with named arguments

hoanhtien: Only for `filename: "C:\\app.js"` or for all arguments?

paulvanbrenk: all 3 would be good in this case.

hoanhtien: Changed.

paulvanbrenk:
:+1:

Issue #413, opened 2015-09-03
TestComment* are failling


hoanhtien:
All test cases TestComment* are failing due to the same reason:

![image](https://cloud.githubusercontent.com/assets/1707813/9672111/933433ca-524c-11e5-9418-fa25deec2ee2.png)

`SnapshotSpan(view.TextBuffer.CurrentSnapshot, span)` throws exception because `view.TextBuffer.CurrentSnapshot` is `null`. 

![image](https://cloud.githubusercontent.com/assets/1707813/9672120/a8470684-524c-11e5-9a49-cd4f3a913629.png)

PR #414, opened 2015-09-03
Implement constructor MockTextBuffer(string)

hoanhtien:
This fixes https://github.com/Microsoft/nodejstools/issues/413.

Most TestComment* test cases use the following constructor of MockTextBuffer which by some reason is not implemented :neutral_face:
```C#
MockTextBuffer.cs
31  public MockTextBuffer(string content) {
32  }
```

All TestComment* pass after this change.

### DIFF
--- a/Nodejs/Tests/Core/CommentBlockTests.cs
+++ b/Nodejs/Tests/Core/CommentBlockTests.cs
@@ -17,7 +17,8 @@
 using Microsoft.NodejsTools.Editor.Core;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Microsoft.VisualStudio.Text;
-using TestUtilities.Mocks;
+using MockTextView = TestUtilities.Mocks.MockTextView;
+using MockTextBuffer = NodejsTests.Mocks.MockTextBuffer;
 
 namespace NodejsTests {
     [TestClass]
--- a/Nodejs/Tests/Core/Mocks/MockTextBuffer.cs
+++ b/Nodejs/Tests/Core/Mocks/MockTextBuffer.cs
@@ -1,8 +1,20 @@
-﻿using Microsoft.NodejsTools;
+﻿using System.IO;
+using Microsoft.NodejsTools;
 
 namespace NodejsTests.Mocks {
     class MockTextBuffer : TestUtilities.Mocks.MockTextBuffer {
-		public MockTextBuffer(string content) : base(content: content, contentType: NodejsConstants.Nodejs, filename: "file.js") {
+		public MockTextBuffer(string content) : base(content: content, contentType: NodejsConstants.Nodejs) {
+        }
+
+        private static string GetRandomFileNameIfNull(string filename) {
+            if (filename == null) {
+                filename = Path.Combine(TestUtilities.TestData.GetTempPath(), Path.GetRandomFileName(), "file.js");
+            }
+            return filename;
+        }
+
+        public MockTextBuffer(string content, string contentType, string filename = null) : base(content: content,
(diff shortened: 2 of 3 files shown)


Common/Tests/Utilities/Mocks/MockTextBuffer.cs
```diff
@@ -28,14 +28,14 @@ public class MockTextBuffer : ITextBuffer {
         /// </summary>
         private PropertyCollection _properties;
 
-        public MockTextBuffer(string content) {
+        public MockTextBuffer(string content) : this(content, "Node.js") {
```
paulvanbrenk: Can you use the contentType constant for this?

mousetraps:
This change is in the common codebase that we share with Python Tools, among others. We'll need to find a way to make it work without adding Node.js-specific code to Common/*

hoanhtien:
Derived TestUtilities.Mocks.MockTextBuffer to avoid modifying it directly. Thank Paul for this solution.

Nodejs/Tests/Core/CommentBlockTests.cs
```diff
@@ -17,7 +17,8 @@
 using Microsoft.NodejsTools.Editor.Core;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Microsoft.VisualStudio.Text;
-using TestUtilities.Mocks;
+using MockTextView = TestUtilities.Mocks.MockTextView;
```
paulvanbrenk: Make sure this works in Dev12

Nodejs/Tests/Core/Mocks/MockTextBuffer.cs
```diff
@@ -0,0 +1,8 @@
+﻿using Microsoft.NodejsTools;
+
+namespace NodejsTests.Mocks {
+    class MockTextBuffer : TestUtilities.Mocks.MockTextBuffer {
```
paulvanbrenk: Add a constructor which takes 2 arguments so people can specify the filename.

In the implementation generate the filename the same way the base class does when it's not specified, except with a js file.

Nodejs/Tests/Core/CommentBlockTests.cs
hoanhtien: Confirmed this works in Dev12.

Nodejs/Tests/Core/Mocks/MockTextBuffer.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.IO;
```
mousetraps: add copyright header

Nodejs/Tests/Core/Mocks/MockTextBuffer.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace NodejsTests.Mocks {
+    class MockTextBuffer : TestUtilities.Mocks.MockTextBuffer {
+		public MockTextBuffer(string content) : base(content: content, contentType: NodejsConstants.Nodejs) {
+        }
+
+        private static string GetRandomFileNameIfNull(string filename) {
+            if (filename == null) {
+                filename = Path.Combine(TestUtilities.TestData.GetTempPath(), Path.GetRandomFileName(), "file.js");
+            }
+            return filename;
```
mousetraps: you can simplify this to `return filename ?? Path.Combine(...);`

mousetraps:
minor comments, otherwise lgtm

mousetraps:
did you get a chance to take a look at these comments yet? should be quick enough to resolve

mjbvz:
See updated PR #826 based on this one. Closing this PR.

Issue #398, opened 2015-08-28
npm refresh hangs on refresh


MarkAurit:
windows 7/64 bit/8G
vs 2013 professional
Ive just created a small 'blank console' project and added express using the npm manager express.  After adding it I noticed there was a red message that the catalog of packages was old, so I hit refresh, and it just hangs on the 'loading published package list' dialog. After a few minutes, if I close it, and then choose the "install new package" option, I just get the same dialog. If I let it go for a total of maybe 5-7 minutes I get a "refresh failed - see output below for details", although there is no output that I can tell.
nothing in C:\Users\<user>\AppData\Local\Microsoft\Node.js Tools\NpmCache looked helpful
SOLVED: closed the project, shutdown vs, opened vs and the project.  Now I get the original
screen when adding a package (including the "updated more than 1 month ago" message,
which apparently I need to ignore).

jamesxv7:
I am experiencing the same issue but using VS2015 on Win8. There is a workaround mentioned by @mousetraps in [another issue](https://github.com/Microsoft/nodejstools/issues/99#issuecomment-99344063) that worked for me but temporarily. It turns out that as soon I erase the cache in **%localappdata%\Microsoft\Node.js Tools\** and click again on Refresh, the window remains *Loading published package list... / Download completed - updating database* eternally. My output log is:
```
Registry url: https://registry.npmjs.org/
Current Time: 9/6/2015 10:14:41 PM
Last Refreshed: 9/6/2015 8:51:53 PM
Number of Results: 192835
Registry url: https://registry.npmjs.org/
Path to npm: C:\Program Files\nodejs\npm.cmd
Downloading package cache to C:\Users\Jaime\AppData\Local\Microsoft\Node.js Tools\NpmCache\1d79c43a-7038-4e54-ac6a-881485af293c\since_packages.json
Downloading list of packages from https://registry.npmjs.org/-/all/since?stale=update_after&startkey=1441586403199
Download complete - updating database
Reading 593255 bytes from C:\Users\Jaime\AppData\Local\Microsoft\Node.js Tools\NpmCache\1d79c43a-7038-4e54-ac6a-881485af293c\since_packages.json (9/6/2015 10:14:57 PM)
```
@mousetraps - Any other information that you need just let me know.

billti:
[...]

PR #437, opened 2015-09-10
NPM catalog schema change causes high CPU (#398)

billti:
This is the fix for the NPM schema change causing the infinite loop and CPU usage - see #398 .

I took the opportunity to reduce the size of some of the test assets for the validation, as they were so huge they took forever to run, and added coverage for the new Array literal schema case. These tests are all passing now.


Nodejs/Product/Npm/SPI/NpmGetCatalogCommand.cs
```diff
@@ -65,115 +65,182 @@ internal class NpmGetCatalogCommand : NpmCommand, IPackageCatalog {
...
-
-                                    InsertCatalogEntry(db, package);
-                                } catch (InvalidOperationException) {
-                                    // Occurs if a JValue appears where we expect JProperty
-                                } catch (ArgumentException) {
-                                    OnOutputLogged(string.Format(Resources.ParsingError, builder.Name));
-                                    if (!string.IsNullOrEmpty(builder.Name)) {
-                                        var package = builder.Build();
-                                        InsertCatalogEntry(db, package);
-                                    }
```
mousetraps: We should continue catching the ArgumentException - occasionally npm changes things about their schema (or some odd package gets published that is contrary to the entire schema) that causes us to parse things incorrectly. For instance, at some point a single package was appearing in the catalog with an array of homepages rather than a single list. 

In these cases, we want to be able to fail gracefully with a small note about what failed rather than failing to retrieve the entire catalog.

Nodejs/Product/Npm/SPI/NpmGetCatalogCommand.cs
```diff
@@ -65,115 +65,182 @@ internal class NpmGetCatalogCommand : NpmCommand, IPackageCatalog {
...
                 });
             }
-        }
-
+        }
+
+        private void ReadPackagesFromObject(SQLiteConnection db, JsonTextReader jsonReader, string registryUrl) {
+            var builder = new NodeModuleBuilder();
+            while (jsonReader.Read()) {
+                if(jsonReader.TokenType == JsonToken.EndObject) {
```
mousetraps: nit: missing spaces between if and parens

Nodejs/Product/Npm/SPI/NpmGetCatalogCommand.cs
```diff
@@ -65,115 +65,182 @@ internal class NpmGetCatalogCommand : NpmCommand, IPackageCatalog {
...
+                }
+
+                // Else the property value should be an object literal representing the package
+                jsonReader.Read();
+                if (jsonReader.TokenType == JsonToken.StartObject) {
+                    IPackage package = ReadPackage(jsonReader, builder);
+                    if (package != null) {
+                        InsertCatalogEntry(db, package);
+                    }
+                }
```
mousetraps: Is there an EndObject token we also need to read in here?

Nodejs/Product/Npm/SPI/NpmGetCatalogCommand.cs
```diff
@@ -65,115 +65,182 @@ internal class NpmGetCatalogCommand : NpmCommand, IPackageCatalog {
...
+            }
+            throw new JsonException("Unexpected end of stream reading the NPM catalog data array");
+        }
+
+        private IPackage ReadPackage(JsonTextReader jsonReader, NodeModuleBuilder builder) {
+            builder.Reset();
+
+            // The JsonTextReader should be positioned at the start of the object literal token for the package
+            var module = JToken.ReadFrom(jsonReader);
```
mousetraps: Is the text reader positioned correctly if the StartObject has already been read?

mousetraps:
left a few comments and lgtm after that

Nodejs/Product/Npm/SPI/NpmGetCatalogCommand.cs
billti: No. ReadPackage will consume the entire token (object literal).

Nodejs/Product/Npm/SPI/NpmGetCatalogCommand.cs
billti: Not sure I follow, but yes it is positioned correctly (hence all the tests passing :wink:

mousetraps: Okay, we should just make sure it still works with vs2013 or remove the conditional altogether and stick wth newtonsoft v6 everywhere (it's been battle tested enough now). There are subtle differences in handling between the two versions and I'm sort of surprised it's both reading through the endobject with ReadFrom, even though we've already read the startobject

Nodejs/Product/Npm/SPI/NpmGetCatalogCommand.cs
billti: OK. Added roughly the same logging/handling into the ReadPackage function.

PR #439, opened 2015-09-10
#388 custom captions don't work with web publish

### LINKED ISSUE #388: NTVS 1.1 folder naming doesn't work with web publish
When folders are renamed from "public" to "public (browser)" or "public (node)" this breaks web publishing as it now publishes files to a "public (browser)" folder instead of "public".

mousetraps:
- The issue was that in some scenarios we were returning the caption,
  rather than the actual name of the file/folder

Fix #388

### DIFF
--- a/Common/Product/SharedProject/HierarchyNode.cs
+++ b/Common/Product/SharedProject/HierarchyNode.cs
@@ -590,7 +590,7 @@ namespace Microsoft.VisualStudioTools.Project {
                     break;
 
                 case __VSHPROPID.VSHPROPID_Name:
-                    result = CommonUtils.GetFileOrDirectoryName(this.Url);
+                    result = this.GetEditLabel();
                     break;
 
                 case __VSHPROPID.VSHPROPID_ExpandByDefault:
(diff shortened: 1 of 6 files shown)


Common/Product/SharedProject/HierarchyNode.cs
```diff
@@ -590,7 +590,7 @@ internal abstract class HierarchyNode :
                     break;
 
                 case __VSHPROPID.VSHPROPID_Name:
-                    result = this.Caption;
+                    result = CommonUtils.GetFileOrDirectoryName(this.Url);
```
billti: If I look though this file for other uses for `this.caption`, then I see it used for `SaveName is the name shown in the Save and the Save Changes dialog boxes` and `GetEditLabel: Called by the shell to get the node caption when the user tries to rename from the GUI`, and `GetRelationalName`. Seems like all of these probably should be using the proper filename/url also, and not the labeled caption.

mousetraps: Yeah, I can change those too - most of those (e.g. GetEditLabel) are overridden in NodejsFolderNode, but may as well fix it in HierarcyNode.

mousetraps: Went ahead and updated references to Caption throughout the project system.

PR #442, opened 2015-09-10
#63 IntelliSense disappears after closing VS when there is an

### LINKED ISSUE #63: intellisense disappears after closing VS when there is an incomplete statement in editor
1) create new console project 
2) npm install async 
3) type code in editor
``` js
var async = require('async')
async.all
```
4) F12 on 'all', and go to definition. (takes you to right place in async.js) 
5) save, close, and reopen project 
6) F12 on 'all' - doesn't work... instead pops up this dialog:
```
---------------------------
Node.js Tools for Visual Studio
---------------------------
Cannot go to definition "async.all"
---------------------------
OK   
---------------------------
And intellisense is completely borked until you delete the .dat file and reload the project. 
```

This issue doesn't seem to occur if the statement is complete

We fixed it earlier, but looks like it's back :( https://nodejstools.codeplex.com/workitem/1706
more discussion: https://nodejstools.codeplex.com/workitem/1963

mousetraps:
incomplete statement in the editor
- Fixes an issue where we were we were inadvertently unloading
  built-in modules from analysis in some scenarios.

Fix #63

### DIFF
--- a/Nodejs/Product/Nodejs/Intellisense/VsProjectAnalyzer.cs
+++ b/Nodejs/Product/Nodejs/Intellisense/VsProjectAnalyzer.cs
@@ -682,7 +682,8 @@ namespace Microsoft.NodejsTools.Intellisense {
             }
 
             foreach (var item in _projectFiles) {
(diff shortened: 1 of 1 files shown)


billti:
Good find!! :+1:

Issue #179, opened 2015-06-06
VS Reformats my source code when I type, even though I told it not to


LegalizeAdulthood:
Even though I have this turned off in Tools / Options, since installing the nodejs tools for VS, VS is now reformatting my code (changing indentation, adding whitespace) whenever I type `;` and during other operations.

This is very annoying.  My code is written the way **I want** and VS should never be fiddling around with my code simply because I'm typing.

I tried toggling the option in case it simply ignored the setting on first time install, but that didn't help.

This is really, really, really annoying.  Now I have to undo every time VS decides to reformat my code which is all the time that I'm typing.  Grrr.

mousetraps:
Did you turn it off in JavaScript formatting options or Node.js formatting options? They are confusinfly in different spots at the moment.

LegalizeAdulthood:
You've got all the formatting options duplicated?!?!?

mousetraps:
Yeeeahhhh... its pretty non-ideal... To provide a little more context...

They diverged when we stopped using the JS language service available in the rest of VS in favor of our own Node.js engine that does static analysis because JSLS didn't scale well to the large node projects people were building. So our formatter and intellisense actually behaves differently and requires different options. 

See #67 for more info (and please contribute to the discussion there.)

LegalizeAdulthood:
Is this also why Ctrl+SPACE doesn't show intellisense suggestions if I delete all the characters in the "complete intellisense when I type these characters" box?

I just want to type my code and **only** use intellisense when I ask for it.  I find the constant popping up of "suggestions" annoying and distracting.  I certainly don't want to be typing code, such as calling a method that doesn't yet exist because I'm about to write it, and have the open parent on my as-yet-non-existing method call tell intellisense to **change** my existing correct code into incorrect code.  This is the most annoying thing about intellisense in IDEs, is that they don't let me simply **write correct code** without constantly molesting it along the way.

When I've done training classes, it was uniformly the case that people who relied on intellisense to "complete" their code every step of the way took the longest to complete any given exercise.  Intellisense should be an assistant, not a crutch and certainly not a backseat driver constantly nagging me to turn here or turn there when I know exactly where I'm going.

LegalizeAdulthood:
Also, apparently, once I've deleted all those characters from that box, there's no way to reset it to the default without resetting all my options to the defaults globally?  that sucks.  But maybe that's just VS sucking.

LegalizeAdulthood:
[...]

PR #445, opened 2015-09-13
#310 Default to requiring tab to trigger a completion

### LINKED ISSUE #310: Default to requiring tab to trigger a completion
Intellisense is not always perfect, so it can be annoying when typing in a method name manually and it auto-selects a different one from the completion list.

Similar to other JavaScript editors, the default should be to accept what the user wrote, not a closest match from the list, unless an explicit completion trigger (i.e. TAB) is selected.

mousetraps:
Remove the custom committed-by characters setting and also default to
requiring tab/enter triggering a completion. This simplifies settings,
is consistent with both C# and JS projects in VS2015, and makes it easier
to work with identifiers that don't appear in the completions list.

Fix #310, related to #179

### DIFF
--- a/Nodejs/Product/Nodejs/Commands/DiagnosticsCommand.cs
+++ b/Nodejs/Product/Nodejs/Commands/DiagnosticsCommand.cs
@@ -165,7 +165,7 @@ namespace Microsoft.NodejsTools.Commands {
                 }
             }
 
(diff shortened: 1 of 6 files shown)


mousetraps:
screenshot: image

billti:
:+1: Please wait for a code review and thumbs up this close to RTM though before merging. Thanks!

PR #448, opened 2015-09-14
#246, #362 - Debugger listening but program not starting, flakey debugger behavior

### LINKED ISSUE #246: Debugger Doesn't Work 75% Of The Time

### LINKED ISSUE #362: Debugger listening but program not starting
Sometimes when i choose to debug a project (debug -> Start Debugging or pressing F5), i will se "Debugger listening on port 5858" in the console window, but the program will not actually run. If i add a breakpoint to first line in the code, throw an exception on the first line or whatever, it will never get so far, no code is executed. If i stop and start the program multiple times it will eventually run, but this is a rather annoying process. I have nodejs v0.12.7. Tried visual studio 2013 and visual studio 2015. Upgraded from windows 8.1 to 10 and it is the same on both platforms. 
error

mousetraps:
The issue in this case stems from sending an AD7LodCompleteRunningEvent
when we should be sending an AD7LoadCompleteEvent. In particular,
_processLoadedRunning is set to true here even though execution ends up
being paused later.

The fix is to only load the "running" value after modules are loaded,
rather than beforehand.
are loaded, so load the property then rather than beforehand.

Fix #362, fix #246


Nodejs/Product/Nodejs/Debugger/DebugEngine/AD7Engine.cs
```diff
@@ -1152,6 +1149,7 @@ public sealed class AD7Engine : IDebugEngine2, IDebugEngineLaunch2, IDebugProgra
                 if (_loadComplete) {
                     SendModuleLoad(adModule);
                 }
+
```
billti: What's this line for?

Nodejs/Product/Nodejs/Debugger/NodeDebugger.cs
```diff
@@ -243,8 +243,16 @@ public NodeDebugger(Uri debuggerEndpointUri, int id)
             if (asyncBreakComplete != null) {
                 asyncBreakComplete(this, new ThreadEventArgs(MainThread));
             }
-        }
-
+        }
+
+        internal bool IsRunning() {
+            var backtraceCommand = new BacktraceCommand(CommandId, _resultFactory, 0, 1);
```
billti: What are 0 & 1? Can we use named args to make the code clearer?

Nodejs/Product/Nodejs/Debugger/NodeDebugger.cs
```diff
@@ -243,8 +243,16 @@ public NodeDebugger(Uri debuggerEndpointUri, int id)
             if (asyncBreakComplete != null) {
                 asyncBreakComplete(this, new ThreadEventArgs(MainThread));
             }
-        }
-
+        }
+
+        internal bool IsRunning() {
+            var backtraceCommand = new BacktraceCommand(CommandId, _resultFactory, 0, 1);
+            if (TrySendRequestAsync(backtraceCommand).GetAwaiter().GetResult()) {
```
billti: I see some of the below commands have a timeout set. Is there potential this will block? For how long? Is there a chance this will be on the UI thread?

Nodejs/Product/Nodejs/Debugger/DebugEngine/AD7Engine.cs
mousetraps: Makes it easier to read between the lines. I'll remove it :smiley:

Nodejs/Product/Nodejs/Debugger/NodeDebugger.cs
mousetraps: This is actually the same code as before, and I wasn't sure why it didn't have the timeout set before, so I just left it as is. I can add a timeout, though.

PR #456, opened 2015-09-14
#454 Don't show completions after "var" keyword

### LINKED ISSUE #454: Don't show completion list after "var"
the user is defining a new variable here, so showing the completion list is unnecessary.

mousetraps:
Fix #454

### DIFF
--- a/Nodejs/Product/Nodejs/Intellisense/VsProjectAnalyzer.cs
+++ b/Nodejs/Product/Nodejs/Intellisense/VsProjectAnalyzer.cs
@@ -1198,11 +1198,10 @@ namespace Microsoft.NodejsTools.Intellisense {
                 // Check for context-sensitive intellisense
                 var lastClass = tokens[tokens.Count - 1];
 
-                if (lastClass.ClassificationType == classifier.Provider.Comment) {
-                    // No completions in comments
-                    return CompletionAnalysis.EmptyCompletionContext;
-                } else if (lastClass.ClassificationType == classifier.Provider.StringLiteral) {
-                    // String completion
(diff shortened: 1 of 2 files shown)


Nodejs/Product/Nodejs/Intellisense/VsProjectAnalyzer.cs
```diff
@@ -1198,11 +1198,10 @@ CollectingErrorSink errorSink
...
 
-                if (lastClass.ClassificationType == classifier.Provider.Comment) {
-                    // No completions in comments
-                    return CompletionAnalysis.EmptyCompletionContext;
-                } else if (lastClass.ClassificationType == classifier.Provider.StringLiteral) {
-                    // String completion
+                if (lastClass.ClassificationType == classifier.Provider.Comment ||
+                    lastClass.ClassificationType == classifier.Provider.StringLiteral ||
+                    lastClass.Span.GetText() == "var") {
+                    // No completions in comments, strings, or after "var" keywords.
```
billti: This seems a little odd. For comment and strings it's checking you are _in_ that token (i.e. inside a comment or a string), not that it's the prior token, which is what it's looking for with 'var'. Did you test this and it behaves as expected? (Can we add a unit test for it?).

Also, is it possible to check the prior token was the _keyword_ `var`, not just that the text was `var`? `var` is actually a valid identiferName, just not a valid identifier (i.e. `var var` is illegal, but something like `obj.var` is perfectly valid).

billti:
:+1:

mousetraps:
Added a unit test and comments to make the behavior clearer.

PR #457, opened 2015-09-14
#345 Include option to show completion list after a character is typed

### LINKED ISSUE #345: include option to immediately display intellisense while typing without pressing ctrl space
For instance, typing 'v' should display all completions matching "v". Currently you would have to type "v<ctrl+space>" 

This is most well known as: "Show completion list after a character is typed"

copied from: https://nodejstools.codeplex.com/workitem/1719

mousetraps:
- trigger completion by default when both AutoListMembers and
  ShowCompletionListAfterCharacterTyped options are selected
- adds option to options page

Fix #345

### DIFF
--- a/Nodejs/Product/Nodejs/Intellisense/IntellisenseController.cs
+++ b/Nodejs/Product/Nodejs/Intellisense/IntellisenseController.cs
@@ -193,6 +193,13 @@ namespace Microsoft.NodejsTools.Intellisense {
                             UpdateCurrentParameter();
                         }
                         break;
+                    default:
+                        if (IsIdentifierFirstChar(ch) && _activeSession == null
+                            && NodejsPackage.Instance.LangPrefs.AutoListMembers
+                            && NodejsPackage.Instance.IntellisenseOptionsPage.ShowCompletionListAfterCharacterTyped) {
+                            TriggerCompletionSession(false);
+                        }
+                        break;
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


Nodejs/Product/Nodejs/Intellisense/IntellisenseController.cs
```diff
@@ -725,8 +732,12 @@ out title
                 || (ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z') || (ch >= '0' && ch <= '9');
         }
 
+        private static bool IsIdentifierFirstChar(char ch) {
+            return ch == '_' || (ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z');
+        }
```
billti: Do such helper methods not already exist elsewhere in the parser code?

Also, interesting to see this is all ASCII only (not just your code changes, but surrounding code). Does NTVS not handle Unicode chars? (ES5 allows identifierNames as per the Unicode 3 standard - see http://es5.github.io/#x7.6 ).

billti:
This will be cool to get in, but changes in these areas are really hard to get just right (just ask @paulvanbrenk how many times he had to redo it for TypeScript :smile: ). Let's not check this in before the last RC and RTM.

Nodejs/Product/Nodejs/Intellisense/IntellisenseController.cs
```diff
@@ -193,6 +193,13 @@ internal sealed class IntellisenseController : IIntellisenseController, IOleComm
                             UpdateCurrentParameter();
                         }
                         break;
+                    default:
+                        if (IsIdentifierFirstChar(ch) && _activeSession == null
```
billti: Did you test this when typing in comments, string literals, etc...? I don't see any filtering here on context (e.g. see some of the checks above like on line 169). It's super annoying when the completion list keeps appears when you're typing in places you don't want it (like those just listed). I've seen those bugs before (I may have written a couple of them :-/ )

mousetraps: Yeah, the other isense code actually blocks that (reverseExpressionParser and VsAnalyzer.) otherwise Ctrl+space would trigger in comments etc.

billti:
:+1:

PR #460, opened 2015-09-15
iterated according to CR comments

mousetraps:
iterated based on CR comments in #448
- cleaned up code
- added timeout to backtraceCommand


billti:
:+1: Thanks!

PR #463, opened 2015-09-16
#455 IntelliSense commits too aggressively when completion is selected

### LINKED ISSUE #455: InteliSense commits too aggressively when completion is selected
```js
var variable = 3;
<ctrl+space>varia;<enter>
```

expected:
```js
var variable = 3;
varia;
<cursor>
```

actual:
```js
var variable = 3;
variable<cursor>
```

mousetraps:
- do not enable COMPLETEWORD when OnlyTabOrEnterToCommit option is
  enabled

fix #455

### DIFF
--- a/Nodejs/Product/Nodejs/Intellisense/IntellisenseController.cs
+++ b/Nodejs/Product/Nodejs/Intellisense/IntellisenseController.cs
@@ -647,7 +647,8 @@ namespace Microsoft.NodejsTools.Intellisense {
                         case VSConstants.VSStd2KCmdID.COMPLETEWORD:
                             ForceCompletions = true;
                             try {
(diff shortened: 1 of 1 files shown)


billti:
So this just causes it to show the list even if only one member? (I noticed this, but didn't think it was a big deal). :+1: if that's the only behavior change.

mousetraps:
Yep, that's the only behavior change - it's a bigger deal when IntelliSense is being automatically displayed because typing "const<space>" automatically completes it to "constructor" since "const" isn't in the list.

mousetraps:
Gah, sorry I conflated two examples. The const thing applied only to one of the unit tests where ctrl+space was used to grab the completions list since spaces don't actually cause use to commit. But the user-facing issue was when assigning to a variable that didn't exist in isense.

PR #464, opened 2015-09-16
#462 completions should not be displayed after funciton, let, const

### LINKED ISSUE #462: completions should not be displayed after function, let, const keywords
This is similar to #454

mousetraps:
keywords
- ignore additional keywords when providing completions and update unit tests

fix #462


mousetraps:
without line ending differences:
https://github.com/Microsoft/nodejstools/pull/464/files?w=1

billti:
:+1:

Issue #192, opened 2015-06-09
Array, Number, and maybe other constructors are not showing full list of completions


mousetraps:
Instead it is only showing ```Array()``` in the completion list.

kant2002:
According to my research, this code somewhere in the [Nodejs\Product\Analysis\Analysis\GlobalBuilder.cs](https://github.com/Microsoft/nodejstools/blob/master/Nodejs/Product/Analysis/Analysis/GlobalBuilder.cs#L222)

PR #477, opened 2015-09-22
Show full list of completions for Array/String/Number/Boolean/etc constructors

hoanhtien:
Fixes https://github.com/Microsoft/nodejstools/issues/192.


kant2002:
When this would be added ?

billti:
We were pretty conservative in taking changes while stabilizing for the final 1.1 release. I expect we'll start cleaning up the outstanding pull requests shortly. 

FYI: We've had a few folks out on training and vacation the past couple weeks, so progress has been a little slow.

kant2002:
Could this be added, since 1.1 is released now?

mjbvz:
@mousetraps Just looking through some of these outstanding pull requests. This one doesn't have any conflicts and the code looks good enough to me. Can you please take a quick look over it too so we can decide to either merge or close this PR?

Issue #469, opened 2015-09-19
VS2015 crashes when closing project


kant2002:
When I close the NTVS project, VS crashes. I wait when WER collect information. Hopefully it would be available to you.

In case if this is related, I open project in VS, then close it, and reopen again, then I next time close project, VS crashes.

**Environment**
VS2015
NTVS 9-17-2015 Dev Build

Node 0.12
NPM 3.3.4

**Event Viewer**
```
Faulting application name: devenv.exe, version: 14.0.23107.0, time stamp: 0x559b7ead
Faulting module name: unknown, version: 0.0.0.0, time stamp: 0x00000000
Exception code: 0xc0000005
Fault offset: 0x181b89f6
Faulting process id: 0xbe0
Faulting application start time: 0x01d0f2ea2039beb3
Faulting application path: C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\devenv.exe
Faulting module path: unknown
Report Id: cbdcf588-45bd-4825-b741-0f437d5d9e41
Faulting package full name: 
Faulting package-relative application ID: 
```

mousetraps:
Is that all the information available in the event viewer?

kant2002:
Initially I start from reporting page, so I miss `.NET Runtime` event source

```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.NullReferenceException
Stack:
   at Microsoft.NodejsTools.Project.NodejsProjectNode.OnIdleNodeModules(System.Object)
   at System.Threading.TimerQueueTimer.CallCallbackInContext(System.Object)
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.TimerQueueTimer.CallCallback()
   at System.Threading.TimerQueueTimer.Fire()
   at System.Threading.TimerQueue.FireNextTimers()
   at System.Threading.TimerQueue.AppDomainTimerCallback()

```

kant2002:
Also I look earleir in the Event log, and exactly the same exception happens to me earlier at least 2 times.

[...]

PR #486, opened 2015-09-24
Dispose OnIdle timer when closing a Nodejs project.

hoanhtien:
Related to https://github.com/Microsoft/nodejstools/issues/469.

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
@@ -122,7 +122,9 @@ namespace Microsoft.NodejsTools.Project {
                     }
                 }
                 var fileNode = DelayedAnalysisQueue.Dequeue();
-                fileNode.Analyze();
+                if (fileNode != null) {
+                    fileNode.Analyze();
+                }
             }
         }
 
@@ -974,6 +976,10 @@ namespace Microsoft.NodejsTools.Project {
                     }
 
                     _analyzer = null;
+                }
+
+                if (_idleNodeModulesTimer != null) {
+                    _idleNodeModulesTimer.Dispose();
                 }
 
                 NodejsPackage.Instance.IntellisenseOptionsPage.SaveToDiskChanged -= IntellisenseOptionsPageSaveToDiskChanged;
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -976,6 +978,10 @@ internal struct LongPathInfo {
                     _analyzer = null;
                 }
 
+                if (_idleNodeModulesTimer != null) {
+                    _idleNodeModulesTimer.Dispose();
```
mousetraps: In addition to disposing the timer, we should also unsubscribe to the event handler before disposing (otherwise the callbacks that are already scheduled will continue to fire).

mousetraps:
minor comment and lgtm after that

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
hoanhtien: Can you elaborate? I am not seeing any event handler related to this Timer.

mousetraps: The following event listeners on fileWatcher are subscribed to, but never unsubscribed from - these trigger the timer. 
```cs
    _nodeModulesWatcher.Changed += OnNodeModulesWatcherChanged;
    _nodeModulesWatcher.Created += OnNodeModulesWatcherChanged;
    _nodeModulesWatcher.Deleted += OnNodeModulesWatcherChanged;
```

mousetraps: Additionally just looked at the code again, there are a couple other issues we'll need to sort out. 
* Additionally, we should dispose the timer and unsubscribe under a lock to prevent race conditions (you can use the _idleNodeModulesLock)
* also set the timer to null during the dispose process, and check for null before _idleNodeModulesTimer.Change otherwise we'll run into an ObjectDisposedException there.

hoanhtien: Updated the pull request. Do we need to dispose _nodeModulesWatcher?

Issue #430, opened 2015-09-08
Add from existing code should ignore any filepaths that include node_modules and bower_components folders by default


mousetraps:
This includes instances where node_modules is not at the root of the project. Additionally, we should remove the checkbox that gives people the option to shoot themselves in the foot by including node_modules (which we explicitly advise against doing when you try to perform that action on a node_modules folder node)

PR #495, opened 2015-09-29
Always exclude node modules

hoanhtien:
* Remove 'exclude node_modules' checkbox. Only the checkbox. I am not removing the data model behind the screen e.g. ExcludeNodeModulesProperty in case we need it back in the future.
* Exclude node_modules subfolders.
* Exclude bower_components folder/subfolders.

This fixes https://github.com/Microsoft/nodejstools/issues/430.


Nodejs/Product/Nodejs/Project/ImportWizard/FileSourcePage.xaml
```diff
@@ -52,12 +52,5 @@
                 </ListView.ItemTemplate>
             </ListView>
         </wpf:LabelledControl>
-        
-        <wpf:LabelledControl Title="Select any other options for your project."
-                           Grid.Row="1">
-            <StackPanel Orientation="Vertical">
-                <CheckBox IsChecked="{Binding ExcludeNodeModules}">Exclude node__modules</CheckBox>
```
mousetraps: You should remove the ExcludeNodeModules binding as well if it's no longer being used.

Nodejs/Product/Nodejs/Project/ImportWizard/ImportSettings.cs
```diff
@@ -244,6 +244,16 @@ internal class ImportSettings : DependencyObject {
             // We won't reject the folder.
             Debug.Assert(!Path.IsPathRooted(dirName));
             return !dirName.Split(new char[] { '/', '\\' }).Any(name => name.StartsWith("."));
+        }
+
+        /// <summary>
+        /// Checks whether a relative and double-backslashed seperated path contains a folder name.
+        /// </summary>
+        private static bool ContainsNodeModulesOrBowerComponentsFolder(string path)
+        {
```
mousetraps: braces on same line here and everywhere else

Nodejs/Product/Nodejs/Project/ImportWizard/FileSourcePage.xaml
mousetraps: Also verify that the ImportWizardTests continue to work.

hoanhtien:
Addressed feedback.

Issue #481, opened 2015-09-24
LiveLogger broken when NodejsPackage.Instance not initialized


mousetraps:
See https://github.com/Microsoft/nodejstools/pull/410#commitcomment-13084935 for more details

hoanhtien:
VS crashes if I reverse my commit.
```C#
private LiveLogger() {
    //if (NodejsPackage.Instance != null) {
        _diagnosticsOptions = NodejsPackage.Instance.DiagnosticsOptionsPage; // crashes here as NodejsPackage.Instance is null
    //}
}
```

PR #497, opened 2015-09-30
Allow LiveLogger to be initialized without the need of NodejsPackage.Instance.

hoanhtien:
This is done by moving _diagnosticsOptions initialization away from LiveLogger constructor.

Fixes https://github.com/Microsoft/nodejstools/issues/481.

### DIFF
--- a/Nodejs/Product/Nodejs/Logging/LiveLogger.cs
+++ b/Nodejs/Product/Nodejs/Logging/LiveLogger.cs
@@ -32,11 +32,19 @@ namespace Microsoft.NodejsTools.Logging {
         private static volatile LiveLogger _instance;
         private static object _loggerLock = new object();
 
-        private NodejsDiagnosticsOptionsPage _diagnosticsOptions = null;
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/Logging/LiveLogger.cs
```diff
@@ -32,11 +32,19 @@ internal sealed class LiveLogger {
...
 
         private LiveLogger() {
-            if (NodejsPackage.Instance != null) {
-                _diagnosticsOptions = NodejsPackage.Instance.DiagnosticsOptionsPage;
+        }
+
+        private NodejsDiagnosticsOptionsPage DiagnosticsOptions {
+            get {
+                if (_diagnosticsOptions == null) {
+                    if (NodejsPackage.Instance != null) {
```
mousetraps: simplify this check so that it's `if(_diagnosticsOptions == null && NodejsPackage.Instance != null)`

mousetraps:
minor comment and lgtm after that

Issue #285, opened 2015-07-14
When close VS/Solution Node windows not closed 


kant2002:
The application which run via `Ctrl+F5` (maybe this is important) not closed after I close VS.
think that after closing solution the NodeJS processes should be terminated, to not leave junk processes in the system.

mousetraps:
Indeed, Ctrl+F5 is important her because the process is terminated when the debugger is detached, whereas with Ctrl+F5 we don't maintain any reference to it at all (we just spawn off a separate process)

kant2002:
I understand why this is happens, but this is not behavoiur which I experience working with C# project. When I close solution the IISExpress wesite terminated, and eventually IISExpress is shutting down.

mousetraps:
Oh, sure, absolutely agreed we should fix it :smiley: I was mostly commenting on the "maybe this is important" parenthetical remark.

kant2002:
Any news on the PR?

billti:
I haven't had a chance to try it out yet, but on quick glance at the code review I have my doubts (as it wires up a debugger event to kill the process, but this is specifically about Ctrl+F5 which doesn't attach the debugger). 

Is this a high-pri issue for you? I assume you can still see the console window spawned and just close it.

kant2002:
That's definitely not hi-prority for me, but fix seems so simple, and I probably implement it in the same way, so since change come from MS, I think it would be incorporated more fast. Don't get me wrong, but I constantly see NTVS not as polished as their C# counterparts. Actually such long review time discouraging me from fixing such small things.

PR #500, opened 2015-09-30
Terminate existing node process when project is being disposed.

hoanhtien:
Add OnDispose event to NodejsProcessNode and let NodeProcess subscribe to it.

Fixes https://github.com/Microsoft/nodejstools/issues/285.


mjbvz:
Resubmitted as #850

PR #512, opened 2015-10-06
#511 npm tree incomplete when there are duplicate entries

### LINKED ISSUE #511: npm tree incomplete when there are duplicate entries
After installing express, jade, and stylus... only the express node is expandable when we expect all three dependencies to be expandable.

mousetraps:
- Always add top-level dependencies even if they've already been added
  before

fix #511


mousetraps:
hrm, this might be awakening some existing perf issues in the project system that we should investigate before pulling in.

PR #515, opened 2015-10-06
Fix test case: TestInsertSnippet 

hoanhtien:
Regarding the following snippet test:
```C#
new Snippet(
   "iife",
    "(function (undefined) {\r\n    $body$\r\n})();\r\n",
    new Declaration("name","(function (name) {\r\n    $body$\r\n})();\r\n")
),
```

Typing the declaration "name" then hit "\t" will trigger auto completion which replaces "name" with "__dirname":

abc

As we still want "name" in the expected result, change to use right arrow instead of tab "\t" to avoid auto completion.

### DIFF
--- a/Nodejs/Tests/Core.UI/SnippetsTests.cs
+++ b/Nodejs/Tests/Core.UI/SnippetsTests.cs
@@ -82,7 +82,6 @@ namespace Microsoft.Nodejs.Tests.UI {
         [HostType("VSTestHost")]
         public void TestInsertSnippet() {
             using (var solution = BasicProject.Generate().ToVs()) {
-                    
                 foreach (var snippet in BasicSnippets) {
                     TestOneInsertSnippet(solution, snippet, "Nodejs");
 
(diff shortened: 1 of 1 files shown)


Nodejs/Tests/Core.UI/SnippetsTests.cs
```diff
@@ -215,14 +214,12 @@ class Declaration {
             foreach (var decl in snippet.Declarations) {
                 Console.WriteLine("Declaration: {0}", decl.Replacement);
                 Keyboard.Type(decl.Replacement);
-                Keyboard.Type("\t");
+                Keyboard.Type("→");
```
mousetraps: You can use ExecuteCommand('Edit.CompleteWord') here instead. Also let's change the rest of the tests to be consistent if we're going to change it in this one spot.

hoanhtien: Edit.CompleteWord does not work here. Edit.CompleteWord only completes the word if there is only one match in the completion set.

mjbvz:
Closing since the PR does not seem to actually fix the test in my testing.

Issue #517, opened 2015-10-08
Description for keywords  is shown as 'unknown type' in completion list.


hoanhtien:
![image](https://cloud.githubusercontent.com/assets/1707813/10355600/e57a259a-6d22-11e5-83a4-470f21bf6d69.png)

PR #520, opened 2015-10-08
Use keyword name as its auto-complete documentation

hoanhtien:
Fixes https://github.com/Microsoft/nodejstools/issues/517.


mousetraps:
:+1:

PR #525, opened 2015-10-12
#467 VS hangs after closing NTVS project

### LINKED ISSUE #467: VS2015 hangs after closing NTVS project
Hi,
First off, I want to say that NTVS is simply amazing! It's the most advanced way to develop and debug Node.js apps as of today. The IntelliSense is so nice to have! And so is the "edit-and-continue". No other IDE is able to match these features, and I tested a lot of them. IntelliSense for JS is extremely hard to pull off. Well done.

One thing that happens randomly is VS2015 freezing when closing a NTVS solution. It happens about 80% of the time. Everything works great until I'm done with the project and want to close it. VS will hang and won't exit even after minutes. Only a process kill via Task Manager to `devenv.exe` can close VS.

Is this a known issue? I tried searching and couldn't find it.

Thanks for your support,
Elad

### LINKED ISSUE #505: Hang while populating completion list
**app.js**
```js
var c= requestId.// long hang after pressing .
```

**package.json:**
```json
{
  "name": "App",
  "main": "app.js",
  "version": "1.0.0",
  "dependencies": {
    "async": "^0.9.0",
    "body-parser": "^1.10.1",
    "cookie-parser": "^1.3.4",
    "crypto": "0.0.3",
    "express": "~4.9.0",
    "mongoose": "^4.1.7",
    "mqtt": "^0.3.13",
    "nodemailer": "^1.3.4",
    "nodemailer-ses-transport": "^1.2.0",
    "path": "^0.11.14",
    "request": "^2.51.0",
    "stripe": "^3.3.4",
    "validator": "^4.0.5"
  },
  "devDependencies": {
    "jshint": "^2.8.0"
  }
}
````

Additionally, it looks like way too many types are flowing through here:
image

### PR DESCRIPTION
**Pull request by mousetraps**
#505 Hang while populating completion list
- we weren't respecting MaxMergeTypes everywhere during analysis
- also fix InvalidCastException in GlobalBuilder.Require

(I found some other places where this may be an issue, but I'm not 100% sure, so I'll have to ask Dino about them verify.)

fix #467, fix #505

### DIFF
--- a/Nodejs/Product/Analysis/Analysis/GlobalBuilder.cs
+++ b/Nodejs/Product/Analysis/Analysis/GlobalBuilder.cs
@@ -1310,8 +1310,7 @@ on that object, and are not inherited from the object's prototype. The propertie
 
         private static IAnalysisSet DefineProperties(FunctionValue func, Node node, AnalysisUnit unit, IAnalysisSet @this, IAnalysisSet[] args) {
             // object, {propName: {desc}, ...}
-            var maxMergeTypes = unit.Analyzer.Limits.MaxMergeTypes;
-            if (args.Length >= 2 && args[0].Count < maxMergeTypes && args[1].Count < maxMergeTypes) {
+            if (args.Length >= 2 && args[0].Count < unit.Analyzer.Limits.MaxMergeTypes) {
                 foreach (var obj in args[0]) {
                     ExpandoValue target = obj.Value as ExpandoValue;
                     if (target != null) {
--- a/Nodejs/Product/Analysis/Analysis/GlobalBuilder.cs
+++ b/Nodejs/Product/Analysis/Analysis/GlobalBuilder.cs
@@ -1310,7 +1310,8 @@ on that object, and are not inherited from the object's prototype. The propertie
 
         private static IAnalysisSet DefineProperties(FunctionValue func, Node node, AnalysisUnit unit, IAnalysisSet @this, IAnalysisSet[] args) {
             // object, {propName: {desc}, ...}
-            if (args.Length >= 2) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: Nodejs/Product/Analysis/Analysis/GlobalBuilder.cs [hunk 1]
```diff
@@ -1310,7 +1310,8 @@ class GlobalBuilder {
 
         private static IAnalysisSet DefineProperties(FunctionValue func, Node node, AnalysisUnit unit, IAnalysisSet @this, IAnalysisSet[] args) {
             // object, {propName: {desc}, ...}
-            if (args.Length >= 2) {
+            var maxMergeTypes = unit.Analyzer.Limits.MaxMergeTypes;
```
billti: I assume MaxMergeTypes always has some positive integer then - regardless of analysis settings?

billti:
One question, else based on my limited knowledge :+1:

Nodejs/Product/Analysis/Analysis/GlobalBuilder.cs
mousetraps: yep - in fact, this value is currently the same with every setting.

Nodejs/Product/Analysis/Analysis/GlobalBuilder.cs
```diff
@@ -1310,7 +1310,8 @@ class GlobalBuilder {
...
+            if (args.Length >= 2 && args[0].Count < maxMergeTypes && args[1].Count < maxMergeTypes) {
```
mousetraps: Only check args[0] here

PR #526, opened 2015-10-12
#70 Visual Studio crashes when I quit debugging

### LINKED ISSUE #70: Visual Studio crashes when I quit debugging
I'm running Visual Studio 2013 Community.  I've previously had VS Express 2012 and 2013 (various versions) and tried out VS 2015 Preview and CTP 6 when VS started crashing on me.

Since then I uninstalled all versions of VS, though `\Program Files (x86)\Microsoft Visual Studio X.0\` directories exist for 10, 11, 12, and 14 still.  I then installed VS Community 2013 "fresh" (as fresh as it can be with those lingering files).

I installed NTVS for VS 2013 (after uninstalling it, and then removing `\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\Extensions\Microsoft\Node.js Tools*` manually, since uninstalling did not do it).  Now I can repro this crashing every time with the following steps (before it may have been intermittent?):

Open VS, create a new "Basic Node.js Express 4 Application" project, press F5.  The browser pops up with the expected text.  Close the Browser.  Stop debugging.  VS crashes.

It's quite frustrating.  Until a couple days ago (when I started messing with the Babel package, attempting to get ECMAScript 6 features to work), everything was working great.

Let me know what more information you need.

Thanks!

mousetraps:
- Fix potential NullReferenceExceptions

fix #70

### DIFF
--- a/Nodejs/Product/Analysis/Analysis/Analyzer/AnalysisUnit.cs
+++ b/Nodejs/Product/Analysis/Analysis/Analyzer/AnalysisUnit.cs
@@ -188,9 +188,9 @@ namespace Microsoft.NodejsTools.Analysis {
         }
 
         internal virtual void AnalyzeWorker(DDG ddg, CancellationToken cancel) {
-            if (Tree != ProjectEntry.Tree) {
-                // we were enqueued and a new version became available, don't re-analyze against
-                // the old version.
+            if (Ast != null && ProjectEntry != null && ProjectEntry.Tree != null && Tree != ProjectEntry.Tree) {
+                // analysis unit properties are invalid or we were enqueued and a new version became available
+                // don't re-analyze against the old version.
                 return;
             }
 
--- a/Nodejs/Product/Analysis/Analysis/Analyzer/AnalysisUnit.cs
+++ b/Nodejs/Product/Analysis/Analysis/Analyzer/AnalysisUnit.cs
@@ -188,7 +188,10 @@ namespace Microsoft.NodejsTools.Analysis {
         }
 
         internal virtual void AnalyzeWorker(DDG ddg, CancellationToken cancel) {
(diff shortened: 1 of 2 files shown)


Nodejs/Product/Analysis/Analysis/Analyzer/AnalysisUnit.cs
```diff
@@ -188,9 +188,9 @@ internal AnalysisUnit(JsAst ast, EnvironmentRecord environment)
         }
 
         internal virtual void AnalyzeWorker(DDG ddg, CancellationToken cancel) {
-            if (Tree != ProjectEntry.Tree) {
-                // we were enqueued and a new version became available, don't re-analyze against
-                // the old version.
+            if (Ast != null && ProjectEntry != null && ProjectEntry.Tree != null && Tree != ProjectEntry.Tree) {
```
billti: Should these ever be null? Is it safe to fall through to the remaining code if they are?

billti:
Minor comment, else :+1:

Nodejs/Product/Analysis/Analysis/Analyzer/AnalysisUnit.cs
mousetraps: I couldn't reproduce this issue, so I'm not entirely sure (there are null checks for these properties elsewhere in the code, so I presumed it would be safe enough here.) This is something I plan on asking dino today before merging in.

mousetraps: I'm going to at least Assert here, so that we cant catch it in debug builds.

mousetraps: Okay... just reviewed this again with my alter ego from the future, and realized I was being an idiot. I intended to null check and *not* to flow through, and I totally misinterpreted your comment. fixed :smiley:

PR #527, opened 2015-10-12
Fixed Process schema version and 0.12 record offset

billti:
Fixes an off-by-one field mapping in Node >= 0.12.0.


mousetraps:
:+1:

Issue #514, opened 2015-10-06
IntelliSense closes after typing some characters like '$'


mousetraps:
Given the scenario below, the IntelliSense popup unexpectedly closes after typing `$`. It reappears after typing `_`
![image](https://cloud.githubusercontent.com/assets/762848/10318131/e1e3e3dc-6c1c-11e5-83f0-dca7d0ef4484.png)

hoanhtien:
Something to do with the following method:

```C#
private static bool IsIdentifierFirstChar(char ch) {
    return ch == '_' || (ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z');
}
```

PR #528, opened 2015-10-12
Add $ to IsIdentifierFirstChar list.

hoanhtien:
Quick fix for https://github.com/Microsoft/nodejstools/issues/514. Technically JavaScript identifier names can contain many other Unicode characters such as π, Σ, λ. We may want to support these in the future.


mousetraps:
:+1:

PR #532, opened 2015-10-13
Fix import wizard tests.

hoanhtien:
+ Add null check for NodejsPackage.Instance.
+ In TypeScript project, StartUpFile should be server.ts instead of server.js.

All ImportWizard* test cases should be passing after these changes (9/12 are failing right now).


mousetraps:
:+1:

PR #539, opened 2015-10-14
Fix ProblematicPackageJsonTests.

hoanhtien:
Exceptions are expected for MissingTralingBrace and MissingTrailingListBrace tests.

### DIFF
--- a/Nodejs/Tests/NpmTests/ProblematicPackageJsonTests.cs
+++ b/Nodejs/Tests/NpmTests/ProblematicPackageJsonTests.cs
@@ -129,11 +129,13 @@ namespace NpmTests {
         }
 
         [TestMethod, Priority(0)]
+        [ExpectedException(typeof(PackageJsonException))]
         public void TestParseMissingTrailingBrace() {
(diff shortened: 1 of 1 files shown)


mousetraps:
This fixes the tests in VS2015, but breaks them in VS2013.

The exception only began ocurring in VS2015, which depends on Newtonsoft.Json 6.0, not 4.5 - see this commit comment for more info. https://github.com/Microsoft/nodejstools/commit/155a3411e813d79af7a92c71f0e41bcd9621ca02#diff-783b0bc1c65b5cb6b1ee09a119cd29eb 

As mentioned in the comment, Newtonsoft.Json 6.0 reliance in NTVS was only supposed to be temporary, but unfortunately VS 2015 never fixed their binding redirect issue. At this point, we may as well bite the bullet and move to Newtonsoft.Json 6.0 everywhere rather than conditionally in VS2015 to keep our codepaths more predictable across VS versions.

mousetraps:
Closing this PR, as it's not the right way to go about addressing the issue. Opened #645 to track.

PR #541, opened 2015-10-16
Fix ProfilingTests.

hoanhtien:
The following code is used in many profiling test cases:
```C#
var session = LaunchProject(app, profiling, project, 
              TestData.GetPath("TestData\\NodejsTypeScriptProfileTest"), false);
while (profiling.IsProfiling) {
    System.Threading.Thread.Sleep(500);
}
var report = session.GetReport(1);
```
The while loop (that keeps checking `profiling.IsProfiling`) is not a correct way to block the execution until the profiling process completes. The reason is because profiling.IsProfiling is set in the `async ProfileProject` method which runs in an asynchronous context, so `profiling.IsProfiling = false` can also mean that profiling process has not been spawned yet:
https://github.com/Microsoft/nodejstools/blob/8169f1e6fe7c83e3f4061631a44de79b3e80ec73/Nodejs/Product/Profiling/NodejsProfilingPackage.cs#L275, 

To fix this issue, instead of waiting for `profiling.IsProfiling = false`, we wait for the profiling report to be generated.


mjbvz:
Resubmitted with #846

PR #555, opened 2015-10-20
Fix some failed test cases

hoanhtien:
BasicIntellisense/CtrlSpace
BasicProjectTests/AddFolderExists
BasicProjectTests/TestAddExistingFolderProject
BasicProjectTests/CopyFullPath

### DIFF
--- a/Nodejs/Tests/Core.UI/BasicIntellisense.cs
+++ b/Nodejs/Tests/Core.UI/BasicIntellisense.cs
@@ -41,13 +41,15 @@ namespace Microsoft.Nodejs.Tests.UI {
             );
 
             using (var solution = project.Generate().ToVs()) {
-                var server = solution.OpenItem("CtrlSpace", "server.js");
+                var optionsPage = NodejsTools.NodejsPackage.Instance.IntellisenseOptionsPage;
+                using (new NodejsOptionHolder(optionsPage, "OnlyTabOrEnterToCommit", false)) {
+                    var server = solution.OpenItem("CtrlSpace", "server.js");
 
-                server.MoveCaret(2, 13);
-                solution.ExecuteCommand("Edit.CompleteWord");
-                Keyboard.Type("\t");
+                    server.MoveCaret(2, 13);
+                    solution.ExecuteCommand("Edit.CompleteWord");
 
-                server.WaitForText("var http = require('http');\r\nhttp.createServer");
+                    server.WaitForText("var http = require('http');\r\nhttp.createServer");
+                }
             }
         }
 
--- a/Nodejs/Tests/Core.UI/BasicIntellisense.cs
+++ b/Nodejs/Tests/Core.UI/BasicIntellisense.cs
@@ -45,6 +45,7 @@ namespace Microsoft.Nodejs.Tests.UI {
 
                 server.MoveCaret(2, 13);
                 solution.ExecuteCommand("Edit.CompleteWord");
+                Keyboard.Type("\t");
 
                 server.WaitForText("var http = require('http');\r\nhttp.createServer");
             }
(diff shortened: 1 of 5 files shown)


Nodejs/Tests/Core.UI/BasicProjectTests.cs
```diff
@@ -993,6 +993,20 @@ public class BasicProjectTests : SharedProjectTest {
...
+        /// </summary>
+        private static string removeLabel(string itemName) {
+            string[] labels = new string[3] { " (browser)", " (node)", " (node, browser)" };
+            foreach (var label in labels) {
+                if (itemName.EndsWith(label)) {
+                    itemName = itemName.Substring(0, itemName.Length - label.Length);
+                    break; // safe to break as there is at most one label per name
+                }
+            }
+            return itemName;
```
mousetraps: Instead of messing around with the string (which is going to break if we ever change the string or the option is turned off), it's better to just turn off the option before running the test. e.g. 
https://github.com/Microsoft/nodejstools/blob/8169f1e6fe7c83e3f4061631a44de79b3e80ec73/Nodejs/Tests/Core.UI/NodejsBasicProjectTests.cs#L272

Nodejs/Tests/Core.UI/BasicProjectTests.cs
```diff
@@ -722,7 +722,7 @@ public class BasicProjectTests : SharedProjectTest {
                     Keyboard.Type(System.Windows.Input.Key.Enter);
 
 #if DEV11_OR_LATER
-                    VisualStudioApp.CheckMessageBox(MessageBoxButton.Ok, "Directory names cannot contain any of the following characters");
+                    VisualStudioApp.CheckMessageBox(MessageBoxButton.Ok, "Please enter a valid name.");
```
mousetraps: Have you tested this across VS versions?

Nodejs/Tests/Core.UI/BasicIntellisense.cs
```diff
@@ -45,6 +45,7 @@ public sealed class BasicIntellisense : NodejsProjectTest {
 
                 server.MoveCaret(2, 13);
                 solution.ExecuteCommand("Edit.CompleteWord");
+                Keyboard.Type("\t");
```
mousetraps: why is this necessary? Edit.CompleteWord should be sufficient, and we should try to fix the ExecuteCommand call if that's not working.

hoanhtien: Edit.CompleteWord is not doing what we think it is. Looking at [IntellisenseControler.cs#L646](https://github.com/Microsoft/nodejstools/blob/5c994768c5678579b76d71ba6ff48a30fa5991cd/Nodejs/Product/Nodejs/Intellisense/IntellisenseController.cs#L646), it does the same thing with ShowMemberList.

Nodejs/Tests/Core.UI/BasicProjectTests.cs
hoanhtien: After a second look, I will remove this change as it is not related to the three test cases I am fixing in this pull request (the only test case that is affected by this change is BasicProjectTests\LoadProjectWithDuplicateItems and it still fails after this change).

Nodejs/Tests/Core.UI/BasicProjectTests.cs
hoanhtien: Looks like it does not work on VS2013 :(

hoanhtien:
Updated the pull request to address feedback.

mjbvz:
Resubmitted and updated PR as #827

PR #574, opened 2015-10-30
Fix return value of GetTargetPath target

mthalman:
The implementation of Microsoft.NodejsTools.targets has a bug in it that prevents the Node.js project type from being referenced properly by CPS-based projects (Common Project System in VS).  Symptoms include the following:
* An additional project reference shows up programmatically when enumerating the project references of the CPS-based project.  This "project reference" doesn't actually have a source project and just references the Microsoft.NodejsTools.WebRole.dll file in the Node.js project's bin folder.
* The project reference shows up as unresolved in Solution Explorer.

The root of the issue is in the implementation of the GetTargetPath target.  The items returned by that target end up having metadata for OriginalItemSpec.  Having that item metadata set is what causes all the issues.  The fix is to return the same set of files but without the OriginalItemSpec metadata.

### DIFF
--- a/Nodejs/Product/Nodejs/Microsoft.NodejsTools.targets
+++ b/Nodejs/Product/Nodejs/Microsoft.NodejsTools.targets
@@ -137,14 +137,23 @@
   <Target
     Name="GetTargetPath"
     DependsOnTargets="CoreCompile"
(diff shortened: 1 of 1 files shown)


paulvanbrenk:
:+1:

PR #585, opened 2015-11-10
Add basic declaration for Mocha test framework.

### LINKED ISSUE #418: Errors in the TypeScript Mocha unit test template file
In both VS 2015 and VS 2013, adding a TypeScript Mocha unit test file will introduce errors to the project. 
Screenshot:

paulvanbrenk:
This fixed #418


billti:
Please use proper sentences in the comments (initial caps and periods), and it might be worth adding the async signatures (which takes the callback when done). Else :+1:

mousetraps:
hrm.. I'd rather just point people to tsd here than polluting every single TS unit test people create with more stuff they need to delete. Maybe create an fwlink that points here?
https://github.com/Microsoft/nodejstools/wiki/TypeScript#adding-typescript-type-definitions

billti:
But just adding a fwlink mean then still that doing: File / New Project / Build gives you errors. Ideally the templates should be self contained and usable. Yes, they need to delete to lines - not ideal, but at least the project builds without error after creation.

mousetraps:
closing this PR in favor of #694

PR #586, opened 2015-11-12
#566 No IntelliSense after npm install with a new project

### LINKED ISSUE #566: No intellisense on Express after npm install with a new project
**Repro steps:**
* file > new project > express 4 template
* wait for packages to install and analysis to complete
* hovering over `app` yields `unknown type`

When this issue occurs, of the top-level packages installed (the ones that appear in `package.json` and were installed with ), some appear to have been analyzed and some don't:

> :x: **body-parser:** unknown type
:x: **cookie-parser:** unknown type
:x: **debug:** unknown type
:x: **express:** unknown type: 
:heavy_check_mark: **jade:** completions available
:heavy_check_mark: **morgan:** completions available
:heavy_check_mark: **serve-favicon:** completions available
:heavy_check_mark: **stylus:** completions available

An interesting coincidence about this list is that the packages yielding `unknown type` are those packages sorted towards the beginning of the alphabet. This is consistently the case, and this issue appears to be unrelated to the sort order in package.json.

The express 3 template yields similar results:
> :x: **express:** unknown type
:x: **jade:** unknown type
:heavy_check_mark: **stylus:** completions available

**Workaround:** if you install any other package, say `azure`, completions are suddenly available for all of the packages (including those towards the beginning of the alphabet.)

mousetraps:
This was due to a race condition with the delayed analysis queue. In
particular, there were some cases where the NodejsFileNode was created
(and therefore an analysis unit enqueued) after the delayed analysis queue
had already finished processing existing entries. This fix simplifies the
logic to make it more deterministic. In particular, get rid of the
separate filewatcher altogether, add entries to the delayed analysis queue
whenever new NodejsFileNodes are created, and restart the idle
node_modules timer after enqueuing the unit.

fix #566

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
@@ -52,8 +52,6 @@ namespace Microsoft.NodejsTools.Project {
 
         // We delay analysis until things calm down in the node_modules folder.
         internal Queue<NodejsFileNode> DelayedAnalysisQueue = new Queue<NodejsFileNode>();
-        private FileWatcher _nodeModulesWatcher;
-        private readonly string[] _watcherExtensions = { ".js", ".json" };
         private object _idleNodeModulesLock = new object();
         private volatile bool _isIdleNodeModules = false;
         private Timer _idleNodeModulesTimer;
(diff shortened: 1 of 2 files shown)


Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -694,7 +657,7 @@ public NodejsProjectNode(NodejsProjectPackage package)
             if (null == ModulesNode) {
                 ModulesNode = new NodeModulesNode(this);
                 AddChild(ModulesNode);
-                CreateIdleNodeModulesWatcher();
+                _idleNodeModulesTimer = new Timer(OnIdleNodeModules);
```
paulvanbrenk: I wouldn't create a new timer each time, but reset/start the existing timer... they are not cheap.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -52,8 +52,6 @@ class NodejsProjectNode : CommonProjectNode, VsWebSite.VSWebSite, INodePackageMo
 
         // We delay analysis until things calm down in the node_modules folder.
         internal Queue<NodejsFileNode> DelayedAnalysisQueue = new Queue<NodejsFileNode>();
-        private FileWatcher _nodeModulesWatcher;
```
paulvanbrenk: :+1:

paulvanbrenk:
LGTM, the timer creation can be pushed to a future release.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
mousetraps: It's not getting created each time - this only happens when ModulesNode is null, which is when the project hasn't been loaded yet. The timer is reset here: https://github.com/mousetraps/nodejstools/blob/i566/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs#L113

paulvanbrenk: Cool, the things you miss when only looking at the diff...

billti:
If this has been validated, then :+1:. Thanks!

Issue #584, opened 2015-11-10
VS 2015 stop working if changes any code while debugging(1.1 rc3)


alanpurple:
VS2015 stop working if I change something and save

mousetraps:
Cant repro this. Can you please provide more info? What do you mean by "stops working"? Can you provide simple repro steps starting from one of the project templates? Or do you have a project you can send us where you can reproduce the issue consistently?

alanpurple:
Crash I mean
On Nov 12, 2015 7:02 AM, "Sara Itani" <notifications@github.com> wrote:
>
> Cant repro this. Can you please provide more info? What do you mean by
"stops working"? Can you provide simple repro steps starting from one of
the project templates? Or do you have a project you can send us where you
can reproduce the issue consistently?

mousetraps:
Thanks, but that's *really* not enough information for us to do anything about it... 
* since it was a crash, there should be a corresponding entry in the Windows event log, and grabbing the stack trace from that is necessary for us to track this down if we can't reproduce the issue. 
* what are steps to reproduce the issue?
* can you reproduce the issue consistently?
* on which projects types does it occur? does it occur in a basic console app?

alanpurple:
@mousetraps 
just for any nodejs project,
press F5 for start debugging
and then modify any file which is included in the process, and then ctrl+s and immediately you can see the VS2015 stops working message

If you still can't reproduce this, I will provide a log, tell me, anyway now I'm outside so I can work later

aruneshchandra:
@alanpurple please provide the log if you can. Also, can you repro this consistently ?

Is anybody else having this issue ? it will be helpful for us to understand how widespread is this issue.

kant2002:
I have similar issue before 1.1rc3. It was sporadic, but I lately avoid debugging in NTVS since it is start using much more memory, and I have to restart VS

billti:
Looks potentially the same as #589 .

zhengbli:
@kant2002 @alanpurple Are you guys on node v5? It runs fine for me on VS 2013 and VS 2015 with node v4.2, though I can indeed repro the crash on node v5, so I wonder if it is related.

alanpurple:
@zhengbli  yes, node 5.0

@mousetraps  I think this is same as #589 as billti said, still need a log?

kant2002:
@zhengbli I was on 0.12, but I have something like that just couple times. Not 100% sure that this was exactly this issue. Could try to reproduce

[...]

PR #594, opened 2015-11-15
#584 VS2015 crashes if change code while debugging w/ Node.js v5

mousetraps:
- v8 has recently removed the "stack_update_needs_step_in" property in
  responses. It looks like it always returned the same value as
  "stack_modified" anyways, so go ahead and ignore the property altogether.
  https://codereview.chromium.org/1247363002/diff/1/src/liveedit-debugger.js

### DIFF
--- a/Nodejs/Product/Nodejs/Debugger/NodeDebugger.cs
+++ b/Nodejs/Product/Nodejs/Debugger/NodeDebugger.cs
@@ -465,14 +465,10 @@ namespace Microsoft.NodejsTools.Debugger {
                 return false;
             }
 
-            // Make step into if required
-            if (changeLiveCommand.NeedStepIn) {
+            // Make step into and update stacktrace if required
+            if (changeLiveCommand.StackModified) {
                 var continueCommand = new ContinueCommand(CommandId, SteppingKind.Into);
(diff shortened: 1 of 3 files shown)


Nodejs/Product/Nodejs/Debugger/NodeDebugger.cs
```diff
@@ -465,14 +465,10 @@ public NodeDebugger(Uri debuggerEndpointUri, int id)
                 return false;
             }
 
-            // Make step into if required
-            if (changeLiveCommand.NeedStepIn) {
+            // Make step into and update stacktrace if required
+            if (changeLiveCommand.StackModified) {
```
billti: I know you stated in the comments...

> It looks like it always returned the same value as "stack_modified" anyways

... but it still makes me nervous just changing the property in the condition but falling through to the same code. Has this change been verified with older version of Node?

mousetraps:
Yes, I verified it with 0.10.20 and 4.1.2 too and have yet to run into a case when the values aren't the same - I can try a few other configs as well. 0.10.33 and 0.12.7 would be good to try too

billti:
Yeah, those would be good, as some tools install (or point to) those specific versions. If they work, then :+1:   Thanks :smile:

PR #599, opened 2015-11-17
#598 remove docker publish hooks

### LINKED ISSUE #598: remove docker publish hooks
We are no longer planning to include docker integration from the Publish dialog, so we should remove these hooks.

mousetraps:
fix #598

### DIFF
--- a/Nodejs/Product/Nodejs/Nodejs.csproj
+++ b/Nodejs/Product/Nodejs/Nodejs.csproj
@@ -901,10 +901,6 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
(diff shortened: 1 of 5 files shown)


billti:
I'd rather this wasn't merged in until it had been reviewed properly or there was some details on the test coverage done. Setup and templates are both touched with this, and they are notoriously error prone.

mousetraps:
Yeeaahhh... I'll leave a comment next time I do something scary looking like that... 

In my perspective... it was 2am, no one had reviewed it yet, it's a push-button unmerge, and I wanted to kick off a 1.1 build with these changes so... I suppose I could have pointed the build to a private branch but that seemed equally risky in my mind... 

Fwiw this doesn't actually touch templates. It touches a file we used to copy so that we could read the dockerfile from outside of the zipped item templates. 

Coverage included... Testing in experimental hive, Building the installer, installing the product, creating various item templates, verifying the docker button no longer shows up, and verifying the install dirs all looked fine.

PR #644, opened 2015-12-12
Fixed tests broken by the changed version of Newtonsoft.Json

kant2002:
This is addressed feedback in #539


mousetraps:
As mentioned in #645, the proper fix will be to upgrade the Newtonsoft.Json version and remove these tests altogether.

PR #647, opened 2015-12-15
#604 Runtime syntax error kills Visual Studio

### LINKED ISSUE #604: Runtime syntax error kills Visual Studio
Windows 10 10586 + VS2015 + Node.js v5.1.0 + NTVS 1.1

Repro:

1. Make new project
2. Type `let foo = 3`
3. Run
4. Node.js throws a syntax error saying `Block-scoped declarations (let, const, function, class) not yet supported outside strict mode`
5. VS becomes unresponsive

mousetraps:
- we should be exiting the debug-connect loop when the connection is closed.

fix #604


billti:
:+1:

PR #653, opened 2016-01-04
Open Command Prompt Here --> Developer Cmd Prompt

SlurpTheo:
If I'm inside Visual Studio and I ask it to open a command prompt it should open and properly configure a *Developer Command Prompt* appropriate for the version of Visual Studio being used.

### DIFF
--- a/Common/Product/SharedProject/CommonFolderNode.cs
+++ b/Common/Product/SharedProject/CommonFolderNode.cs
@@ -109,6 +109,16 @@ namespace Microsoft.VisualStudioTools.Project {
                                 "cmd.exe"
                             )
                         );
+
+                        string visualStudioVersion = Environment.GetEnvironmentVariable("VisualStudioVersion");
+                        string commonToolsDir = Environment.GetEnvironmentVariable(string.Format("VS{0}COMNTOOLS", visualStudioVersion.Replace(".", "")));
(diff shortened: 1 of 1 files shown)


mousetraps:
Thanks for the PR! Just so we understand your scenario better... any reason you prefer the developer command prompt with Node.js projects?

SlurpTheo:
@mousetraps I think the last `PATH`, environment variable setup I missed when attempting **Open Command Prompt** from Visual Studio was missing having `tsc.exe` available at the launched-in-the-right-location command prompt.

mousetraps:
Gotcha, thanks. I don't think we're going to pull this one in - I just took a look at VsDevCmd.bat, and a few observations:
* it adds a lot of unnecessary weight/complexity for most scenarios
* it doesn't get updated if new versions of TS are installed, which means that tsc won't resolve to the proper location anyways
* if you've installed typescript globally, it will override %PATH% with the version installed with Visual Studio, which creates potentially confusing inconsistencies between the default command prompt behavior and VS behavior.

That said, I do recommend taking a look at the Open Command Line extension, which should enable you to customize the experience a little more than currently possible with NTVS. 
https://visualstudiogallery.msdn.microsoft.com/4e84e2cf-2d6b-472a-b1e2-b84932511379

Issue #646, opened 2015-12-14
Fix failing unit tests


mousetraps:
Right now it's difficult to detect regressions.
https://ci.appveyor.com/project/mousetraps/nodejstools/branch/master

mjbvz:
"Fixed" by #731, in that we just disabled a bunch of tests that were failing in Appveyor.

I'm going to close this for now, but I suggest we open more scoped issues to track enabling specific sets of these tests again.

PR #669, opened 2016-01-25
#666 Disable failing tests to make it easier to detect regressions

### LINKED ISSUE #666: Disable failing tests
It's unfortunate that a subset of our tests are broken, but it's going to be pretty tough to clean them all up soon. We should disable them so that the AppVeyor run passes and its easier to detect regressions.

mousetraps:
Used the following (imperfect, but good enough) script to disable tests based on test run results: https://github.com/mousetraps/disable-failing-mstests

I suspect there may still be some lingering issues, so I plan to iterate after the test run completes. Additionally I plan to update #646 with priorities re: re-enabling tests. For instance, many of our debugger tests are failing due to an external issue in Node.js that is preventing us from detecting the end of program execution.

close #666

### DIFF
--- a/Common/Tests/SharedProjectTests/BasicProjectTests.cs
+++ b/Common/Tests/SharedProjectTests/BasicProjectTests.cs
@@ -1186,6 +1186,7 @@ namespace Microsoft.VisualStudioTools.SharedProjectTests {
             }
         }
 #endif
+        [Ignore]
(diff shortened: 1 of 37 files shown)


mousetraps:
I disabled all the tests that were broken locally, and now I'm working through some of the CI-specific issues. Baby steps :smiley:

PR #704, opened 2016-02-25
Check for null before dereference on RootPackage.HomePages

mjbvz:
**Defect**
`RootPackage.Homepage` can return null to indicate that no root package
exists. This causes two potential null dereferences in `DependencyNode.cs`

**Fix**
This fix just adds a null check before accessing the enumerator on the
`HomePages` property.


mousetraps:
:+1: - looks fine, though afaik Homepages is never null. Out of curiosity, was specific time when this was not the case, or did you just fix up as you came across it?

mjbvz:
Debugging new project creation, I was hitting null dereferences in these two locations. I don't know specifically what step of project it was in at the time, but the `RootPackage.Package` property was null, so `RootPackage.Homepages` was returning null.

mousetraps:
Oh, interesting... good catch!

PR #706, opened 2016-02-25
Lowercase and trim npm package names for templates

### LINKED ISSUE #705: Project Templates Generate Invalid package.json name
When creating a new project using a template, the current Vs Project name to `package.json` package name mapping in `NodejsPackageParametersExtension.cs` does not always generate valid package names: https://docs.npmjs.com/files/package.json#name

The main issue is that package names should be all lowercase, but the current transform does not change the case. Other limits, such as the size, are less likely to be hit normally.

For the name casing, we can either just lowercase the name, or split camel cased names and insert hyphens between the sections. The latter is probably a better experience.

mjbvz:
Adds a little extra normalization when converting a VS project names to npm project names when generating the `package.json` for the templates:
* Lowercase names. Inserts hyphens between camelcased strings: "ExpressApp4" -> "express-app4"
* Trim name length to 214 characters.

closes #705


Nodejs/Product/ProjectWizard/NodejsPackageParametersExtension.cs
```diff
@@ -14,23 +14,36 @@
...
+
 namespace Microsoft.NodejsTools.ProjectWizard {
     class NodejsPackageParametersExtension : IWizard {
-        public void RunStarted(object automationObject, Dictionary<string, string> replacementsDictionary, WizardRunKind runKind, object[] customParams) {
-            var projectName = replacementsDictionary["$projectname$"];
+        private const int NpmPackageNameMaxLength = 214;
 
-            // Remove all leading url-invalid, underscore, and period characters from the string
+        /// <summary>
+        /// Normalize a project name to be a valid Npm package name.
```
mousetraps: Could you include a link to the docs on valid package names in the comment here?
https://docs.npmjs.com/files/package.json#name

Nodejs/Product/ProjectWizard/NodejsPackageParametersExtension.cs
```diff
@@ -14,23 +14,36 @@
...
+        /// <param name="projectName">Name of a VS project.</param>
+        private string GetNpmSafeProjectName(string projectName) {
+            // Remove all leading url-invalid, underscore, and period characters
             var npmProjectNameTransform = Regex.Replace(projectName, "^[^a-zA-Z0-9-~]*", string.Empty);
 
             // Replace all invalid characters with a dash
-            npmProjectNameTransform = Regex.Replace(npmProjectNameTransform, "[^a-zA-Z0-9-_~.]", "-");
+            npmProjectNameTransform = Regex.Replace(npmProjectNameTransform, "[^a-zA-Z0-9-_~.]", "-");
+
+            // insert hypens between camelcased sections.
```
mousetraps: nit: spelling

Nodejs/Product/ProjectWizard/NodejsPackageParametersExtension.cs
```diff
@@ -14,23 +14,36 @@
...
+            npmProjectNameTransform = Regex.Replace(npmProjectNameTransform, "([a-z0-9])([A-Z])", "$1-$2").ToLowerInvariant();
+
+            return npmProjectNameTransform.Substring(0, Math.Min(npmProjectNameTransform.Length, NpmPackageNameMaxLength));
+        }
```
mousetraps: nit: move below public methods so that it's easier to understand the interfaces.

mousetraps:
minor comments, and :+1: after that!

PR #708, opened 2016-02-25
#145 Remove Npm logging to status bar

### LINKED ISSUE #145: npm install should not print output to status bar
this can pollute the more useful analysis queue updates

mjbvz:
**Defect**
Npm logging to status bar hides more important messages, such as the analysis queue updates.

**Fix**
Delete functions that were logging npm output to the status bar.

closes #145


Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
```diff
@@ -339,19 +306,6 @@ private class NpmPathProvider : INpmPathProvider {
...
-                    e.Cancelled ? SR.NpmCancelledWithErrors : SR.NpmCompletedWithErrors,
-                    e.CommandText
-                );
-            } else if (e.Cancelled) {
-                message = SR.GetString(SR.NpmCancelled, e.CommandText);
-            } else {
-                message = SR.GetString(SR.NpmSuccessfullyCompleted, e.CommandText);
-            }
-
-            ForceUpdateStatusBarWithNpmActivitySafe(message);
```
mousetraps: Let's keep this logic in here. The primary issue was that we were flooding the status bar with activity *during* the install that wasn't particularly useful. It's useful, however, to report back the final result of the npm install to the status bar.

mjbvz: Ok, sounds good. I misunderstood the fix the bug was asking for.

Will send out an updated pull request shortly.

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
```diff
@@ -359,6 +346,18 @@ private class NpmPathProvider : INpmPathProvider {
...
+        private static string GetStatusBarMessage(NpmCommandCompletedEventArgs e) {
+            if (e.WithErrors) {
+                return SR.GetString(
+                    e.Cancelled ? SR.NpmCancelledWithErrors : SR.NpmCompletedWithErrors,
+                    e.CommandText);
+            }
+            if (e.Cancelled) {
+                return SR.GetString(SR.NpmCancelled, e.CommandText);
+            }
+            return SR.GetString(SR.NpmSuccessfullyCompleted, e.CommandText);
```
mousetraps: nit: even though it's fine to place the return within the if blocks like this, the code is more readable when you use if -> else if because it's clearer at a glance that the if blocks are mutually exclusive.

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
```diff
@@ -217,7 +217,7 @@ private class NpmPathProvider : INpmPathProvider {
         }
 
 #if INTEGRATE_WITH_ERROR_LIST
-
+
```
mousetraps: nit: trim whitespace

mousetraps:
minor comments, also I merged in the line endings change, which messed this one up.

:+1: after that!

mousetraps:
Oops, looks like you introduced an extra file change in this iteration - NodejsPackageParametersExtension.cs

mjbvz:
Sorry, that was a whitespace only change and the merge got messed up on that file. Should be fixed now.

PR #709, opened 2016-02-26
Normalize line endings to CR LF in *.cs files

mjbvz:
All *.cs files should be using windows style line endings (CR+LF) for source code.

The change updates the .gitattributes file to reflect this and converts*.cs files. Having a standard for the *.cs files should make contributions and reviews easier.


mousetraps:
awesome! :+1:

PR #710, opened 2016-02-26
Remove *.ts and *.js extension check code duplication

mjbvz:
**Defect**
There are currently a number of places that implement logic for checking if a file is Javascript or Typescript based on file extension. This code is hard to read and can get out of sync.

**Fix**
Created two helper methods to implement these checks. Replace most of the old extension check logic with calls to these helpers instead.

### DIFF
--- a/Nodejs/Product/Nodejs/NodejsFileExtensions.cs
+++ b/Nodejs/Product/Nodejs/NodejsFileExtensions.cs
@@ -1,0 +1,31 @@
+﻿//*********************************************************//
+//    Copyright (c) Microsoft. All rights reserved.
+//    
+//    Apache 2.0 License
+//    
+//    You may obtain a copy of the License at
+//    http://www.apache.org/licenses/LICENSE-2.0
+//    
+//    Unless required by applicable law or agreed to in writing, software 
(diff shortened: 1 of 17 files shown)


Nodejs/Product/Nodejs/NodejsFileExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+//    implied. See the License for the specific language governing 
+//    permissions and limitations under the License.
+//
+//*********************************************************//
+
+using System;
+using System.IO;
+
+namespace Microsoft.NodejsTools {
+    static class NodejsFileExtensions {
```
mousetraps: Can you change the name of this to be something like NodejsFileTypeHelpers? Files named with "Extensions" tend to be extension methods, so it's a bit overloaded.

mousetraps:
There are also several more locations that are missed with this change that we may as well replace.

:+1: after that!

mousetraps:
LGTM, but looks like there are conflicts now (might have been due to #730) - can you update and merge this in?

mjbvz:
Closing since the line ending change makes merging back in to messy. Will reapply the change and submit another PR.

PR #716, opened 2016-02-29
Added simple script for running all tests in NTVS

mjbvz:
Adds a simple powershell script that can be used to run all tests that are part of NTVS.

I'm also working on documenting how contributors can run our tests on the wiki: https://github.com/Microsoft/nodejstools/wiki/Running-Our-Tests

### DIFF
--- a/Tools/runtests.ps1
+++ b/Tools/runtests.ps1
@@ -1,24 +1,26 @@
 # Run the basic sets of tests using powershell
-$vsversions = @("12.0", "14.0", "15.0")
+$vsversions = @("15.0", "14.0", "12.0")
 
 # Set of tests to run
 # AzurePublishing.Tests.UI are currently skipped.
 $tests = @("JSAnalysisTests", "Nodejs.Tests.UI", "NodeTests","NpmTests", "ProfilerTests", "SharedProjectTests")
 
 # Try to get the path to the vstest.console test runner
+$selectedVsVersion = ''
 $vstest =  ''
 foreach($vsversion in $vsversions) {
 	$path = "C:\Program Files (x86)\Microsoft Visual Studio $vsversion\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe"
-	if (Test-Path $path) {
+	if (($selectedVsVersion -eq '') -and (Test-Path $path)) {
+		$selectedVsVersion = $vsversion
 		$vstest = $path
 	}
 }
 
-if ($vstest -eq '') {
+if ($selectedVsVersion -eq '') {
 	echo "Could not find valid vstest.console"
 	return
 }
 
-$testFiles = $tests | % { ".\BuildOutput\Release14.0\Tests\$_.dll" }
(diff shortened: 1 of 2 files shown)


kant2002:
Great thanks for your effort. It really helps!

runtests.ps1
```diff
@@ -0,0 +1,7 @@
+# Run the basic sets of tests using powershell
+$vstest = "C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe"
```
mousetraps: This script will not work for multiple versions of Visual Studio.

runtests.ps1
```diff
@@ -0,0 +1,7 @@
+# Run the basic sets of tests using powershell
+$vstest = "C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe"
+
+$tests = @("NpmTests", "NodeTests", "ProfilerTests", "JSAnalysisTests", "MockVsTests", "Nodejs.Tests.UI")
```
mousetraps: We should also run SharedProjectTests and add a comment that AzurePublishing.Tests.UI are being skipped.

runtests.ps1
```diff
@@ -0,0 +1,7 @@
+# Run the basic sets of tests using powershell
```
mousetraps: This file shouldn't be top-level - can you move this into a directory within Tools?

mjbvz:
Ok, updated based on your comments.

Tools/runtests.ps1
```diff
@@ -0,0 +1,24 @@
...
+$tests = @("JSAnalysisTests", "Nodejs.Tests.UI", "NodeTests","NpmTests", "ProfilerTests", "SharedProjectTests")
+
+# Try to get the path to the vstest.console test runner
+$vstest =  ''
+foreach($vsversion in $vsversions) {
+	$path = "C:\Program Files (x86)\Microsoft Visual Studio $vsversion\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe"
+	if (Test-Path $path) {
+		$vstest = $path
+	}
+}
```
mousetraps: what if there are multiple VS versions installed?

Tools/runtests.ps1
```diff
@@ -0,0 +1,24 @@
...
+
+if ($vstest -eq '') {
+	echo "Could not find valid vstest.console"
+	return
+}
+
+$testFiles = $tests | % { ".\BuildOutput\Release14.0\Tests\$_.dll" }
```
mousetraps: this is still targeting 14.0

Tools/runtests.ps1
mjbvz: Thrice the testing!

mousetraps:
:+1: looks good - we can continue to iterate on the script as we need more features (accepting different args, etc.)

Issue #689, opened 2016-02-19
AppVeyor CI -> 100% passing


mousetraps:
We should investigate the remaining issues, and see if we can figure out what's up. If we cannot within any reasonable amount of time, we should file an issue, and skip these tests on our CI server for now.

PR #717, opened 2016-02-29
Temp disable all tests that are failing in AppVeyor

mjbvz:
We need to get CI back into a good state so we can have some confidence when merging in pull requests. This change just disables all tests that are failing in appveyor. Related to #689

The major source of problems is the REPL tests, which seem to be failing nondeterministicly. I think there is because of `SocketLock` and threading. I just disabled them all for now but will revise this later.

### DIFF
--- a/Nodejs/Tests/Core/ReplWindowTests.cs
+++ b/Nodejs/Tests/Core/ReplWindowTests.cs
@@ -493,6 +493,7 @@ undefined";
             }
         }
 
+        [Ignore]
         // https://nodejstools.codeplex.com/workitem/1575
         [TestMethod, Priority(0), Timeout(180000)]
         public async Task TestNpmReplCommandProcessExitSucceeds() {
--- a/Nodejs/Tests/Core/ReplWindowTests.cs
+++ b/Nodejs/Tests/Core/ReplWindowTests.cs
@@ -43,8 +43,7 @@ namespace NodejsTests {
             }
         }
 
-        [Ignore]
-        [TestMethod, Priority(0)]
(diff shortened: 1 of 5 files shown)


mjbvz:
I've opened #718 with some ideas of how to deal with the skipped unit tests.

Nodejs/Tests/Core/ReplWindowTests.cs
```diff
@@ -428,6 +446,7 @@ public class ReplWindowTests {
         }
 
         [Ignore]
+        [Ignore]
```
mousetraps: Hrm, this is already being ignored.

Nodejs/Tests/Core/ReplWindowTests.cs
```diff
@@ -517,6 +538,7 @@ public class ReplWindowTests {
...
```
mousetraps: already being ignored

mousetraps:
For the tests that are *only* failing in AppVeyor, and not locally with the same configuration, let's instead use the vstest TestCaseFilter option to ignore them because we might be able to fix some of these without modifying the code itself.

mousetraps:
Also, have you changed anything about the AppVeyor run lately? It looks like it's failing after 2 min, and it should very much pass for this PR :smiley:

mjbvz:
Ok, that sounds like a good plan. Will send out the updated review shortly.

I have not intentionally changed anything in AppVeyor but it did seem to start failing yesterday. Let's see what happens when I update this PR.

mjbvz:
I've marked the unreliable/failing appveyor only tests with the category `AppVeyorIgnore`. I also ignored a few other tests that really were failing locally.

Issue #694, opened 2016-02-20
Automatic typings acquisition 


mousetraps:

mousetraps:
added in #747

mousetraps:
closed in #747

PR #719, opened 2016-03-01
First Stage of Auto Typing Download Work - Detecting new packages

mjbvz:
First step of implementing automatic type downloading -  #694 . This pull request just adds an internal interface for detecting when a new npm package is installed. We will hook into this interface to automatically download the typings for new packages.

**New Interface**
The new interface is the `PackagesAdded` event on `NodeModulesNode`. This event is invoked whenever `NodeModulesNode` detects that a new package has been added. 

Adding this logic at this level is a little strange since `NodeModulesNode` is conceptually a user interface component, but it was the best place I found that had all the module data available and would also know which modules have already been loaded.

**Other Changes**
The majority of the other changes are simple method extractions from reshaper. I'm still trying to find the best way to hook this work up for testing, and I explored a few different options up and down the object hierarchy. These extractions were made to make testing easier or clarify the existing implementations of various methods.

I still don't have a good story for testing at the level of `NodeModulesNode` yet though. It's too reliant on UI concepts and too tightly coupled to everything else at the moment.

### DIFF
--- a/Nodejs/Product/Nodejs/Project/AbstractNpmNode.cs
+++ b/Nodejs/Product/Nodejs/Project/AbstractNpmNode.cs
@@ -75,6 +75,51 @@ namespace Microsoft.NodejsTools.Project {
             //  it'll all be in the collapsed state, which will be annoying for users who
             //  have drilled down into the hierarchy
             var recycle = new Dictionary<string, DependencyNode>();
+            var remove = GetNodesToRemoveOrRecycle(parent, modules, recycle);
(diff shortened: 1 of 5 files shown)


mousetraps:
Haven't gotten a chance to take a look at this, but can you instead make this PR on the auto-dts feature branch (just created it)?
https://github.com/Microsoft/nodejstools/tree/auto-dts

mjbvz:
Moved to #721 to target the new branch.

PR #722, opened 2016-03-02
Disable Debug Assertion for Parsing NPM Person objects

mjbvz:
**Defect**
While debugging, Npm `Person` ends up throwing thousands and thousands of exceptions because the Json data it is receiving from NPM is invalid according to its current check. It expects an object, but a person in a package.json may also be a string: https://docs.npmjs.com/files/package.json#people-fields-author-contributors

**Fix**
* Moved the parsing logic for `Person` to a factory named `CreateFromJsonSource`. This should make it clear to callers what this function does.
* For parsing, also accept just regular json strings as input. The new factory does not parse this field currently but uses it as the name field for now. Parsing may be added later.
* Added some unit tests to pin the existing behavior and test the new string behavior.

A whitespace only change to InteractiveWindow.cs is included in this PR. Git automatically created this change and I cannot undo it without manually editing the .gitattributes file. I think we just need to commit this file again so that git is no longer confused about the line endings in this file.


Nodejs/Product/Npm/SPI/Person.cs
```diff
@@ -29,68 +29,99 @@ internal class Person : IPerson {
...
             // and simply handling them causes performance issues.
-            var matches = RegexPerson.Matches(source);
+            var matches = ObjectPersonRegex.Matches(source);
             if (matches.Count >= 1) {
                 foreach (Match match in matches) {
                     var group = match.Groups["name"];
-                    if (group.Success) {
-                        Name = group.Value;
+                    if (@group.Success) {
+                        name = @group.Value;
```
mjbvz: Weird... I was wondering about those names but thought it was intentional. Reshaper's extract messed something up here. Will undo this.

Common/Tests/Utilities.UI/UI/InteractiveWindow.cs
```diff
@@ -1,310 +1,310 @@
-﻿/* ****************************************************************************
```
mousetraps: You should be able to remove this change now that #709 has been merged

Nodejs/Product/Npm/SPI/Person.cs
```diff
@@ -29,68 +29,99 @@ internal class Person : IPerson {
...
         // so we should match them separately.
-        private static readonly Regex RegexPerson = new Regex(
-        "\"name\":\\s*\"(?<name>[^<]+?)\"" +
-        "|" +
-        "\"email\":\\s*\"(?<email>[^<]+?)\"" +
-        "|" +
-        "\"url\":\\s*\"(?<url>[^<]+?)\"",
-        RegexOptions.Singleline);
-    
+        private static readonly Regex ObjectPersonRegex = new Regex(
```
mousetraps: nit: `PersonObjectRegex` and `StringPersonRegex` would be clearer

Nodejs/Product/Npm/SPI/Person.cs
```diff
@@ -29,68 +29,99 @@ internal class Person : IPerson {
...
+
+            var objectPerson = TryCreatePersonFromObject(source);
+            if (objectPerson != null)
+                return objectPerson;
+
+            var stringPerson = TryCreatePersonFromString(source);
+            if (stringPerson != null)
+                return stringPerson;
+
+            return new Person(source);
```
mousetraps: `return TryCreatePersonFromObject(source) ?? TryCreatePersonFromString(source) ?? new Person(source);`

Nodejs/Product/ProjectWizard/NodejsPackageParametersExtension.cs
```diff
@@ -19,7 +19,7 @@
 using System.Text.RegularExpressions;
```
mousetraps: remove this file from changeset too - again, I think #730 should make things easier :smiley:

Nodejs/Tests/NpmTests/PersonTests.cs
```diff
@@ -0,0 +1,96 @@
...
+//
+//*********************************************************//
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.NodejsTools.Npm.SPI;
+
+namespace NpmTests {
+    [TestClass]
+    public class PersonTests {
+        [TestMethod, TestCategory("UnitTest")]
```
mousetraps: TestCategory("npm-integration") if we want to start working it in?

Nodejs/Product/Npm/SPI/Person.cs
```diff
@@ -29,68 +29,99 @@ internal class Person : IPerson {
...
+        /// </summary>
+        /// <param name="source">Json source</param>
+        private static Person TryCreatePersonFromString(string source) {
+            var matches = StringPersonRegex.Matches(source);
+            if (matches.Count == 1) {
+                var match = matches[0];
+                var group = match.Groups["name"];
+                if (group.Success)
+                    return new Person(group.Value);
             }
```
mousetraps: If we are not parsing the string, then why are we sending it through a regex?

Nodejs/Product/Npm/SPI/Person.cs
mjbvz: Just removed `StringPersonRegex` since it was not really being used

mjbvz:
Ok, I removed the string person regex stuff for now, addressed the other comments, and have synced with master to get rid of any line ending changes.

mjbvz:
@mousetraps can you please take another quick look at this review. This currently makes debugging a pain if you ever accidentally trigger the npm download, plus the assertion is not even correct to being with.

mousetraps:
test categories still need to be updated but :+1: after that!

Issue #712, opened 2016-02-27
Mark Non Unit Tests


mjbvz:
Some of the existing tests are not unit tests. We need a way to distinguish these from the true unit tests in the project.

Rough guide for identifying non-unit tests in a first pass:
* Takes longer than 250ms to run (should be decreased later).
* Non-deterministic.
* Would fail in a sandbox without file system or network access.
* Depends on test execution order.
* UI automation.

Goal is to have a set of unit tests that can be run from inside visual studio directly, quickly and with complete confidence. This will allow detecting regressions more quickly.

At this stage, it may be easier to mark the true unit tests instead of marking the non-unit tests. 
Related to #646 .

mousetraps:
Perhaps a better demarcation would be UI Tests and non-UI tests (this would simply be based on the VSTestHost attribute, and also - for the most part - should satisfy the "runs quickly" property because we don't have to wait for UI.)

Re: "UnitTest" and "IntegrationTest" categories - I'm not sure how valuable this is given the constraints of our codebase. Many of the longer-running tests are the closest we can get to a 'unit test' due to some of the constraints of Node / VS, and I think that - in our codebase - just running things that would be classified as "unit tests" provides a false sense of security.

mjbvz:
You are correct that unit tests will not provide 100% end-to-end confidence, but that's not their job. The unit test should run quickly, continuously, and reliably to drive development and catch simple errors. Once all the unit tests pass, then you can kick off the 5-20min full test suite or a more targeted subset of integration tests to make sure no regressions have been introduced. There aren't very tests that test individual units of the code as it stands.

Marking UI tests also makes sense since they are slower, have more dependencies, and may be less reliable than the rest of our tests. Perhaps everything without a category should just be assumed to be a normal IntegrationTest then? But I still think there is value is marking the unit tests separately and using them as a tool for development and code maintenance.

mousetraps:
I think we both agree on what might make a good unit test in an *ideal* world. However, unfortunately we are not building an application in isolation - we are building on top of platforms that we have no control over, and oftentimes that's the most important functionality to test - at the whim of 20yo VS interfaces, of of Node.js, npm, etc.  So the definition proposed is *far* too strict for our taste. 

A unit test is, quite simply, something that tests a unit of functionality. 

And, of course, we don't expect unit tests to provide 100% end-to-end confidence...
![](https://media.giphy.com/media/3o85xF1mzqyOw7Ri1O/giphy.gif)

[...]

PR #724, opened 2016-03-02
Revert "Marking unit tests - Pass 1"

mousetraps:
Reverts Microsoft/nodejstools#713


mousetraps:
reverting this change until we land on an approach in #712

PR #728, opened 2016-03-02
Added *.temp_proj to gitignore

mjbvz:
Saw this showing up a few times during development. It always is auto deleted before being commited, but this file should not even be picked up in the first place.


mousetraps:
:+1:

PR #729, opened 2016-03-03
Use generated resource helper

mjbvz:
**Defect**
To get string resources, we currently, access everything through an SR class. This class hardcodes a bunch of strings keys to lookup resources in a resx. Hardcoding these strings is not great because these strings can get out of sync with the actual rest file .

**Fix**
This hooks up the resx file to generate an typed cs file automatically. This ensures that the resx file and the string resource uses cannot get out of sync.

The existing class is also kept around since a number of places are using the resources ids as attribute arguments currently.

Sorry about the `Common/Tests/Utilities.UI/UI/InteractiveWindow.cs` getting included in this change. It's just a line ending change that should be resolved once #727 goes in.


mousetraps:
Let's table this change for now. I understand it's a simple refactor, but it's is a rather large change to take and verify so close to a release, and isn't really solving any immediate problems.

PR #730, opened 2016-03-03
Update .gitattributes and normalize all line endings

mousetraps:
Related to issue described in #727 

@mjbvz can you pull this branch onto your machine first to verify this change resolves the issue?


mousetraps:
Diff without whitespace: https://github.com/Microsoft/nodejstools/pull/730/files?w=1

mjbvz:
Thanks! this should make future changes a lot easier (and also easier to review). Sorry for not taking a look too, I mixed up the name with the other changes that were similarly titled.

I believe the change may have introduced a regression with the tests though: https://ci.appveyor.com/project/mousetraps/nodejstools/build/1.0.147/job/4kgj98i6q666oj8b/tests 

It looks like a number of Analysis tests are now failing ( #731 will disable all the repl window tests and some of the other ones that are failing). These also fail locally.

If this change is the root cause, we should open a bug to track fixing this. Reverting this change is probably a bad idea given the number of files touched, and hopefully the fix is as easy as adjusting a few numbers (but it'd be nice if the tests were more robust).

mousetraps:
Interesting - I actually ran those tests successfully on my machine, since several of them are sensitive to line endings (e.g. ParserTests.cs requires CRLF). Anyways, it appears my assumptions about how git automatically decides what's best were invalid, so I took another shot at it in #736.

In any case, I agree that the tests shouldn't be so sensitive.

PR #735, opened 2016-03-07
Remove snippets tests

mjbvz:
As far as I can tell, these tests test visual studio itself and are therefore worthless. It's just more tests to maintain.


mousetraps:
Could you expand on that observation? Unfortunately snippets don't come for free in Visual Studio - the behavior must be separately implemented by each language service. See https://github.com/Microsoft/nodejstools/pull/339

mjbvz:
No you're right. I missed part of the logic.

PR #737, opened 2016-03-08
.gitattributes adventures (continued!) 

mousetraps:
Let git decide good defaults for which files qualify as text files and should be normalized.

Details here: https://github.com/Microsoft/nodejstools/pull/736#issuecomment-193960611


mjbvz:
Thanks for the detailed explanation. Change looks fine so feel free to merge it in, but it'll be interesting to see if appveyor passes this time.

PR #738, opened 2016-03-08
Fix warnings for having AbstractFilesystemPackageJsonTests use TestInitialize

mjbvz:
**Defect**
`AbstractFilesystemPackageJsonTests` generates a few test warnings because it has `[TestInitialize]` and `[TestUninitialize]` methods, but is not marked `[TestClass]`.

**Fix**
Remove these init methods and make `AbstractFilesystemPackageJsonTests` use an explicit  TemporaryFileManager for its functions. This makes it clearer what tests need this functionality.

I also renamed the `AbstractFilesystemPackageJsonTests` class since it now can be made static.


mousetraps:
:+1:

PR #739, opened 2016-03-08
Disable two more failing tests in AppVeyor

mjbvz:
These two tests only faily on specific platforms (or are non deterministic) in appveyor. They seem to pass locally.

This change just disables them on AppVeyor for now.


mousetraps:
❤️ CHECKMARKS!!!! 🎉🎉🎉

PR #741, opened 2016-03-09
Fix Format exception for Two JScript resource strings

mjbvz:
**Defect**
Currently, we throw an exception when trying to format the resource strings 'NoLeftCurly' or 'NoRightCurly' since they contain '{' and '}'.

**Defect**
Escape '{' and '}' so `string.Format` does not complain.


mousetraps:
:+1:

Also, in the future, can you please file a corresponding issue? At some point we'll want to write release notes and complete issue lists make those much easier to write :smiley:

Issue #702, opened 2016-02-23
Interactive Windows Hangs When Displaying Representation of Object with Many Properties


mjbvz:
## Repo
Try to display the representation of an object with lots of properties in the Node.js interactive window

```
> var a = []
undefined
> 
> for (var i = 0; i < 1000; ++i) a[i] = i;
999
> a  // hang here for about 5 sec
[ 0,
  1,
  2, 
  ...,
  999]
```

## Expected
Prints the representation without hanging.

## Actual
VS hangs for about 5 seconds and then prints the result. The entire application appears to be hanging as the window is marked as "(Not Responding)".

mjbvz:
I originally thought this had something to do with the debugger integration, but I think the root cause is actually `OutputBuffer.Flush`: 

```cs
  for (int i = 0; i < entries.Length; i++) {
                    var entry = entries[i];

                    _window.AppendOutput(entry.Properties.Color, entry.Buffer.ToString(), i == entries.Length - 1);
                }
                _window.TextView.Caret.EnsureVisible();
```

specifically `Repl.AppendOutput`  which seems to be comically slow. It took 60 seconds to print an array of 1000 ints while I was debugging.

mjbvz:
Improved this with #742. Closing this in favor of #387 which tracks a similar problem that would offer a more complete solution to this problem.

PR #742, opened 2016-03-10
Repl window output perf improvements

mjbvz:
Currently, printing large structures to the repl window can take a very long time (#702). I believe the primary cause is we end up appending output almost one character at a time, repeating the same operations over and over again inside of a loop.

This change updates ReplWindow to take a set of outputs to be appended instead of a single output. This allows moving some of the computations out of the inner loops and reusing previously computed values.

Improves #702 . Still not great performance but it is noticeably better.


mjbvz:
Confirmed that multicolor printing still seems to work:

image

Common/Product/ReplWindow/Repl/OutputBuffer.cs
```diff
@@ -298,7 +294,7 @@ struct OutputEntryProperties {
             }
         }
 
-        enum OutputEntryKind {
+        public enum OutputEntryKind {
```
mousetraps: internal

also check usages and see if it makes sense to move into ReplWindow altogether

Common/Product/ReplWindow/Repl/ReplWindow.cs
```diff
@@ -2110,14 +2110,18 @@ private sealed class CommandFilter : IOleCommandTarget {
         /// <summary>
         /// Appends text to the output buffer and updates projection buffer to include it.
         /// </summary>
-        internal void AppendOutput(ConsoleColor color, string text, bool lastOutput) {
+        internal void AppendOutput(OutputBuffer.OutputEntry[] entries) {
+            if (!entries.Any())
+                return;
```
mousetraps: add braces

Common/Product/ReplWindow/Repl/ReplWindow.cs
```diff
@@ -2110,14 +2110,18 @@ private sealed class CommandFilter : IOleCommandTarget {
...
+
+            var spans = new List<KeyValuePair<Span, OutputBuffer.OutputEntryProperties>>(entries.Length);
```
mousetraps: move this closer to the actual usage (also check in other places). Also see if there's better variable names or something that will make the behavior more easily understood

Common/Product/ReplWindow/Repl/ReplWindow.cs
```diff
@@ -2126,47 +2130,58 @@ private sealed class CommandFilter : IOleCommandTarget {
...
+                    startPosition += text.Length;
+                }
+                if (!_readingStdIn && !EndsWithLineBreak(text)) {
                     var lineBreak = GetLineBreak();
                     edit.Insert(oldBufferLength, lineBreak);
-                    newOutputLength += lineBreak.Length;
                     _addedLineBreakOnLastOutput = true;
-                } else {
-                    _addedLineBreakOnLastOutput = false;
+                    // Adust last span.
```
mousetraps: nit: this comment doesn't add much info - either make it more specific or remove

mousetraps:
minor comments and :+1: after that!

Common/Product/ReplWindow/Repl/ReplWindow.cs
mjbvz: This has to be outside of the using block here but I'll move it down a few lines.

PR #744, opened 2016-03-12
Fix for #15 - build: Numeric comparison error

### LINKED ISSUE #15: Opening up nodejstools in visual studio generates A numeric comparison was attempted on "$(VSUpdateVersion)" error
When trying to open the nodesjtool project in Visual Studio 2013, I receiver A numeric comparison was attempted on "$(VSUpdateVersion)" error on every project. I have some some searching and it looks like this issue is related to resharped, but i dont have resharped installed. How can I get the projects loaded as I would like to work on some customizations.

nodejstool_error_projectload

mjbvz:
@johnmcbride reported this error when opening the NodeJsProject with VS2013 #15 . He also came up with a solution, but this was never merged into our branch.

This change applies the suggested solution of removing Wow6432Node from the registry path. I confirmed that msbuild works, as does building through visual studio 2015.

Closes #15

### DIFF
--- a/Build/Common.Build.settings
+++ b/Build/Common.Build.settings
@@ -109,7 +109,7 @@
     definitions.
     -->
     <VSMajorVersion>$([System.Double]::Parse($(VSTarget)))</VSMajorVersion>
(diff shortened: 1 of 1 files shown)


mjbvz:
I also updated the build instructions for msbuild: https://github.com/Microsoft/nodejstools/wiki/Build-Instructions 

The old instructions did not work for me.

mousetraps:
:+1:

PR #747, opened 2016-03-14
Merge auto-dts branch into master

mjbvz:
This change merges the auto-dts branch work into master.


Common/Tests/Utilities/AssertUtil.cs
```diff
@@ -68,30 +68,24 @@ public static void Throws<TExpected>(Action throwingAction, string description)
...
-                    Assert.Fail("AssertUtils.ArrayEquals failure. Expected value {0} at position {1} but got value {2}",
-                        expected[i], i, actual[i]);
-                }
+            Assert.AreEqual(expected.Count(), actual.Count());
+
+            var expectedIt = expected.GetEnumerator();
+            var actualIt = actual.GetEnumerator();
+
+            while (expectedIt.MoveNext() && actualIt.MoveNext()) {
+                Assert.AreEqual(expectedIt.Current, actualIt.Current);
```
mousetraps: The previous loop was more verbose, but also more accurately pinpointed the error in the code

Common/Tests/Utilities/AssertUtil.cs
```diff
@@ -68,30 +68,24 @@ public static void Throws<TExpected>(Action throwingAction, string description)
...
             }
+            Assert.IsFalse(expectedIt.MoveNext());
+            Assert.IsFalse(actualIt.MoveNext());
```
mousetraps: Unnecessary - the Count check above should take care of any issues here. Additionally, the logic is off because both assertions would pass if `expected.Count() == actual.Count()+1`

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsControl.cs
```diff
@@ -33,12 +33,14 @@ public partial class NodejsNpmOptionsControl : UserControl {
 
         internal void SyncControlWithPageSettings(NodejsNpmOptionsPage page) {
             _showOutputWhenRunningNpm.Checked = page.ShowOutputWindowWhenExecutingNpm;
+            _enableAutoTypeAcquisition.Checked = page.EnableAutomaticTypeAcquisition;
```
mousetraps: I see `AutoType`, `AutoTyping`, and `AutoTypings` being used  - let's stick with the latter-most form to keep things consistent.

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsControl.cs
```diff
@@ -71,5 +73,13 @@ public partial class NodejsNpmOptionsControl : UserControl {
...
             }
         }
+
+        private void groupBox3_Enter(object sender, EventArgs e) {
+
+        }
+
+        private void checkBox1_CheckedChanged(object sender, EventArgs e) {
+
+        }
```
mousetraps: stray code

Nodejs/Product/Nodejs/Project/AbstractNpmNode.cs
```diff
@@ -68,13 +68,57 @@ protected AbstractNpmNode(NodejsProjectNode root)
...
+        private void RemoveUnusedModuleNodesFromHeirarchy(HierarchyNode parent, List<HierarchyNode> remove) {
+            foreach (var obsolete in remove) {
+                parent.RemoveChild(obsolete);
+                ProjectMgr.OnItemDeleted(obsolete);
+            }
+        }
+
+        private IEnumerable<IPackage> BuildModuleHeirarchy(HierarchyNode parent, IEnumerable<IPackage> modules, IReadOnlyDictionary<string, DependencyNode> recycle) {
+            if (modules == null)
+                return Enumerable.Empty<IPackage>();
```
mousetraps: braces around if statements

Nodejs/Product/Nodejs/Project/DependencyNode.cs
```diff
@@ -28,54 +28,25 @@
 using VsCommands2K = Microsoft.VisualStudio.VSConstants.VSStd2KCmdID;
 
 namespace Microsoft.NodejsTools.Project {
-    internal class DependencyNode : HierarchyNode {
+    internal interface IDependencyNodeBasePath {
+        string GetRelativeUrlFragment();
+    }
```
mousetraps: what's the purpose of this interface? It's only referenced once and the value-add is unclear.

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
```diff
@@ -130,20 +136,20 @@ private class NpmPathProvider : INpmPathProvider {
             }
         }
 
-        private INpmController CreateNpmController() {
-            if (null == _npmController) {
```
mousetraps: Any reason you removed this null check?

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
```diff
@@ -130,20 +136,20 @@ private class NpmPathProvider : INpmPathProvider {
...
+                false,
+                pathProvider);
+        }
+
+        private void RegisterWithNpmController(INpmController controller) {
+            controller.CommandStarted += NpmController_CommandStarted;
+            controller.OutputLogged += NpmController_OutputLogged;
+            controller.ErrorLogged += NpmController_ErrorLogged;
+            controller.ExceptionLogged += NpmController_ExceptionLogged;
+            controller.CommandCompleted += NpmController_CommandCompleted;
```
mousetraps: The purpose of this refactoring is unclear, and makes things less readable IMO.

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
```diff
@@ -362,9 +377,9 @@ private class NpmPathProvider : INpmPathProvider {
             }
         }
 
-#endregion
+        #endregion
 
-#region Updating module hierarchy
+        #region Updating module hierarchy
```
mousetraps: nit: If you format the document, do these #region tags stay in place? I was under the impression that they don't.

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
```diff
@@ -362,9 +377,9 @@ private class NpmPathProvider : INpmPathProvider {
...
 
         internal void ReloadHierarchySafe() {
```
mousetraps: Calling the typings acquisition code at ReloadHierarchy will cause serious performance issues because it will operates on the UI thread and gets called far more often than necessary for our purposes. Instead, move this logic to `NodejsProjectNode.OnIdleNodeModules`. Additionally, from `NodeModules`, you can add a method to GetTopLevelDependencies that you populate via `AddTopLevelModule`

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
```diff
@@ -378,39 +393,76 @@ private class NpmPathProvider : INpmPathProvider {
...
-                    ReloadHierarchy(_optionalModulesNode, optional);
-
-                    var root = controller.RootPackage.Modules.Where(package => 
-                        package.IsDependency || 
-                        !package.IsListedInParentPackageJson);
-                    
-                    ReloadHierarchy(this, root);
-                }
+            if (null == controller)
+                return;
```
mousetraps: braces around if statements.

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
```diff
@@ -378,39 +393,76 @@ private class NpmPathProvider : INpmPathProvider {
...
+                    null).ContinueWith(x => x);
+            }
+        }
+
+        private PackageSet.Diff ReloadPackageHierarchies(INpmController controller) {
+            return ReloadDevPackageHierarchy(controller)
+                .Concat(ReloadOptionalPackageHierarchy(controller))
+                .Concat(ReloadRootPackageHierarchy(controller))
+                .Concat(ReloadGlobalPackageHierarchy(controller));
+        }
```
mousetraps: As mentioned earlier, this isn't the right place for this code. Additionally the flow is pretty confusing, and if you choose to keep this refactoring even after moving the code out into `NodejsProjectNode`, please try and refactor them out into using the same method, otherwise the flow becomes annoying to decipher.

Nodejs/Product/Nodejs/Resources.resx
```diff
@@ -600,4 +600,13 @@ You will need to restart Visual Studio after installation.</value>
...
     <value>Surround With</value>
   </data>
+  <data name="TsdErrorOccured" xml:space="preserve">
+    <value>Error installing typings.</value>
+  </data>
+  <data name="TsdInstallCompleted" xml:space="preserve">
+    <value>Successfully installed typings.</value>
+  </data>
+  <data name="TsdNotInstalledError" xml:space="preserve">
+    <value>Could not find TSD. Make sure you have installed it globally.</value>
```
mousetraps: It's best to provide people with resolution steps whenever there are errors rather than just telling them what the error is, and pinpoint exactly what the error is. Additionally, in user-land, tsd means nothing to people. What is tsd? Why do I need it? So let's try to simplify this to user-language `"Error installing IntelliSense for package {0}: {1}", packageName, errorDescriptionAndResolution.` Same thing with the rest of these.

Nodejs/Product/Nodejs/TypingAcquisition.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Linq;
+using System.IO;
+
+using SR = Microsoft.NodejsTools.Project.SR;
+
+namespace Microsoft.NodejsTools {
+    static class TypingAcquisition {
+        private const string TsdExe = "tsd.cmd";
+
+        public static async Task<bool> AquireTypings(
```
mousetraps: spelling: `Acquire`

Nodejs/Product/Nodejs/TypingAcquisition.cs
```diff
@@ -0,0 +1,93 @@
...
+            if (!packages.Any()) {
+                return true;
+            }
+
+            var tsdPath = GetTsdPath(pathToRootNpmDirectory);
+            if (tsdPath == null) {
+                if (redirector != null)
+                    redirector.WriteErrorLine(SR.GetString(SR.TsdNotInstalledError));
+                return false;
+            }
```
mousetraps: Try installing globally before complaining.

Nodejs/Product/Nodejs/TypingAcquisition.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            using (var process = ProcessOutput.Run(
+                tsdPath,
+                TsdInstallArguments(packages),
+                pathToRootProjectDirectory,
+                null,
+                false,
+                redirector,
+                quoteArgs: true)) {
```
mousetraps: Can you verify that this'll actually timeout at some point? I know npm fails to do this, so we're left with lingering processes if there's something up with the internet connection...

Nodejs/Product/Nodejs/TypingAcquisition.cs
```diff
@@ -0,0 +1,93 @@
...
+                    }
+                    return false;
+                }
+            }
+        }
+
+        private static string GetTsdPath(string pathToRootNpmDirectory) {
+            var path = Path.Combine(pathToRootNpmDirectory, TsdExe);
+            if (File.Exists(path))
+                return path;
```
mousetraps: braces around single line ifs

Nodejs/Product/Npm/INpmController.cs
```diff
@@ -37,5 +37,7 @@ public interface INpmController : INpmLogSource, IDisposable {
         IPackageCatalog MostRecentlyLoadedCatalog { get; }
 
         string ListBaseDirectory { get; }
+
+        string FullPathToRootPackageDirectory { get; }
```
mousetraps: Unnecessary. You should be able to access this via `RootPackage.Path`

Nodejs/Product/Npm/PackageSet.cs
```diff
@@ -0,0 +1,110 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.NodejsTools.Npm {
+    /// <summary>
+    /// Immutable collection of packages.
+    /// </summary>
+    public class PackageSet : IEnumerable<IPackage> {
```
mousetraps: This logic can be simplified greatly. Instead of taking package diffs, let's just keep a list of package names that we've already tried downloading via tsd + the list of .d.ts files that we've already downloaded (it should be sufficient to query the package directory for any .d.ts files). Then simply compare this combined list against the names of the top level packages. This will be much simpler and less expensive. Again, we're not going for 100% correctness here - there's absolutely no way we're going to achieve that given the current state of DefinitelyTyped.

Nodejs/Product/Npm/PackageSet.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// <summary>
+        /// Different between a left package set and right package set.
+        /// </summary>
+        public class Diff {
```
mousetraps: This entire class strikes me as pretty opaque. 
```cs
List<string> GetMissingTypings(List<string> topLevelPackages, List<string> typingsFiles, List<string> typingsCache)
```
is more descriptive and gets to the heart of what we're actually trying to do here.

Nodejs/Product/Npm/SPI/NodeModules.cs
```diff
@@ -167,6 +148,31 @@ internal class NodeModules : AbstractNodeModules {
 
             return depth;
         }
+
+        private static IEnumerable<KeyValuePair<string, IPackageJson>> GetTopLevelDependencies(string modulesBase) {
```
mousetraps: Rename to `GetTopLevelPackageDirectories` - these aren't really the top level dependencies in npm v3.

Nodejs/Product/Npm/SPI/NodeModules.cs
```diff
@@ -167,6 +148,31 @@ internal class NodeModules : AbstractNodeModules {
...
+                if (moduleDir.Length < NativeMethods.MAX_FOLDER_PATH && !_ignoredDirectories.Any(toIgnore => moduleDir.EndsWith(toIgnore))) {
+                    IPackageJson json = null;
+                    try {
+                        json = PackageJsonFactory.Create(new DirectoryPackageJsonSource(moduleDir));
+                    } catch (PackageJsonException) {
+                        // Fail gracefully if there was an error parsing the package.json
+                        Debug.Fail("Failed to parse package.json in {0}", moduleDir);
+                    }
+                    if (json != null)
+                        yield return new KeyValuePair<string, IPackageJson>(moduleDir, json);
```
mousetraps: braces around all if statements.

Nodejs/Product/Npm/SPI/NpmController.cs
```diff
@@ -63,7 +63,7 @@ internal class NpmController : AbstractNpmLogSource, INpmController {
             } catch (NpmNotFoundException) { }
         }
 
-        internal string FullPathToRootPackageDirectory {
+        public string FullPathToRootPackageDirectory {
```
mousetraps: keep this internal

Nodejs/Product/Npm/SPI/RootPackage.cs
```diff
@@ -25,13 +25,34 @@ internal class RootPackage : IRootPackage {
...
+        :
+            this(LoadRootPackageJson(fullPathToRootDirectory), fullPathToRootDirectory, showMissingDevOptionalSubPackages, allModules, depth)
+        { }
+    
+        public RootPackage(
+            IPackageJson packageJson,
+            string fullPathToRootDirectory,
+            bool showMissingDevOptionalSubPackages,
+            Dictionary<string, ModuleInfo> allModules = null,
+            int depth = 0)
```
mousetraps: Why the extra constructor if it's only being used once? Again, just makes things less readable because you have to follow a flow chain when reading through the code.

Nodejs/Tests/NpmTests/PackageSetTests.cs
```diff
@@ -0,0 +1,138 @@
...
+using System.Collections.Generic;
+using TestUtilities;
+using Microsoft.NodejsTools.Npm;
+using System.Linq;
+using Moq;
+
+namespace NodejsTests {
+    [TestClass]
+    public class PackageSetTests {
+        [TestMethod]
```
mousetraps: Add test categories

mousetraps:
Really looking forward to this :smiley:. Left a bunch of comments. Lmk if you have any Qs!

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
mjbvz: This method got turned from a method that sets a member variable, to a factory method. The actual member variable is set in the ctor now instead of inside of `CreateNpmController`.

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
mjbvz: Dependency injection. I was originally looking at mocking out the INpmController, and one way to do that would be to inject a mock `NpmControllerFactory`. 

I ended up not needing to do this, but this change makes it more clear that getting the instance from the factory and registering the controller are two separate steps.

Nodejs/Product/Nodejs/TypingAcquisition.cs
mjbvz: I think we agreed previously that this was not part of the minimum required work for this first stage of the work.

Common/Tests/Utilities/AssertUtil.cs
mjbvz: I can add back in the index, but the string formatting is just overhead. The assertions already tell you what was expected and what was found.

mjbvz: Here'e an example of the new output: 

```
Message: Assert.AreEqual failed. Expected: <IPackageProxy>. Actual: <IPackageProxy>. Assertion failed at index 0.
```

Before it was:

```
Message: Assert.AreEqual failed. Expected: <IPackageProxy>. Actual: <IPackageProxy>. AssertUtils.ArrayEquals failure. Expected value <IPackageProxy> at position 0 but got value <IPackageProxy>
```

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsControl.cs
mjbvz: Fixed the places I noticed this, but let me know if you see any other places that use the wrong names

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
mjbvz: Yes, I don't recall touching these lines directly in the first place so I think they were auto formatted. I can revert the change, but this seems to be what other places are doing.

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
mjbvz: Thanks, this sounds like a better idea and should cut down the size of the change considerably.

mjbvz:
Thanks for the comments. 

I'm marking all the issues that have been addressed with 🎉 . I'm not sure if there is a better way to track comment status other than leaving the reply. Let me know if this is unclear or you have some other recommendation (code flow seemed to be a lot better for this).

Fixed a few easy to fix problems, and replied to a few comments to clarify the intention.

Now working on moving the code to NodejsProject node, which will be a larger update. Should get that in sometime this afternoon.

Nodejs/Tests/NpmTests/PackageSetTests.cs
mjbvz: Deleting this entire class.

Nodejs/Product/Npm/PackageSet.cs
mjbvz: Deleting this entire class. 

I think the `PackageSet` concept is a good [Whole value](http://c2.com/ppr/checks.html) that we may want to revisit at some point, but it's not required for this change.

Nodejs/Product/Npm/PackageSet.cs
mjbvz: Good idea. Merging in this part of the logic to `TypingsAcquisition` and removing this class.

Nodejs/Product/Nodejs/Resources.resx
mjbvz: I agree that we should either: install Tsd automatically if it is not found, or, as you state, tell that user what steps they should take. I think we are targeting the second option for now.

For errorDescriptionAndResolution, how about `Could not find TSD TypeScript package manager. Please install it globally by running 'npm install -g tsd'`.

I think that clearly says what is wrong and how to fix it. We may also want to include a link to a help page or something.

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsPage.cs
```diff
@@ -23,6 +23,7 @@ namespace Microsoft.NodejsTools.Options {
     [ComVisible(true)]
     public class NodejsNpmOptionsPage : NodejsDialogPage {
         private const string ShowOutputWindowRunningNpm = "ShowOutputWindowRunningNpm";
+        private const string EnableAutomaticTypeAcquisitionSetting = "EnableAutomaticTypeAcquisition";
```
mjbvz: will rename to `EnableAutomaticTypingsAcquisitionSetting `

Common/Tests/Utilities/AssertUtil.cs
```diff
@@ -68,29 +68,20 @@ public static void Throws<TExpected>(Action throwingAction, string description)
...
-                    Assert.Fail("AssertUtils.ArrayEquals failure. Expected value {0} at position {1} but got value {2}",
-                        expected[i], i, actual[i]);
-                }
+            Assert.AreEqual(expected.Count(), actual.Count());
+
+            var expectedIt = expected.GetEnumerator();
+            var actualIt = actual.GetEnumerator();
+            for (var i = 0; expectedIt.MoveNext() && actualIt.MoveNext(); ++i) {
+                Assert.AreEqual(expectedIt.Current, actualIt.Current, string.Format("Assetion failed at index {0}", i));
             }
```
mousetraps: why the enumerator? How about:
```
for (var i = 0; i < expected.Count; i++) {
    Assert.AreEqual(expected[i], actual[i], string.Format("Assertion failed at index {0}", i));
}
```

Nodejs/Product/Nodejs/Project/DependencyNode.cs
mousetraps: did you see this comment?

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
mousetraps: That makes sense, but we should only optimize for patterns if we're actually using them. Otherwise we get into situations where we have to keep hoping around the code in order to actually get to the heart of what it's doing. Right now this code is

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -95,6 +96,33 @@ public NodejsProjectNode(NodejsProjectPackage package)
...
             }
+
+            TryToAcquireTypings();
+        }
+
+        private bool ShouldAcquireTypingsAutomatically {
+            get {
+                return !_projectHasTypeScriptFiles
+                    && NodejsPackage.Instance.IntellisenseOptionsPage.EnableES6Preview
+                    && NodejsPackage.Instance.NpmOptionsPage.EnableAutomaticTypingsAcquisition;
```
mousetraps: Dev15 Salsa is supposed to automatically acquire typings, so let's ifdef this check so that the feature only lights up in Dev14.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -95,6 +96,33 @@ public NodejsProjectNode(NodejsProjectPackage package)
...
+                        package.IsDependency
+                        || package.IsOptionalDependency
+                        || package.IsDevDependency
+                        || !package.IsListedInParentPackageJson);
+
+                TypingsAcquisition.AcquireTypings(
+                    controller.ListBaseDirectory,
+                    controller.RootPackage.Path,
+                    packages,
+                    null).ContinueWith(x => x);
```
mousetraps: Interesting continuation... what exactly was the intention here?

Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.IO;
+
+using SR = Microsoft.NodejsTools.Project.SR;
+
+namespace Microsoft.NodejsTools {
+    static class TypingsAcquisition {
+        private const string TsdExe = "tsd.cmd";
+        private const string TypingsDirecctoryName = "typings";
```
mousetraps: Spelling

Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -0,0 +1,123 @@
...
+            IEnumerable<IPackage> packages,
+            Redirector redirector) {
+            if (!packages.Any()) {
+                return true;
+            }
+
+            var tsdPath = GetTsdPath(pathToRootNpmDirectory);
+            if (tsdPath == null) {
+                if (redirector != null)
+                    redirector.WriteErrorLine(SR.GetString(SR.TsdNotInstalledError));
```
mousetraps: braces

Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -0,0 +1,123 @@
...
+            }
+
+            using (var process = ProcessOutput.Run(
+                tsdPath,
+                TsdInstallArguments(packages),
+                pathToRootProjectDirectory,
+                null,
+                false,
+                redirector,
+                quoteArgs: true)) {
```
mousetraps: We shouldn't invoke tsd if there are no additional typings to acquire.

Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        private static string GetTsdPath(string pathToRootNpmDirectory) {
+            var path = Path.Combine(pathToRootNpmDirectory, TsdExe);
+            return File.Exists(path) ? path : null;
+        }
+
+        private static string GetPackageTsdName(IPackage package) {
+            return package.Name;
+        }
```
mousetraps: nit: Can you include these below the method that is invoking them? Makes it easier to read through the code.

Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        private static IEnumerable<string> TsdInstallArguments(IEnumerable<IPackage> packages) {
```
mousetraps: Method names should start with verbs (eg `GetTsdInstallArguments`) here and everywhere else.

Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -0,0 +1,123 @@
...
+            return new[] { "install", }.Concat(packages.Select(GetPackageTsdName)).Concat(new[] { "--save" });
+        }
+
+        private static IEnumerable<IPackage> TypingsToAcquire(string pathToRootProjectDirectory, IEnumerable<IPackage> packages) {
+            var currentTypings = new HashSet<string>(CurrentTypingsPackages(pathToRootProjectDirectory));
+            return packages.Where(package =>
+                !currentTypings.Contains(GetPackageTsdName(package)));
```
mousetraps: Let's simplify to `package.Name` for now, and create `GetPackageTsdName` when it's necessary. Consider the experience of debugging through this code... the best case scenario is that you get to just hover on the variable to see the input, and unfortunately you can't do that with a method like this.

Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        private static IEnumerable<string> CurrentTypingsPackages(string pathToRootProjectDirectory) {
```
mousetraps: In addition to this, we should be storing a list of files we've already tried to acquire, and we shouldn't attempt to acquire them again for the duration of that session.

Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -0,0 +1,123 @@
...
+            var typingsDirectoryPath = Path.Combine(pathToRootProjectDirectory, TypingsDirecctoryName);
+            if (Directory.Exists(typingsDirectoryPath)) {
+                foreach (var file in Directory.EnumerateFiles(typingsDirectoryPath, "*.d.ts", SearchOption.AllDirectories)) {
```
mousetraps: We need some more robust error handling here... should at least catch IOException.

Nodejs/Product/Npm/PackageComparer.cs
```diff
@@ -30,4 +30,22 @@ public class PackageComparer : IComparer<IPackage> {
...
+                && p1.IsMissing == p2.IsMissing
+                && p1.IsOptionalDependency == p2.IsOptionalDependency;
+        }
+
+        public override int GetHashCode(IPackage obj) {
+            if (obj.Name == null || obj.Version == null)
+                return obj.GetHashCode();
+            return obj.Name.GetHashCode() ^ obj.Version.GetHashCode();
+        }
+    }
```
mousetraps: Do we need this anymore?

Nodejs/Product/Npm/SPI/RootPackage.cs
mousetraps: did you see this comment?

Nodejs/Tests/NpmTests/NpmTests.csproj
```diff
@@ -81,6 +81,7 @@
   <ItemGroup>
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="Microsoft.VisualStudio.QualityTools.UnitTestFramework" />
+    <Reference Include="Moq, Version=4.2.1312.1622, Culture=neutral, PublicKeyToken=69f491c39445e920, processorArchitecture=MSIL" />
```
mousetraps: Are we adding any new tests?

mousetraps:
Left a bunch more comments. Almost there!

Nodejs/Product/Npm/PackageComparer.cs
mjbvz: Yes, it's an extracted version of the inline logic we had in `ReloadHeirarchy`. It was used by `PackageSet` as well, but the extraction still makes sense even with that code removed as `ReloadHeirarchy` uses it.

Common/Tests/Utilities/AssertUtil.cs
mjbvz: IEnumerable does not support random access. I can revert this change because it's not strictly used anymore, but the old definition is more restrictive than needed.

Nodejs/Product/Nodejs/Project/DependencyNode.cs
mjbvz: I can changes in this file. The intention is that `DependencyNode` does not actually depend on having a `DependencyNode` as it's parent, but rather some object that can provide a base path. We can mock out the latter object much more easily than mocking out a `DependencyNode`. But I never ended up going down the path of testing this code so it's not required.

Nodejs/Product/Nodejs/TypingsAcquisition.cs
mjbvz: This consistently defines how we map IPackages to strings. It doesn't matter that the implementation is a single line.

This sort of abstraction is useful because when looking at the code, since you only have to know that we map the IPackage to a string somehow. This also makes any future logic changes (such as including the version as part of the string) much easier and safer since we only have to update a single function, instead of every consumer of this logic.

Nodejs/Product/Nodejs/TypingsAcquisition.cs
mjbvz: This is handled on the first line of this function, which `return true` if the passed in package set is empty.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
mjbvz: There is a warning if you don't consume the result of an async function. A noop continuation is one easy way to get around that.

Common/Tests/Utilities/AssertUtil.cs
mousetraps: Ohhh, I totally missed that you changed the signature... I thought the brace had just moved to the next line. Yeah, your call. :smiley:

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsControl.Designer.cs
```diff
@@ -160,14 +166,61 @@ partial class NodejsNpmOptionsControl {
...
+            this.groupBox3.AutoSizeMode = System.Windows.Forms.AutoSizeMode.GrowAndShrink;
+            this.groupBox3.Controls.Add(this.tableLayoutPanel5);
+            this.groupBox3.Dock = System.Windows.Forms.DockStyle.Top;
+            this.groupBox3.Location = new System.Drawing.Point(6, 131);
+            this.groupBox3.Margin = new System.Windows.Forms.Padding(6, 3, 6, 3);
+            this.groupBox3.Name = "groupBox3";
+            this.groupBox3.Size = new System.Drawing.Size(376, 14);
+            this.groupBox3.TabIndex = 4;
+            this.groupBox3.TabStop = false;
+            this.groupBox3.Text = "Type Acquisition";
```
mousetraps: Just as we did with the error messages, let's stay consistent with the language and also make it clearer to the end user what's actually going on here.

Nodejs/Product/Npm/PackageComparer.cs
mousetraps: Ahh, doesn't look like anything got pulled from ReloadHierarchy then. Additionally, there's a PackageComparer above - can you take a look at how that's being used see whether it makes sense to consolidate anything?

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsPage.cs
```diff
@@ -47,6 +48,8 @@ public NodejsNpmOptionsPage()
         /// </summary>
         public bool ShowOutputWindowWhenExecutingNpm { get; set; }
 
+        public bool EnableAutomaticTypingsAcquisition { get; set; }
```
mousetraps: Hmm... I may have mislead you here... I just checked on Dev15 and turns out that they don't expose any option in the UI - currently they only plan to expose it through a typingOptions section in a project-specific jsconfig.json (which we won't support in Dev14). So for now, I think that means we have a couple options to reduce confusion:
* only display the option in Dev14
* remove the option altogether

I'm leaning towards the latter, but it's your call.

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsControl.Designer.cs
```diff
@@ -160,14 +166,61 @@ partial class NodejsNpmOptionsControl {
...
+            // 
+            this._enableAutoTypingsAcquisition.AutoSize = true;
+            this._enableAutoTypingsAcquisition.Checked = true;
+            this._enableAutoTypingsAcquisition.CheckState = System.Windows.Forms.CheckState.Checked;
+            this._enableAutoTypingsAcquisition.Dock = System.Windows.Forms.DockStyle.Top;
+            this._enableAutoTypingsAcquisition.Location = new System.Drawing.Point(3, 3);
+            this._enableAutoTypingsAcquisition.Name = "_enableAutoTypingsAcquisition";
+            this._enableAutoTypingsAcquisition.Size = new System.Drawing.Size(364, 17);
+            this._enableAutoTypingsAcquisition.TabIndex = 2;
+            this._enableAutoTypingsAcquisition.Text = "Enable automatic typings acquisition for JavaScript projects";
```
mousetraps: -> "Automatically download IntelliSense typings for JavaScript projects (ES6 Preview)"

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsPage.cs
mjbvz: @mousetraps Just to be super clear, for option two, that would be: remove the setting in the UI all together?

mousetraps: Yes, I was referring to removing the UI option, and auto dts would be enabled by default for all NTVS projects (and you wouldn't be able to turn it off). The other benefit of this is that users will come to us and describe the issue they're running into so we can fix the problem that makes them want to turn it off. We can always add the option back then and put out a dev build if it is indeed problematic. 

But just to be super clear, the options are whatever we want them to be :wink:

Nodejs/Product/Npm/SPI/RootPackage.cs
mjbvz: Reverted this file.

Nodejs/Product/Nodejs/Project/AbstractNpmNode.cs
```diff
@@ -68,13 +68,58 @@ protected AbstractNpmNode(NodejsProjectNode root)
 
         public abstract void ManageNpmModules();
 
-        protected void ReloadHierarchy(HierarchyNode parent, IEnumerable<IPackage> modules) {
+        protected IEnumerable<IPackage> ReloadHierarchy(HierarchyNode parent, IEnumerable<IPackage> modules) {
```
mousetraps: this should still return void.

Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -0,0 +1,143 @@
...
+                return packages;
+            }
+            try {
+                foreach (var file in Directory.EnumerateFiles(typingsDirectoryPath, "*.d.ts", SearchOption.AllDirectories)) {
+                    var directory = Directory.GetParent(file);
+                    if (directory.FullName != typingsDirectoryPath && Path.GetFullPath(directory.FullName).StartsWith(typingsDirectoryPath)) {
+                        packages.Add(directory.Name);
+                    }
+                }
+            } catch (SystemException e) when (e is IOException || e is SecurityException || e is UnauthorizedAccessException) {
```
mousetraps: man, already making breaking changes to VS2013! :wink: For the sake of our CI... let's at least wait until after we have a conditional AppVeyor build configuration before we start using all the fancy new features. :smiley:

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -49,6 +49,9 @@ class NodejsProjectNode : CommonProjectNode, VsWebSite.VSWebSite, INodePackageMo
         internal readonly RequireCompletionCache _requireCompletionCache = new RequireCompletionCache();
         private string _intermediateOutputPath;
         private readonly Dictionary<NodejsProjectImageName, int> _imageIndexFromNameDictionary = new Dictionary<NodejsProjectImageName, int>();
+        private bool _projectHasTypeScriptFiles = false;
```
mousetraps: why not just check the EnableTypescript project property?

mjbvz: `EnableTypescript` can only be called on a UI thread, which I realized after extracting out duplicate code for that property.

Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -0,0 +1,143 @@
...
+        public TypingsAcquisition(
+            string pathToRootNpmDirectory,
+            string pathToRootProjectDirectory) {
+            _pathToRootNpmDirectory = pathToRootNpmDirectory;
+            _pathToRootProjectDirectory = pathToRootProjectDirectory;
+
+            _acquiredTypingsPackageNames = new Lazy<HashSet<string>>(() => {
+                return new HashSet<string>(CurrentTypingsPackages(_pathToRootProjectDirectory));
+            });
+        }
```
mousetraps: Did you get a chance to look into using the Singleton pattern here yet?

Nodejs/Product/Nodejs/Project/DependencyNode.cs
```diff
@@ -95,7 +62,7 @@ internal class DependencyNode : HierarchyNode {
 
         #region HierarchyNode implementation
 
-        private string GetRelativeUrlFragment() {
+        public string GetRelativeUrlFragment() {
```
mousetraps: -> private

Nodejs/Product/Nodejs/TypingsAcquisition.cs
mjbvz: The way things are setup right now, we need the typings files in each project, so a global singleton does not make sense. Instead, there is a single instance of this class stored on the `NodeProjectNode`.

mousetraps: Yes, but there can be multiple NodejsProjectNodes in a solution. DownloadTypings is pretty componentized - have you considered pulling that out?

mjbvz: Are you proposing that we have a top level `typings` directory instead of one in each project's root? For example, with projects A & B, the current setup is:

```
A/
    ...
    package.json
    typings/...
B/
    ...
    package.json
    typings
```

With the singleton, are you instead proposing:

```
typings/...
A
    ...
    package.json
B
    ...
    package.json
```

mousetraps: Let's say you open a solution with 20 projects - we're going spin up 20 tsd instances as the project is loading, all of which are potentially longer-running and are going to spike the CPU at the same time. Additionally, you might get into a situation where we're we're printing out 20 error messages about tsd not being found (no need to focus on that one just yet, though). The ideal solution would be to have a single always-on service (so we don't have to wait for separate node processes to start) which every project can call into, and that's what we're aiming for in Dev15. For this, however, I think that "good enough" means that we're okay with spinning up several instances of tsd, but just make sure that happens in series rather than in parallel.

mjbvz: OK. I know we talked about this, but singleton is a very confusing name for this.

I think each project would still own a `TypingsAquisition`, but we would have a static, "tsd" work queue with a concurrency of `1` that they all enqueue work items on. This ensures that only a single task is executing, no matter how many instances there are.

mousetraps:
:+1: 

I'm going to go ahead and merge these changes in. Known issues that remain to be addressed include:
* automatically download tsd if it is not already downloaded
* automatically download node.d.ts file even if no packages are installed
* more comprehensive testing (with special attention to performance)

Issue #679, opened 2016-02-04
VS locks up during ReloadHierarchy


mousetraps:
from @RobertSmart
```
 	[Managed to Native Transition]	
 	mscorlib.dll!System.IO.File.FillAttributeInfo(string path, ref Microsoft.Win32.Win32Native.WIN32_FILE_ATTRIBUTE_DATA data, bool tryagain, bool returnErrorOnNotFound)	Unknown
 	mscorlib.dll!System.IO.Directory.InternalExists(string path, out int lastError)	Unknown
 	mscorlib.dll!System.IO.Directory.InternalExistsHelper(string path, bool checkHost)	Unknown
 	Microsoft.NodejsTools.Npm.dll!Microsoft.NodejsTools.Npm.SPI.Package.IsMissing.get()	Unknown
 	Microsoft.NodejsTools.dll!Microsoft.NodejsTools.Project.AbstractNpmNode.ReloadHierarchy.AnonymousMethod__0(Microsoft.NodejsTools.Npm.IPackage module)	Unknown
 	[Frames below may be incorrect and/or missing, no binaries loaded for System.Core.dll]	
 	Microsoft.NodejsTools.dll!Microsoft.NodejsTools.Project.AbstractNpmNode.ReloadHierarchy(Microsoft.VisualStudioTools.Project.HierarchyNode parent, System.Collections.Generic.IEnumerable<Microsoft.NodejsTools.Npm.IPackage> modules)	Unknown
 	Microsoft.NodejsTools.dll!Microsoft.NodejsTools.Project.AbstractNpmNode.ReloadHierarchy(Microsoft.VisualStudioTools.Project.HierarchyNode parent, System.Collections.Generic.IEnumerable<Microsoft.NodejsTools.Npm.IPackage> modules)	Unknown
 	Microsoft.NodejsTools.dll!Microsoft.NodejsTools.Project.AbstractNpmNode.ReloadHierarchy(Microsoft.VisualStudioTools.Project.HierarchyNode parent, System.Collections.Generic.IEnumerable<Microsoft.NodejsTools.Npm.IPackage> modules)	Unknown
>	Microsoft.NodejsTools.dll!Microsoft.NodejsTools.Project.NodeModulesNode.ReloadHierarchy()	Unknown
 	mscorlib.dll!System.Threading.Tasks.Task.InnerInvoke()	Unknown
 	mscorlib.dll!System.Threading.Tasks.Task.Execute()	Unknown
[...]

PR #752, opened 2016-03-16
Limit Execution time of IsMissing check for #679

mjbvz:
Investigating #679. I can't repo the problem, but from the provided dump, believe that one work around may be to limit how long `Directory.Exists` spends executing. This call may hang for network drives and in other cases.

This logic adds a 250ms timeout to the call. If the call times out, or the directory does not exist, we return that the package is missing.

### DIFF
--- a/Nodejs/Product/Npm/SPI/Package.cs
+++ b/Nodejs/Product/Npm/SPI/Package.cs
@@ -18,6 +18,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Threading.Tasks;
 
 namespace Microsoft.NodejsTools.Npm.SPI {
     internal class Package : RootPackage, IPackage {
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Npm/SPI/Package.cs
```diff
@@ -55,7 +56,14 @@ internal class Package : RootPackage, IPackage {
         }
 
         public bool IsMissing {
-            get { return IsListedInParentPackageJson && !Directory.Exists(Path); }
+            get {
+                if (!IsListedInParentPackageJson)
```
mjbvz: Will add braces 😀

mousetraps:
:-1:

If we think the hang is due to a network drive, we should ask if it's a network drive, not assume. And if that is indeed the case, this issue is likely to occur elsewhere in the code and fixing it in this one location probably won't do much good in the grand scheme of things. 

If the dump is indeed accurate... More likely what is occurring is that we are stuck in a very long loop rather than hanging at this one call.

mjbvz:
Thanks for the clarification regarding the broader issue.

Just note that using `Task` [does seem to be a fine way to prevent timeouts for network drives](http://stackoverflow.com/questions/726602/how-to-prevent-timeout-when-inspecting-unavailable-network-share-c-sharp) since both `Directory.Exists` and `DirectoryInfo` can hang.

Closing since this may not be the root cause of #679.

PR #757, opened 2016-03-18
Vs14u2 add dts files to project

mjbvz:
There are two problems with Intellisense in Dev14 Update 2, when using typescript for Javascript Intellisense:
* You have to rename the d.ts files to trigger Intellisense to work. Just including the file in the project is not enough
* typings files are not automatically picked up or added to the project.

This change fixes both of these issues.

The first for the first is triggering an `OnItemAdded` to get an included d.ts file noticed by type script

The second part is fixed by auto including any file node that is in the typings directory.

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
@@ -17,6 +17,7 @@
 using System;
 using System.IO;
 using Microsoft.VisualStudioTools.Project;
+using Microsoft.VisualStudioTools;
 #if DEV14_OR_LATER
 using Microsoft.VisualStudio.Imaging.Interop;
 using Microsoft.VisualStudio.Imaging;
@@ -36,6 +37,16 @@ namespace Microsoft.NodejsTools.Project {
             } else {
                 Analyze();
             }
+        }
+
+        protected override void OnParentSet(HierarchyNode parent) {
+#if DEV14_OR_LATER
+            if (Url.EndsWith(".d.ts", StringComparison.OrdinalIgnoreCase) && Url.IndexOf(@"\typings\", StringComparison.OrdinalIgnoreCase) >= 0) {
+                ProjectMgr.Site.GetUIThread().Invoke(() => {
+                    this.IncludeInProject(true);
+                });
+            }
+#endif
         }
 
         internal void Analyze() {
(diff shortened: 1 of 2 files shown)


Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
```diff
@@ -17,6 +17,7 @@
 using System;
 using System.IO;
 using Microsoft.VisualStudioTools.Project;
+using Microsoft.VisualStudioTools;
```
mousetraps: sort usings

Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
```diff
@@ -38,6 +39,16 @@ public NodejsFileNode(NodejsProjectNode root, ProjectElement e)
             }
         }
 
+        protected override void OnParentSet(HierarchyNode parent) {
+#if DEV14
+            if (Url.EndsWith(".d.ts", StringComparison.OrdinalIgnoreCase) && Url.IndexOf(@"\typings\", StringComparison.OrdinalIgnoreCase) >= 0) {
+                ProjectMgr.Site.GetUIThread().Invoke(() => {
+                    this.IncludeInProject(true);
+                });
```
mousetraps: is it necessary to invoke the UI thread here? Let's double check before merging in.

Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
```diff
@@ -80,16 +91,22 @@ public NodejsFileNode(NodejsProjectNode root, ProjectElement e)
...
-            
+
             UpdateParentContentType();
             ItemNode.ItemTypeChanged += ItemNode_ItemTypeChanged;
 
+#if DEV14
+            ProjectMgr.Site.GetUIThread().Invoke(() => {
+                ProjectMgr.OnItemAdded(this.Parent, this);
+            });
+#endif
```
mousetraps: here too

mousetraps:
@mjbvz is on vacation this week, but this change is staged to go into our upcoming release, so before merging in... I'll pull this branch down, and make corrections as I see fit. Biggest concern is the UI thread calls - would be great to limit those if possible.

mousetraps:
Okay, just looked into the calls here and verified invoking from the UI thread is necessary in both instances. Merging in.

PR #772, opened 2016-03-27
#687 Enable Salsa by default

### LINKED ISSUE #687: Enable Salsa by default
We should enable Salsa by default once we have automatic .d.ts acquisition

mousetraps:
- Preview flag is on by default
- change name of old settings so they are not used instead
- require TS tools version 1.8

close #687


mjbvz:
Looks good.

My only comment is that `NodeLs*`is not very clear name for someone new looking at the code. Perhaps expand it to `NodeLanguageService*` or come up with another name to describe those settings.

mousetraps:
@mjbvz - we might want to change the name up again, so I didn't want to spoil the best one :smile:

PR #780, opened 2016-03-28
Make sure auto-dts is disabled for typescript projects

### LINKED ISSUE #776: Auto-dts should not be enabled for typescript projects
This floods the error list because the versions are often mismatched.

mjbvz:
**Defect**
Previously, auto typings acquisition was based on an opt-out flag. We opted a project out of typings acquisition if it had any TypeScript files. This caused a race condition for projects that had not fully loaded.

**Fix**
This inverts the condition to make it opt-in instead. Only if we fully load a project and find that it contains no typescript files do we enable auto-dts.

Closes #776

### DIFF
--- a/Nodejs/Product/Nodejs/NodejsConstants.cs
+++ b/Nodejs/Product/Nodejs/NodejsConstants.cs
@@ -14,10 +14,13 @@
 //
 //*********************************************************//
 
+using System;
+
 namespace Microsoft.NodejsTools {
     internal class NodejsConstants {
         internal const string JavaScriptExtension = ".js";
         internal const string TypeScriptExtension = ".ts";
+        internal const string TypeScriptDeclarationExtension = ".d.ts";
         internal const string MapExtension = ".map";
         internal const string NodejsProjectExtension = ".njsproj";
 
@@ -69,5 +72,14 @@ namespace Microsoft.NodejsTools {
         internal const string TypeScript = "TypeScript";
 
         internal const string NodeToolsProcessIdEnvironmentVariable = "_NTVS_PID";
+
+        /// <summary>
+        /// Checks whether a relative and double-backslashed seperated path contains a folder name.
+        /// </summary>
+        public static bool ContainsNodeModulesOrBowerComponentsFolder(string path) {
+            string tmp = "\\" + path + "\\";
+            return tmp.IndexOf("\\" + NodeModulesFolder + "\\", StringComparison.OrdinalIgnoreCase) >= 0
+                || tmp.IndexOf("\\" + BowerComponentsFolder + "\\", StringComparison.OrdinalIgnoreCase) >= 0;
+        }
     }
 }
--- a/Nodejs/Product/Nodejs/NodejsConstants.cs
+++ b/Nodejs/Product/Nodejs/NodejsConstants.cs
@@ -76,7 +76,7 @@ namespace Microsoft.NodejsTools {
         /// <summary>
         /// Checks whether a relative and double-backslashed seperated path contains a folder name.
         /// </summary>
-        public static bool ContainsNodeModulesOrBowerComponentsFolder(string path) {
+        internal static bool ContainsNodeModulesOrBowerComponentsFolder(string path) {
             string tmp = "\\" + path + "\\";
             return tmp.IndexOf("\\" + NodeModulesFolder + "\\", StringComparison.OrdinalIgnoreCase) >= 0
                 || tmp.IndexOf("\\" + BowerComponentsFolder + "\\", StringComparison.OrdinalIgnoreCase) >= 0;
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
@@ -51,7 +51,6 @@ namespace Microsoft.NodejsTools.Project {
         private readonly Dictionary<NodejsProjectImageName, int> _imageIndexFromNameDictionary = new Dictionary<NodejsProjectImageName, int>();
 
 #if DEV14
-        private bool _projectIsOnlyJavascript = false;
         private TypingsAcquisition _typingsAcquirer;
 #endif
 
@@ -108,7 +107,7 @@ namespace Microsoft.NodejsTools.Project {
 #if DEV14
         private bool ShouldAcquireTypingsAutomatically {
             get {
-                return _projectIsOnlyJavascript
+                return !IsTypeScriptProject
                     && NodejsPackage.Instance.IntellisenseOptionsPage.EnableES6Preview;
             }
         }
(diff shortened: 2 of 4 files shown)


Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -244,8 +244,10 @@ public NodejsProjectNode(NodejsProjectPackage package)
         }
 
         protected override void FinishProjectCreation(string sourceFolder, string destFolder) {
+            bool foundTypeScriptFile = false;
             foreach (MSBuild.ProjectItem item in this.BuildProject.Items) {
                 if (String.Equals(Path.GetExtension(item.EvaluatedInclude), NodejsConstants.TypeScriptExtension, StringComparison.OrdinalIgnoreCase)) {
+                    foundTypeScriptFile = true;
```
mousetraps: Hmm this check looks like it'll capture both .d.ts and .ts files. Is that correct? If so we should try to limit it to just .ts

mousetraps: Also let's exclude node_modules and bower_components from his search since that shouldn't have any bearing on whether the project is a TS project. 

Additionally, we may want to consider just checking if EnableTypeScript flag is set since we explicitly have separate TS and JS project types.

mousetraps:
Left a couple comments and :+1: after that!

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
mjbvz: That sounds good. I think your point about `node_modules` and `bower_components` exposes a flaw with the existing logic. Updating both of the conditionals to handle those cases seems like the correct behavior, as right now, we can end up enabling typescript for projects that only include TS files in those locations.

Will update the review shortly to fix this and also use `EnableTypeScript` by dispatching to the UI thread.

mjbvz:
@mousetraps Please take another quick look at this review when you have a chance. The update was a little more substantial, so it probably needs to be reviewed again.

Nodejs/Product/Nodejs/NodejsConstants.cs
```diff
@@ -69,5 +72,14 @@ internal class NodejsConstants {
         internal const string TypeScript = "TypeScript";
 
         internal const string NodeToolsProcessIdEnvironmentVariable = "_NTVS_PID";
+
+        /// <summary>
+        /// Checks whether a relative and double-backslashed seperated path contains a folder name.
+        /// </summary>
+        public static bool ContainsNodeModulesOrBowerComponentsFolder(string path) {
```
mousetraps: internal, not public

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -202,7 +201,11 @@ public NodejsProjectNode(NodejsProjectPackage package)
 
         public bool IsTypeScriptProject {
             get {
-                return string.Equals(GetProjectProperty(NodejsConstants.EnableTypeScript), "true", StringComparison.OrdinalIgnoreCase);
+                var task = ProjectMgr.Site.GetUIThread().InvokeAsync(() => {
+                    return string.Equals(GetProjectProperty(NodejsConstants.EnableTypeScript), "true", StringComparison.OrdinalIgnoreCase);
+                });
+                task.Wait();
+                return task.Result;
```
mousetraps: why invoke the UI thread here? If it is indeed necessary to do so, we should probably cache the value so we would only need to make a blocking call once per session.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -243,10 +246,15 @@ public NodejsProjectNode(NodejsProjectPackage package)
             get { return NodejsConstants.IssueTrackerUrl; }
         }
 
+        private static bool IsProjectTypeScriptSourceFile(string path) {
+            return string.Equals(Path.GetExtension(path), NodejsConstants.TypeScriptExtension, StringComparison.OrdinalIgnoreCase)
+                && !string.Equals(Path.GetExtension(path), NodejsConstants.TypeScriptDeclarationExtension, StringComparison.OrdinalIgnoreCase)
+                && !NodejsConstants.ContainsNodeModulesOrBowerComponentsFolder(path);
+        }
```
mousetraps: nit: can you move this below the methods that are referencing it?

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -202,7 +201,11 @@ public NodejsProjectNode(NodejsProjectPackage package)
 
         public bool IsTypeScriptProject {
             get {
```
mousetraps: Does the race condition still exist here? Except now, we might not download typings to some JS projects? Or is this why you're making use of the UI thread?

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
mjbvz: Yes, this must be called on the UI thread for some reason, which is why I created the separate variable to track this for the initial dts work. This latest change tried to reuse this variable so that they could not get out of sync.

I'm not sure the value can be cached though, since that will change the behavior of our existing code. What I can do though is move the task logic to the call sites of the property. That way, only that caller has the additional overhead.

mousetraps: Yep, let's move the logic to the caller.

mousetraps: Also, an we check on the other callers here? Are they on the UI thread? If so, we should also assert that we're on the UI thread here so we don't run into issues like this in the future.

mjbvz: `GetProjectProperty` asserts if invoked on any thread except the UI thread.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
mjbvz: I believe that this handled the main race condition. There are still a few potential problems, but these are much more corner cases.

The latest iteration check a project property to tell if we are in a typescript project or not. That handles the main race condition we were running into, since that property should be accurate, even before the project has fully loaded.

One edge case is when you have a JavaScript project with a typescript file in it. In that case, we mark the project as a typescript project. But before this takes place, we may try to acquire typings. I don't see any easy way around this case.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
mousetraps: awesome. thanks for checking.

mousetraps:
:+1:

PR #784, opened 2016-03-29
Support unc paths for command execution 

### LINKED ISSUE #755: support UNC paths when running processes
From VS Feedback:

I have my Documents folder set to use a UNC path via VMWare Fusion's sharing feature to share my Mac OS X Documents folder with Windows. When using NTVS 1.1 and trying to install missing npm packages using the option in Visual Studio the output pane reports that CMD.exe cannot use UNC paths and thus the NPM command fails because CMD didn't open with the correct folder and so NPM then tries to install packages into C:\Windows instead of the project folder.

image

mjbvz:
**Defect**
#755 Unc paths cause a problem currently when running commands .

**Fix**
Use pushd to set the working directory when running a command. Tested by opening a project from  `\\localhost\c$`.

closes #755


Common/Product/SharedProject/ProcessOutput.cs
```diff
@@ -237,27 +234,18 @@ sealed class ProcessOutput : IDisposable {
...
-            } else {
-                args = string.Join(" ", arguments.Where(a => a != null));
-            }
-            psi.Arguments = string.Format("/S /C \"{0} {1} >>{2} 2>>{3}\"",
-                QuoteSingleArgument(filename),
-                args,
-                QuoteSingleArgument(outFile),
-                QuoteSingleArgument(errFile)
-            );
-            psi.WorkingDirectory = workingDirectory;
```
mousetraps: Hrm... I'm concerned about the implications of not setting the WorkingDirectory property. In particular, the working directory property has different implications depending on the value of `UseShellExecute`. Did you verify the edge cases here? 
https://msdn.microsoft.com/en-us/library/system.diagnostics.processstartinfo.workingdirectory(v=vs.110).aspx

Common/Product/SharedProject/ProcessOutput.cs
```diff
@@ -184,23 +184,20 @@ sealed class ProcessOutput : IDisposable {
...
-            psi.CreateNoWindow = !visible;
-            psi.UseShellExecute = false;
-            psi.RedirectStandardError = !visible || (redirector != null);
-            psi.RedirectStandardOutput = !visible || (redirector != null);
-            psi.RedirectStandardInput = !visible;
+                    errorEncoding);
+            }
+
+            var psi = new ProcessStartInfo("cmd.exe") { 
+                Arguments = string.Format(@"/S /C pushd {0} & {1} {2}",
```
mousetraps: does pushd have a timeout? What happens if there is no network access or the drive is inaccessible?

Common/Product/SharedProject/ProcessOutput.cs
mjbvz: I tried testing the edge cases I could think of, but this is widely used code that is also shared with common so I'm sure I missed some areas. My understanding is that `pushd` replaces the need to set the working directory for executing the actual command. We are only relying on the initial, default  working directory to pick up `cmd.exe`.

Any specific edge cases you had in mind?

mousetraps: yes, see the UseShellExecute stuff in the link above. In particular... 
> When the UseShellExecute property is false, gets or sets the working directory for the process to be started. When UseShellExecute is true, gets or sets the directory that contains the process to be started.

Common/Product/SharedProject/ProcessOutput.cs
mjbvz: Yes, between 45 and 120 seconds from what I can gather (it seem to differ by Windows version.) I don't see any easy way to adjust the timeout for a single command.

Not ideal, but also not a huge problem. It's more an external dependency that we should be aware of when trying to diagnose reported problems.

Common/Product/SharedProject/ProcessOutput.cs
mjbvz: I think my question is more: what testing would make you more confident in the change?

I generally understand what `UseShellExecute` and `WorkingDirectory` do, but also believe that, with this change, working directory is only going to be used directly for looking up `cmd.exe`. Once we actually run `node ...`, `pushd` has already changed the working directory.

I verified all the basic scenarios against this change, while also running node programs to look at `process.cwd()` and opening relative file paths. Any other thoughts?

Common/Product/SharedProject/ProcessOutput.cs
```diff
@@ -184,23 +184,20 @@ sealed class ProcessOutput : IDisposable {
...
+                    QuoteSingleArgument(workingDirectory),
+                    QuoteSingleArgument(filename),
+                    GetArguments(arguments, quoteArgs)),
+                CreateNoWindow = !visible,
+                UseShellExecute = false,
```
mousetraps: make this true

mjbvz: Nvrmind, reverted the suggested fix. This has to be false because we set the `Redirect*` properties. The elevated example manually redirects to a file, which is why it can get away with setting this to `true`.

PR #785, opened 2016-03-29
Fix for #765 - Open With and Window/open

### LINKED ISSUE #765: Problem opening multiple views of the same JavaScript (.js) file using 'window > new window'
# Problem opening multiple views of the same JavaScript (.js) file using 'window > new window'

## Bug:
Cannot open multiple views of the same JavaScript (.js) file using Visual Studo command 'window > new window'.

### Expected result
Second instance / view of the file opens in a new tab

### Actual result
**Get a dialogue message:**
'The document [FILE LOCATION].js is already open. Do you want to close it? Yes / No'

Clicking Yes = Closes current tab and opens a new tab. Therefore still only one tab open.
Clicking No = No action (close dialogue)

### Project types affected
NTVS projects (.njsproj) e.g. 'Blank Node.js Web Application'.

## Verification / testing
Works as expected when:
- JavaScript file (.js) in 'standard' project type e.g. 'ASP.NET Web Application'
- JavaScript file opened outside of any project

## Environment
- Windows 7 Professional, Service Pack 1
- Visual Studio 2015 Version 14.0.24720.00 Update 1
- NTVS v.1.1

mjbvz:
#765 - Can't open the same js file in mutliple windows

I believe the root cause is `InitializeLanguageService ` where we are passing in the `NodejsEditorFactory` guid instead of the `NodejsLanguageServiceGuid`. This causes `InitializeLanguageService` to throw an exception.

Change also removed a copy and pasted guid string.

closes #765

### DIFF
--- a/Nodejs/Product/Nodejs/Guids.cs
+++ b/Nodejs/Product/Nodejs/Guids.cs
@@ -20,8 +20,7 @@ using System;
 
 namespace Microsoft.NodejsTools
 {
-    static class Guids
-    {
+    static class Guids {
         public const string NodejsPackageString = "FE8A8C3D-328A-476D-99F9-2A24B75F8C7F";
         public const string NodejsCmdSetString = "695e37e2-c6df-4e0a-8833-f688e4c65f1f";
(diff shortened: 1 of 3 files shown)


mousetraps:
:+1:

PR #787, opened 2016-03-29
Auto install TSD if it is not installed

### LINKED ISSUE #771: auto-dts should automatically download tsd/typings if not already installed
currently we prompt the user to install if it doesn't exist, which isn't a great first-run experience.

mjbvz:
For auto type acquisition, try to automatically install tsd globablly if we do not find the exe. This is only attempted once per session.

closes #771

### DIFF
--- a/Nodejs/Product/Nodejs/TypingsAcquisition.cs
+++ b/Nodejs/Product/Nodejs/TypingsAcquisition.cs
@@ -14,34 +14,39 @@
 //
 //*********************************************************//
 
-using System.Threading.Tasks;
-using Microsoft.VisualStudioTools.Project;
-using Microsoft.NodejsTools.Npm;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.IO;
-
-using SR = Microsoft.NodejsTools.Project.SR;
-using System;
 using System.Security;
 using System.Threading;
+using System.Threading.Tasks;
+
+using Microsoft.VisualStudioTools.Project;
+using Microsoft.NodejsTools.Npm;
+using Microsoft.NodejsTools.Npm.SPI;
+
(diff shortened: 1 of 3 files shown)


Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -65,27 +70,21 @@ internal class TypingsAcquisition {
...
             }
 
-            var tsdPath = GetTsdPath(pathToRootNpmDirectory);
-            if (tsdPath == null) {
-                if (redirector != null) {
-                    redirector.WriteErrorLine(SR.GetString(SR.TsdNotInstalledError));
-                }
+            string tsdPath = await EnsureTsdInstalled();
+            if (string.IsNullOrEmpty(tsdPath)) {
+                redirector.WriteErrorLine(SR.GetString(SR.TsdNotInstalledError));
```
mousetraps: any reason you removed the null check on the redirector?

mjbvz: No, good catch.

Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -14,34 +14,39 @@
...
-using SR = Microsoft.NodejsTools.Project.SR;
-using System;
 using System.Security;
 using System.Threading;
+using System.Threading.Tasks;
+
+using Microsoft.VisualStudioTools.Project;
+using Microsoft.NodejsTools.Npm;
+using Microsoft.NodejsTools.Npm.SPI;
+
```
mousetraps: nit: can you remove the spaces between usings - I don't think VS's automatic-usings-sort works otherwise.

mousetraps:
minor comments and :+1: after that!

PR #788, opened 2016-03-29
Always install node.d.ts typing file

### LINKED ISSUE #770: auto-dts should automatically download node.d.ts on all Node.js project types
right now it only triggers on package installs

mjbvz:
This change ensures that we install `node.d.ts` for intellisense, even if `node` is not explicitly required by another installed package. 

I think we should merge this change in because it makes things better. However, there is currently an issue where `OnIdleNodeModules` is never fired for projects without any node modules. Because this callback is never fired, we never try to acquire any typings. I'm looking into handling that, but a proper fix may involve changing the existing behavior of the code.

closes #770

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
@@ -152,11 +152,9 @@ namespace Microsoft.NodejsTools.Project {
                     package.IsDependency
                     || package.IsOptionalDependency
                     || package.IsDevDependency
-                    || !package.IsListedInParentPackageJson)
-                .Select(package => package.Name)
-                .Concat(new[] { "node" });
+                    || !package.IsListedInParentPackageJson);
 
-            TryToAcquireTypings(currentPackages);
+            TryToAcquireTypings(currentPackages.Select(package => package.Name));
         }
 #endif
 
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
@@ -100,7 +100,7 @@ namespace Microsoft.NodejsTools.Project {
                 }
             }
 #if DEV14
-            TryToAcquireTypings();
+            TryToAcquireCurrentTypings();
 #endif
         }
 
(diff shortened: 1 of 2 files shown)


mousetraps:
Gotcha. How about instead of handling node.d.ts in `OnIdleNodeModules`, we handle just this one case in `ReloadHierarchy`?

mjbvz:
How about on `FinishProjectCreation`? That's where we create the TS `Scripts/typings` stuff currently.

mousetraps:
Pretty sure that only gets fired when the project is being created. Not if it's being opened.

Also, oops - I totally meant `NodejsProjectNode.Reload`, not `ReloadHierarchy`

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -133,21 +133,30 @@ public NodejsProjectNode(NodejsProjectPackage package)
...
-                || package.IsDevDependency
-                || !package.IsListedInParentPackageJson);
+            var currentPackages = controller.RootPackage.Modules
+                .Where(package =>
+                    package.IsDependency
+                    || package.IsOptionalDependency
+                    || package.IsDevDependency
+                    || !package.IsListedInParentPackageJson)
+                .Select(package => package.Name)
+                .Concat(new[] { "node" });
```
mousetraps: I don't think you need to include "node" here at all anymore since Reload will always get called first.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -482,6 +489,10 @@ public NodejsProjectNode(NodejsProjectPackage package)
                 NodejsPackage.Instance.CheckSurveyNews(false);
                 ModulesNode.ReloadHierarchySafe();
 
+#if DEV14
+                TryToAcquireTypings(new[] { "node" });
```
mousetraps: Could you verify that line *always* gets called in Reload first? Would be more performant to call tsd just once if possible.

mousetraps:
Minor comment and LGTM after that!

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
mjbvz: Yes, when I was debugging I see `Reload` is always call before the npm stabilization callback. We can't easily avoid having two tsd calls in that case, one on `Reload` to get `node.d.ts` and one after we pick up all the node modules to install their `typings`.

Is that what you were asking?

mousetraps: What if you open an existing folder that doesn't have the typings installed - will this line get executed first? Or what if you create a new project?

Basically, I'm wondering if there's some set of cases that we can anticipate where there are many packages installed and we could execute tsd just once.

mousetraps:
We can sort out the remaining comment later - going to go ahead and merge in the changes now and kick off a build.

Issue #761, opened 2016-03-22
remove Dev11 and Dev12 support


mousetraps:
The v1.1 release will continue to support Dev11, Dev12, and Dev14, but we are shifting focus to Dev14 and Dev15  for future releases (mostly Dev15, as we'll be dependent upon many of the new platform improvements, but we're otherwise hoping to maintain parity.)

Issue #763, opened 2016-03-22
Create conditional AppVeyor configurations


mousetraps:
* `master` should target VS 2015, VS vNext
* `v1.1.x` should target VS 2012, VS 2013, and VS 2015

PR #791, opened 2016-03-29
Remove dev11 and dev12 guards in NodeJs files

mjbvz:
Part of #761. We now only support dev14 and dev15, so these guards are not needed going forward.

Does not touch files in `common`.

### DIFF
--- a/Common/Tests/MockVsTests/MockVsRunningDocumentTable.cs
+++ b/Common/Tests/MockVsTests/MockVsRunningDocumentTable.cs
@@ -19,11 +19,7 @@ using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell.Interop;
 
 namespace Microsoft.VisualStudioTools.MockVsTests {
-    class MockVsRunningDocumentTable : IVsRunningDocumentTable
-#if DEV12_OR_LATER
-        , IVsRunningDocumentTable4
-#endif
-        {
+    class MockVsRunningDocumentTable : IVsRunningDocumentTable, IVsRunningDocumentTable4 {
         private readonly MockVs _vs;
         private readonly Dictionary<uint, DocInfo> _table = new Dictionary<uint, DocInfo>();
         private readonly Dictionary<string, uint> _ids = new Dictionary<string, uint>(StringComparer.OrdinalIgnoreCase);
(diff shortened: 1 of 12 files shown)


Common/Tests/MockVsTests/MockVsRunningDocumentTable.cs
```diff
@@ -19,11 +19,7 @@
 using Microsoft.VisualStudio.Shell.Interop;
```
mjbvz: Whoops, guess I did touch this one file in common too. I'll revert it if needed.

mjbvz:
Looks like #763 is required before we can merge this in.

mousetraps:
If we are going to tackle this, we should do a full fix, rather than sets of partial fixes which simply clutter up the history and make it harder to verify that we actually got to everything. Additionally, please remove relevant conditionals from installer, csproj, etc. and undo any changes in Common.

mjbvz:
@mousetraps Should we create a new development branch for these changes?

I want to keep PR's as small, focused, and as easy to review as possible. That's why this change only targets the `NodeJs` project compiler guards (except for that one `common` file sneaking in too :smile:). I'm much more in favor of small, targeted, incremental changes that make the code better somehow, but may not fix everything, with issues tracking when something is actually 'done'. That model doesn't work for all changes, but is a good general guide.

Trying to handle large changes in a single PR makes the PRs harder to review and it makes it harder to properly respond to feedback.

mjbvz:
Let's see if this builds now that the appveyor.yaml change #763 is merged in.

mousetraps:
Yeah, feel free to create a dev branch for this if you'd find it helpful. Re: reviewing - one of the things I want to ensure is that we don't miss anything and that's very difficult to do when the changes are coming in piecemeal. In this case, there is quite a bit missing, so it's not entirely clear what sort of comments are even valuable. If the full review is too much, I'll typically browse through the individual commits (especially helpful when they follow a logical train of thought), but only complete logical pieces should be brought into master. also makes it much easier to look back on in the future when we want to do something similar.

DanielRosenwasser:
:+1:  for granular commits with complete PRs, it makes it difficult for me to follow the sort of work going on on the team.

mjbvz:
Will submit combined PR that removes all dev11+dev12 support.

PR #792, opened 2016-03-30
fix branding

mousetraps:


Nodejs/Setup/BuildRelease.ps1
```diff
@@ -248,7 +248,7 @@ $managed_files = (
 $native_files = @()
 
 $supported_vs_versions = (
-    @{number="15.0"; name="VS vNext"; build_by_default=$true},    
+    @{number="15.0"; name="VS 15"; build_by_default=$true},
```
mjbvz: Is this the consumer facing name as well? It seems a little confusing to have `VS 15` and `VS 2015`.

mjbvz:
Looks good. Just was curious about the branding names.

PR #793, opened 2016-03-30
revert back to requiring TS version 1.6 so that we don't mess with pe…

mousetraps:
…ople's settings unnecessarily.

### DIFF
--- a/Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsPage.cs
+++ b/Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsPage.cs
@@ -34,7 +34,7 @@ namespace Microsoft.NodejsTools.Options {
         private bool _showCompletionListAfterCharacterTyped;
         private string _toolsVersion;
         private readonly bool _enableES6Preview;
-        private readonly Version _typeScriptMinVersionForES6Preview = new Version("1.8");
+        private readonly Version _typeScriptMinVersionForES6Preview = new Version("1.6");
 
         public NodejsIntellisenseOptionsPage()
             : base("IntelliSense") {


mjbvz:
👍

Was there any reason the version was bumped in the first place or anything we lose if a user has 1.6 or 1.7 vs 1.8+ ?

mousetraps:
Yes, the experience is significantly worse prior to VS Update 2, but there are some folks who've started using it.

mjbvz:
It's not clear to me why we would want to support a significantly worse experience. Could you please explain a bit more to help me understand the problem.

Is it that the 'ES6 Intellisense' option is persisted somehow and may be incorrectly enabled after this update, even though the user does not have 1.8+?

Or is it that people have been using ES6 preview option without TS 1.8+, and actually found it helpful even in this more broken form? In the latter case, we should figure out a way to tell the users that they should upgrade to TS 1.8+. Having a bad / crippled user experience like this seems worse than disabling the option entirely and offering a explanation of what is wrong and how to fix it.

mousetraps:
Sorry for being terse yesterday - was trying to kick off a new build in time for the release. 

Previously if users installed NTVS 1.2 alpha, and fired up vs before installing vs update 2, the setting will revert to the old intellisense experience without ES6. This setting will be saved to the registry, which means that even when they install update 2, this will be the default experience. Now it will use the crippled ES6 experience *until* the user installs vs update 2. 

In both cases, the user ends up with not what they expected, but the latter case is more easily repairable. 

In our release notes, we say we require Update 2. I agree that this is not the best way to do it, but I wasn't keen on making any more wide-sweeping changes hours before a release. 

Going forward, you are absolutely right that we should strive to better inform the user about the issue.

PR #795, opened 2016-03-30
Revert "Fix potential parse error for large hex numbers"

mousetraps:
Reverts Microsoft/nodejstools#783


mousetraps:
too close to a release to take this change.

PR #798, opened 2016-03-31
Use appveyor.yaml for CI config

mjbvz:
This allows us to use some more fancy configuration options for CI.

Creates two configurations, one for master and one for v1.1.x. Master no longer builds VS12, but v1.1.x will

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,9 +10,6 @@
     - nodejs_version: 4
       vs_version: 14.0
       architecture: x64
-    - nodejs_version: 0.12
-      vs_version: 12.0
-      architecture: x64
     - nodejs_version: 5
       vs_version: 14.0
       architecture: x86
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,20 +16,29 @@
   install: &default_install_script
     - ps: >-
         Install-Product node $env:nodejs_version $env:architecture
+
         node -v
+
         "C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\Tools\VsDevCmd.bat"
+
         $node = which node
+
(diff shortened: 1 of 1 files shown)


mjbvz:
We will have to merge this into `v1.1.x` too. AppVeyor recommends that you use a common `appveyor.yml` across all branches to avoid merge conflicts.

mjbvz:
Also, after this change, the AppVeyor UI settings will no longer do anything..

mousetraps:
:+1: 

Regarding our earlier discussion... what happens in the event that we are locking down for a release and potentially submitting patches to a different branch? Do we need to add a new entry to the AppVeyor run every time?

mjbvz:
@mousetraps Yes, but there are a few options in my understanding: 

* Before creating the new branch, update the `appveyor.yaml` to in the parent branch to reference the new branch as well. 
* Update the `appveyor.yaml` to use simple patterns for the branch name matching. That would allow us to build all `v1.x.x` branches for a given config.
* Create a fallback, default entry in the `appveyor.yaml` that builds any branch not explicitly matched.

# New branch creation process
If neither branch name glob matching or the fallback would work, and we instead want to make new branches fully opt-in, I believe that the steps to create a new branch would look somewhat like this (new branch is called `v2.7.1` and branched off of `master` and using its config in this example):

* In `master`, add a new entry for `v2.7.1` to the `only` block in the `master` AppVeyor config.
  * We generally want to update master instead of editing the `appveyor.yaml` in the branch to avoid merge conflicts if `v2.7.1` is merged back in.
* Submit the `appveyor.yaml` change to `master`
* Create branch `v2.7.1`

Now any changes pushed to `v2.7.1` should get built using the updated config.

PR #799, opened 2016-03-31
Appveyor.yaml change for v1.1.x branch

mjbvz:
Same change as #798 but applied to v1.1.x branch


mousetraps:
see https://github.com/Microsoft/nodejstools/pull/798#issuecomment-205709318

PR #804, opened 2016-03-31
Revert "Revert "Show full list of completions for Array/String/Number/Boolean/etc constructors""

mjbvz:
Since we now have shipped 1.2 Alpha, it now seems safe to merge this in. It addresses an open issue and should improve the overall user experience. Reverts Microsoft/nodejstools#794


mousetraps:
:+1:

PR #807, opened 2016-03-31
Gate VS 2015 *.d.ts include in project

### LINKED ISSUE #805: Automatic typing acquisition adds duplicate declaration files to the project
**Version**: NTVS 1.2 Alpha

**Bug 1**

When a project contains a folder called `typings`, every time the project is rebuilt, the `.d.ts` files in that folder are automatically included in the project file. 

This appears to be the expected behaviour, however, the files are included in ***duplicate***.

To reproduce

* Open up the project in the attached zip file.
* Right-click on the project and ***rebuild***.
* Unload and edit the project and observer the multiple copies of `myTypes.d.ts`.

Some observations

* This only affects `.d.ts` files. The file `bar.ts` is not duplicated.
* Only files in the `typings` folder are affected. The file `foo.d.ts` in folder `typings2` is not duplicated.
* A new copy is added after every rebuild.

[NodejsConsoleApp1.zip](https://github.com/Microsoft/nodejstools/files/197627/NodejsConsoleApp1.zip)

**Bug 2**

The automatic acquisition should only occur when the following is true:

* You are using the NTVS 1.2 Alpha.
* You are on VS 2015 (not VS “15”)
* You have a Javascript project (not Typescript)
* Intellisense level is set to “ES6 Preview”.

However, the above problem occurs under the following:

* A typescript NTVS project in VS 2015
* JavaScript intellisense has been switched off.

mjbvz:
**Defect**
We added some new auto include-in-project logic for *.d.ts files in order to support Salsa Intellisense for Javascript. Unfortunately, while the typings acquisition is gated properly, the logic to automatically add `*.d.ts` files to the VS solution was not gated currently.

This results in *.d.ts files being automatically included in TypeScript projects, potentially breaking people who already have a `typings` directory

**Fix**
Gate the include-in-project work.

Switched to using feature gating too in `NodeJsFileNode` instead of compiler switches. This is generally more maintainable.

closes #805

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
@@ -40,8 +40,7 @@ namespace Microsoft.NodejsTools.Project {
         }
 
         protected override void OnParentSet(HierarchyNode parent) {
-#if DEV14
-            if (ProjectMgr == null || ProjectMgr.Analyzer == null) {
+            if (ProjectMgr == null || ProjectMgr.Analyzer == null || !ProjectMgr.ShouldAcquireTypingsAutomatically) {
                 return;
             }
 
@@ -50,7 +49,6 @@ namespace Microsoft.NodejsTools.Project {
                     this.IncludeInProject(true);
                 });
             }
-#endif
         }
 
         internal void Analyze() {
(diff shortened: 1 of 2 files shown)


mousetraps:
:+1:

PR #809, opened 2016-03-31
Check the existance of the target dir in `GetTopLevelPackageDirectories`

mjbvz:
Just check if the directory existing before attempting to enumerate over it.

No user bad here, but this has been annoying me while debugging. Plus, there's no reason to throw an exception if we don't have to.


mousetraps:
:+1:

PR #814, opened 2016-04-01
Ensure Types are only Automatically Acquired for ES6 Preview

### LINKED ISSUE #813: Disable Automatic Type Acquisition for Other Intellisense levels
In Alpha 1.2, typings are still being acquired for all levels of Intellisense. They should only be acquired for "ES6 preview."

mjbvz:
This previously was gated incorrectly using the `EnableES6Preview` property, which actually just tells if the "ES6" preview option should be shown at all.

We should gate using the selected Analysis level instead. 

closes #813


mousetraps:
:+1:

Issue #264, opened 2015-07-08
Possible Memory Leak


atom0s:
Some info on my system:
 - Intel Core i7-4700MQ (2.40GHz)
 - 8GB RAM
 - Windows 8.1 Pro x64
 - Visual Studio 2013 Ultimate w/ Update 4

This is a fresh reformat as of 6 days ago. I have nothing on this system except for Visual Studio, MySQL (server and tools), and Chrome. 

Here is the results of some testing:

**Idle Launch**
On launch of Visual Studio, with no project loaded, I am idle at ~150MB of RAM usage by the devenv.exe

**Open C# Application Solution**
Loading up a C# application solution, I peaked at ~300MB of RAM usage. Which after loading and a few minutes passed, some garbage collection occurred and dropped the usage down to about ~275MB.

**Open Node.JS Solution**
Loading up a Node.JS solution immediately jumps the RAM usage to ~500MB while loading, and continues to climb almost consistently at ~100MB per second. At peak, I hit around 2.5GB of RAM usage with the amount constantly adjusting over the 2GB mark after the project has been fully loaded and the IDE is sitting idle.
Image: http://i.imgur.com/lkfpqS9.png

Testing wise, I do believe the leak (if it even is a leak) is in the intellisense that the tools apply. If the Node.JS text editor settings are altered to completely disable the intellisense and remove the saving to disk, the usage idles out at about 400MB of RAM.

Not too sure if this is a leak or just how the intellisense works, but figured I'd report it since it was very noticeable on my laptop as it started to make the CPU fan spin louder while having VS open while using a NodeJS project.

kant2002:
1. What version of NTVS you are using?
2. How big you project code and how big is you dependencies?
3. What Intellisense options is set: Full or Quick?

atom0s:
* NTVS.Dev.2015-06-14.VS.2013.msi
  * Project is a RESTful API backend. Nothing too huge. Dependencies wise, the node_modules folder is at 228 folders, most of which are dependencies of dependencies. We specifically include about 20ish modules in our package.json file.
  * I do not see an option for 'Quick', as for what setting was made, it was set to whatever the toolset installs default with. I did not alter any settings before making this report. 

I have Node.js Tools intellisense completely disabled now and do not have this issue anymore. However I am unsure what, if anything, I am losing by having it off since intellisense seems to be working well still.

mleanos:
[...]

Issue #144, opened 2015-05-28
Clean up public API surface


mousetraps:
Eventually we'll want to consider documenting it and officially supporting extensibility. Right now, though, it cannot be depended upon.

mjbvz:
Closing as out of scope for NTVS project currently. 

If someone has a specific application or need to extend NTVS in this way, please let me know and we can discuss this further.

PR #815, opened 2016-04-01
Reduce memory usage by not holding onto Json.net objects

mjbvz:
**Bug**
After creating a new Express 4 app, about 6MB memory is currently being used to hold onto Json.net objects. With more complex dependency trees, memory usage for these objects is even higher.

**Fix**
Updated many locations to no longer hold onto the Json.net objects, but serialize the values on construction. This adds a bit more initial computation overhead, but allows the Json.net objects to be collected, lowering memory usage accordingly.

Also removed a few classes that were only used in tests. If we ever need them, they're in the source control.


mjbvz:
Related finding in to #264

Nodejs/Product/Npm/IBugs.cs
```diff
@@ -1,22 +0,0 @@
...
-//    Unless required by applicable law or agreed to in writing, software 
-//    distributed under the License is distributed on an "AS IS" BASIS, 
-//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or 
-//    implied. See the License for the specific language governing 
-//    permissions and limitations under the License.
-//
-//*********************************************************//
-
-namespace Microsoft.NodejsTools.Npm {
-    public interface IBugs {
```
DinoV: This all appears to be public surface area, is it an okay time to be making all of these breaking changes?

mjbvz: @mousetraps would know more, but I don't think we officially support #144 extensions to NTVS or a public API at the moment.

PR #822, opened 2016-04-05
Remove Nodejsfilenode watcher logic

mjbvz:
This logic is currently all commented out. This change just removes it. It is always in source control if we need to restore it


mousetraps:
:+1:

PR #825, opened 2016-04-05
Fix line endings for files added in #804

mjbvz:
PR #804 merged in some rather old files that had the wrong style line endings. This causes these files to always show as edited on my machine.

This change just updates the line endings of those files to be consistent with the rest of our repo using a `git add` and a `git commit`.


mousetraps:
:+1:

PR #827, opened 2016-04-05
Resubmit of PR #555 - Fix some failing tests

mjbvz:
This is a resubmission of PR #555. Fixed a few merge conflicts and enabled the four tests that were previously failing.


Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsPage.cs
```diff
@@ -113,7 +113,7 @@ public NodejsIntellisenseOptionsPage()
             }
         }
 
-        internal bool OnlyTabOrEnterToCommit {
+        public bool OnlyTabOrEnterToCommit {
```
mousetraps: no need to make this public - we should add InternalsVisibleTo if this is just for tests

mousetraps:
minor comment and :+1: after that!

PR #828, opened 2016-04-05
Pass setting to runtests.ps1 script

mjbvz:
Small change to automatically pass `settings` flag to `vstest.console.exe` for in our `runtests.ps1` script to simplify running UI tests.

### DIFF
--- a/Tools/runtests.ps1
+++ b/Tools/runtests.ps1
@@ -21,6 +21,6 @@ if ($selectedVsVersion -eq '') {
 	return
 }
 
-$testFiles = $tests | % { ".\BuildOutput\Release$selectedVsVersion\Tests\$_.dll" }
+$testFiles = $tests | % { ".\BuildOutput\Debug$selectedVsVersion\Tests\$_.dll" }
 
-& "$vstest" @args $testFiles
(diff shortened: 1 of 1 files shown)


mjbvz:
Looks like today's github outage caused AppVeyor to think that there was a merge conflict. Github is now correctly showing that there is no merge conflict.

Tools/runtests.ps1
```diff
@@ -23,4 +23,4 @@ if ($selectedVsVersion -eq '') {
 
 $testFiles = $tests | % { ".\BuildOutput\Release$selectedVsVersion\Tests\$_.dll" }
 
-& "$vstest" @args $testFiles
+& "$vstest" /settings:$(".\Build\default." + $selectedVsVersion + "Exp.testsettings") @args $testFiles
```
mousetraps: Just noticed this script won't work if you're running from a different directory. For convenience, let's use something like $PSScriptRoot to establish filepaths.

mousetraps:
:+1: with a suggestion for improvement while you're at it.

Tools/runtests.ps1
mjbvz: Thanks, I didn't know about `PSScriptRoot` and it sounds like this would be a great improvement. I'll update the code to use it.

Issue #831, opened 2016-04-06
VS2015 Update 2 Hangs when opening an node Angular application


AcousticGuitar:
Using this cloned repository: https://github.com/angular-dashboard-framework/angular-dashboard-framework  and attempting to open in VS2015 Community Edition using the "Create New Project from existing Node Application"  VS2015 seems to hang while getting NPM or Bower Packages.  If the application is terminated, and reopened, many of the pacakges are there (which weren't before).  Attempts to finish the job result in losing control of VS2015.

Please advise.

mjbvz:
Can you please go to `Help -> About Visual Studio` to let us know with version of NTVS you are using

I tried to repo this but was not able to.

Also, please try going into `Tools -> Node.js Tools -> Diagnostic Info` and checking `Enable Node.js Tools Live Diagnostics output pane` to see if any interesting messages are printed to the output window.

(#427 may also be relevant here)

mjbvz:
Actually, third attempt I was able to get a freeze using NTVS 1.2 Alpha. Here's the relevant callstack:

```
 	[External Code]	
 	Microsoft.NodejsTools.dll!Microsoft.VisualStudioTools.Project.ProjectNode.GetMsBuildProperty(string propertyName, bool resetCache) Line 5073	C#
 	Microsoft.NodejsTools.dll!Microsoft.VisualStudioTools.Project.ProjectNode.GetProjectProperty(string propertyName, bool resetCache) Line 1720	C#
 	Microsoft.NodejsTools.dll!Microsoft.VisualStudioTools.Project.ProjectNode.GetProjectProperty(string propertyName) Line 3074	C#
 	Microsoft.NodejsTools.dll!Microsoft.NodejsTools.Project.NodejsProjectNode.IsTypeScriptProject.get() Line 224	C#
 	Microsoft.NodejsTools.dll!Microsoft.NodejsTools.Project.NodejsProjectNode.get_ShouldAcquireTypingsAutomatically.AnonymousMethod__18_0() Line 116	C#
 	Microsoft.NodejsTools.dll!Microsoft.VisualStudioTools.UIThread.InvokeAsyncHelper<bool>(System.Func<bool> func, System.Threading.Tasks.TaskCompletionSource<bool> tcs) Line 179	C#
 	Microsoft.NodejsTools.dll!Microsoft.VisualStudioTools.UIThread.InvokeAsync<bool>(System.Func<bool> func) Line 116	C#
[...]

PR #834, opened 2016-04-06
1.2 Alpha Hang on Create From Existing Project

mjbvz:
Fix for #831, VS hanging when creating a new project from existing Node code in 1.2 Alpha

**Bug**
Root cause seems to be `MSBuild.Project.CreateProjectInstance()` hanging when we call into `NodesProjectNode.GetProjectProperty` in a new `IsTypeScriptProject` check.

**Fix**
Two fixes:
* Check `IsProjectOpened` in `IsTypeScriptProject` before attempting to get the project property.
* Reorder the checks in `NodeJsFileNode` to only call into `IsTypeScriptProject ` when we really need to.


Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
```diff
@@ -35,14 +35,16 @@ public NodejsFileNode(NodejsProjectNode root, ProjectElement e)
...
-            if (ProjectMgr == null || ProjectMgr.Analyzer == null || !ProjectMgr.ShouldAcquireTypingsAutomatically) {
+            if (ProjectMgr == null || ProjectMgr.Analyzer == null) {
                 return;
             }
 
             if (Url.EndsWith(NodejsConstants.TypeScriptDeclarationExtension, StringComparison.OrdinalIgnoreCase) && Url.IndexOf(@"\typings\", StringComparison.OrdinalIgnoreCase) >= 0) {
-                ProjectMgr.Site.GetUIThread().Invoke(() => {
-                    this.IncludeInProject(true);
-                });
+                if (ProjectMgr.ShouldAcquireTypingsAutomatically) {
```
mousetraps: It appears that ShouldAcquireTypingsAutomatically is an unnecessarily expensive operation - instead of performing unnecessary string comparisons, we should just cache the value.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -220,7 +220,7 @@ public NodejsProjectNode(NodejsProjectPackage package)
 
         public bool IsTypeScriptProject {
             get {
-                return string.Equals(GetProjectProperty(NodejsConstants.EnableTypeScript), "true", StringComparison.OrdinalIgnoreCase);
+                return IsProjectOpened && string.Equals(GetProjectProperty(NodejsConstants.EnableTypeScript), "true", StringComparison.OrdinalIgnoreCase);
```
mousetraps: Hmm, I think we're still susceptible to race conditions here... why is GetProjectProperty hanging?

Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
mjbvz: This hang happens when the value is first needed, so caching will not fit it.

This reordering is not needed for the fix, but ensures that even if the other part of the change does not fix the hang, we would only run into the hang when creating a project from code with an existing `typings` directory.

The actual operation `IsTypeScriptProject` is not all that expensive when called at the correct time.

mousetraps: I understand it's not needed for this fix in particular, but the fact is that IsTypeScriptProject requires us to be on the UI thread, and we should try and reduce that dependency as much as possible.

mousetraps: Also, we should try to fix the hang here rather than potentially obfuscate its existence.

mousetraps:
bump - any news on this one? Do we want to pull it into Alpha 2?

mjbvz:
Under certain cases, I still see this hang. #907 improved the situation, but this change is still probably worth it.

mousetraps:
But we still have no idea what's actually causing the hang?

mjbvz:
I think we do have an understanding of what changed in our code to cause the hang: after the typings acquision work, we may call `NodesProjectNode.GetProjectProperty` much earlier in the project open flow than we did previously (specifically, this happens in the 'open from existing node code' flow.) When we make this earlier call, `CreateProjectInstance` hangs inside of `GetProjectProperty `. I noticed that the project is not fully loaded in the cases where the hang occurs. 

I can investigate `CreateProjectInstance` further but this PR does not attempt to fix the core problem of `CreateProjectInstance` hanging. It mitigates the problem in the one specific case where the regression was introduced.

mjbvz:
Closing since I haven't been able to reproduce the issue recently.

PR #838, opened 2016-04-07
Fix potential out of range access in ShowSetAsStartupFileCommandOnNode

mjbvz:
I started hitting an out of range exceptions in `ShowSetAsStartupFileCommandOnNode` after enabling `Start web browser on launch` for a short running console app. This fix just makes sure we don't try to access elements of empty arrays in this method.


mousetraps:
:+1:

PR #839, opened 2016-04-07
Update Extension More Info Setting to point to Github

mjbvz:
Updates both the `more info` and `Getting Started` links to point to our GitHub instead of to CodePlex


mousetraps:
Let's create fwlinks for these instead so that we don't have to edit the code in the future.

mjbvz:
Sounds like a good idea, I'll create two new fwd links: one for MoreInfo and one for GettingStarted.

Issue #852, opened 2016-04-12
No tests were discovered when running test under vstest.console.exe


compulim:
When I run it under VS2015 Test Explorer, it works (list tests + run tests). But when running under `vstest.console.exe`, the TestDiscoverer is not able to pickup any tests listed in the `njsproj` file.

`"C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe" /ListTests:mytestproject.njsproj /UseVSIXExtensions:true`

It outputs:

```
Microsoft (R) Test Execution Command Line Tool Version 14.0.25123.0
Copyright (c) Microsoft Corporation.  All rights reserved.

The following Tests are available:
```

It says nothing about the tests. In my other projects, `vstest.console.exe` is able to list out what tests I have in the test case.

Without able to discover any tests in the `njsproj` file, `vstest.console.exe` completed without running any tests. It generates no `TestResults` and outputs:

```
Microsoft (R) Test Execution Command Line Tool Version 14.0.25123.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
```

Tried to debug but `vstest.console.exe` does not has diagnostic log mode, so I can't see the debug log in the CS file.

The reason why I need it to run under `vstest.console.exe` is to able to put Node.js tests under Visual Studio Team Services (a.k.a. Visual Studio Online).

mjbvz:
Thanks for the report.

 I can confirm that running NTVS tests using `vstest.console.exe` does not work as expected, even when explicitly providing some extra hints and verifying that the NTVS test discoverers and test executors are loaded:

```
vstest.console.exe Example.njsproj /TestAdapterPath:"%LOCALAPPDATA%\14.0Exp\Extensions\Microsoft\Node.js Tools\1.2\"  /UseVsixExtensions:true
```

I believe the problem is that `vstest.console.exe` attempts to treat `Example.njsproj` as a dll during loading, which fails with a `BadImageFormatException`. This causes no tests to be found.

### Debugging The Test Adapter
I don't know too much about `vstest.console.exe` and this does not seem like one of our highest priorities at the moment, but anyone is free to take a look and submit a PR with a fix.

Here are some basic instructions for actually debugging the NTVS test adapter for this scenario (on VS 2015 with NTVS 1.2 Alpha):

[...]

PR #855, opened 2016-04-13
Log exception and re-throw

compulim:
Fixing bug #852.

When running under `vstest.console.exe`, exceptions were swallowed. We should log it out and re-throw.

Repro:
`"C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe" MyProject.njsproj`

Expected:
```
Microsoft (R) Test Execution Command Line Tool Version 14.0.25123.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
Error: The imported project "C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v11.0\Node.js Tools\Microsoft.NodejsTools.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.  C:\Users\JohnDoe\Source\Repos\MyProject\MyProject.njsproj
```

Actual:
```
Microsoft (R) Test Execution Command Line Tool Version 14.0.25123.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
```

### DIFF
--- a/Nodejs/Product/TestAdapter/TestDiscoverer.cs
+++ b/Nodejs/Product/TestAdapter/TestDiscoverer.cs
@@ -1,15 +1,15 @@
 ﻿//*********************************************************//
 //    Copyright (c) Microsoft. All rights reserved.
-//    
+//
 //    Apache 2.0 License
-//    
+//
 //    You may obtain a copy of the License at
 //    http://www.apache.org/licenses/LICENSE-2.0
-//    
-//    Unless required by applicable law or agreed to in writing, software 
-//    distributed under the License is distributed on an "AS IS" BASIS, 
-//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or 
-//    implied. See the License for the specific language governing 
+//
+//    Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


Nodejs/Product/TestAdapter/TestDiscoverer.cs
```diff
@@ -89,6 +89,8 @@ class TestDiscoverer : ITestDiscoverer {
                         //Debug.Fail("Before Discover");
                         DiscoverTests(testItems, proj, discoverySink, logger);
                     }
+                } catch (Exception ex) {
+                    logger.SendMessage(TestMessageLevel.Error, ex.Message);
```
mjbvz: I think we should probably rethrow after logging using `throw;` to preserve the existing behavior. Let me know if that causes some sort of problem.

Other than that, the change looks good. Thank you for looking into this.

compulim: Stupid me forget to re-throw. 😛

compulim:
Stupid me. Fixed and squashed into a single commit.

PR #861, opened 2016-04-15
Improve Memory Usage by not holding onto multiple copies of `GlobalPackages`

### LINKED ISSUE #354: Memory consumption of npm nodes is too high for slns w/ many projects
We create a new GlobalModulesNode for every project even though they are the same - we should be able to significantly cut down on memory consumption by ensuring we only create/refresh a single instance of global modules

mjbvz:
This change builds on #815 to further reduce memory usage for solutions with multiple projects.

**Bug**
We currently end up creating `IGlobalPackage` objects for each project. For solutions with multiple projects, if users have many global packages installed (and without #815), this ends up consuming many tens of MBs for no reason.

**Fix**
Added the concept of a `INpmGlobalPackageProvider` which is an object that provides global packages for a given npm path to `NpmController`. `NpmController` is where we currently create the duplicate `IGlobalPackage` objects.

At a higher level, `INpmGlobalPackageProvider` allows us to introduce a `CacheGlobalPackageProivder` object that holds onto `GlobalPackages` objects with a weakmap and can reuse the objects if possible.

Closes #354


Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
```diff
@@ -27,9 +27,35 @@
 using Microsoft.VisualStudioTools.Project;
 using MessageBox = System.Windows.MessageBox;
 using Timer = System.Threading.Timer;
+using System.Threading.Tasks;
```
mousetraps: sort usings (System first)

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
```diff
@@ -27,9 +27,35 @@
...
 
 namespace Microsoft.NodejsTools.Project {
     internal class NodeModulesNode : AbstractNpmNode {
+        /// <summary>
+        /// `INpmGlobalPackageProvider` that caches `IGlobalPackages` objects for reuse.
+        /// </summary>
+        private class CacheGlobalPackageProivder : INpmGlobalPackageProvider {
```
mousetraps: nit: spelling, also move below usage. also keeping 'npm' in the name might make it easier to search for

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
```diff
@@ -27,9 +27,35 @@
...
+            private Dictionary<string, WeakReference<IGlobalPackages>> _packages = new Dictionary<string, WeakReference<IGlobalPackages>>();
```
mousetraps: Not really sure i see a significant enough advantage of using weak references here, especially since a strong reference is already being held onto by the project node and I don't anticipate a) the global folder frequently changing between projects and b) projects being loaded and unloaded often. Am I missing something?

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
```diff
@@ -27,9 +27,35 @@
...
+
+            public async Task<IGlobalPackages> GetGlobalPackages(string pathToNpm) {
+                WeakReference<IGlobalPackages> reference;
+                if (_packages.TryGetValue(pathToNpm, out reference)) {
+                    IGlobalPackages current;
+                    if (reference.TryGetTarget(out current) && current != null) {
+                        return current;
```
mousetraps: nit: the less generic names, the better

Nodejs/Product/Nodejs/Project/NodeModulesNode.cs
```diff
@@ -27,9 +27,35 @@
...
+                    }
+                }
+                var provider = new Npm.SPI.NpmBinGlobalPackageProvider();
```
mousetraps: the additional level of abstraction of a bin global package provider doesn't seem necessary here. there's nothing special about "bin" - that's just how we're getting the location of the global packages.

Nodejs/Product/Npm/SPI/NpmController.cs
```diff
@@ -15,23 +15,31 @@
...
+    /// Standard `INpmGlobalPackageProvider` that uses `npm bin` to get global packages.
+    /// </summary>
+    public class NpmBinGlobalPackageProvider : INpmGlobalPackageProvider {
+        public async Task<IGlobalPackages> GetGlobalPackages(string pathToNpm) {
+            var command = new NpmBinCommand(String.Empty, true, pathToNpm);
+            return (await command.ExecuteAsync())
+                ? RootPackageFactory.Create(command.BinDirectory)
+                : null;
+        }
+    }
```
mousetraps: nit: move below usage

Nodejs/Product/Npm/SPI/NpmController.cs
```diff
@@ -15,23 +15,31 @@
...
 
+    internal class NpmController : AbstractNpmLogSource, INpmController {
         private IPackageCatalog _sRepoCatalog;
         private string _fullPathToRootPackageDirectory;
         private string _cachePath;
         private bool _showMissingDevOptionalSubPackages;
         private INpmPathProvider _npmPathProvider;
-        private IRootPackage _rootPackage;
-        private IGlobalPackages _globalPackage;
-        private readonly object _lock = new object();
```
mousetraps: are you sure these are safe to remove? auto-implemented properties are not thread-safe

Nodejs/Product/Npm/SPI/NpmController.cs
```diff
@@ -41,19 +49,20 @@ internal class NpmController : AbstractNpmLogSource, INpmController {
...
             string cachePath,
             bool showMissingDevOptionalSubPackages = false,
-            INpmPathProvider npmPathProvider = null) {
+            INpmPathProvider npmPathProvider = null,
+            INpmGlobalPackageProvider globalPackageProvider = null) {
             _fullPathToRootPackageDirectory = fullPathToRootPackageDirectory;
             _cachePath = cachePath;
             _showMissingDevOptionalSubPackages = showMissingDevOptionalSubPackages;
             _npmPathProvider = npmPathProvider;
+            _globalPackageProvider = globalPackageProvider ?? new NpmBinGlobalPackageProvider();
```
mousetraps: why not create a CacheGlobalPackageProvider here instead?

Nodejs/Product/Npm/SPI/NpmController.cs
mjbvz: Will revert this. I believe auto properties are threadsafe in that reference get/set are atomic so you can't end up reading a bad pointer, but not threadsafe in that they do not force synchronization of threads, so another thread may end up reading a previous cached value instead of the most recently set one. I'm not convinced we need these locks (and there's a ton of other places that should use them if we did) but keeping them does not hurt anything.

mjbvz:
Closing in favor of #1126

PR #864, opened 2016-04-18
Reduce memory usage and boot time by deferring Analyzer creation for ES6 Preview

### LINKED ISSUE #863: Defer Analyzer Creation Until it is Actually Needed
After #815, the next biggest consumer of memory are the analysis objects. These end up using several tens of MBs currently, even if they are not directly used (such as when analysis is set to `None` or `ES6 preview`). Most of this memory is used to store information about globals and function signatures. Besides memory, creating these objects slows down boot time. 

Really, all the global object data should be created lazily, but that fix is far to complex. So we should instead just not create the analyzer until it is needed. In my experiment creating and opening a new node console project, this saves ~20MB.

mjbvz:
For #863, trying to address memory usage further. The analyzer objects end up consuming around 10mb of data per instance. They also do a lot of processing on construction, slowing down boot time. For solutions with more than one project, both of these factors end up becoming significant.

This change ensures we do not create an analyzer until needed for Intellisense levels `ES6 Preview` and `None`.
 
closes #863


mjbvz:
Right now this messes up the case where you switch Intellisense levels from None/preview to Medium/high. The current behavior relies on even ES6 preview creating the analysis logic behind the scenes, which is transferred over if the intellisense level is switched. With that change, there is no analysis object to transfer from, so the newly created  analyzer ends up throwing all sorts of assertions about not being able to find nodes if you edit a file.

Looking into fixing that.

mjbvz:
I think I fixed the switching problem by defering the JSAnalyzer creation instead of the `VSAnalyzer` creation. Still needs more testing.

Nodejs/Product/Nodejs/Extensions.cs
```diff
@@ -195,7 +195,9 @@ out project
 
             if (pyProj != null) {
                 analyzer = pyProj.Analyzer;
-                return analyzer;
+                if (analyzer != null) {
+                    return analyzer;
+                }
```
mousetraps: In what event don't we want to return null here?

Nodejs/Tests/Core/Debugger/DataTipTests.cs
```diff
@@ -22,6 +22,7 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Microsoft.VisualStudio.TextManager.Interop;
 using TestUtilities.Mocks;
+using Microsoft.NodejsTools.Options;
```
mousetraps: nit: sort usings

mousetraps:
What sort of testing did you do on this change? Mostly concerned about whether we added all the required null checks here.

Nodejs/Product/Nodejs/Extensions.cs
mjbvz: This was needed for the previous logic. Will revert since now only the JsAnalyzer is lazily created instead of the VsProjectAnalyzer.

mjbvz:
I did manual testing of the change around core scenarios and edge cases (such as switching analyzers.) The first iterations of this were much more unstable than those after 43f4420, and I'm not seeing any problems now. Automated tests are passing but that on its own does not give me much confidence that nothing is going to break.

The change is risky because it involves lots of paths, but that's one reason I'd like to get this out into devbuild or alpha. It's an important change for both memory usage and stability, and gets us one step closer to moving off of our homegrown analyzer.

mousetraps:
Great. Also run AnalysisDriver, though it won't give too much coverage on this change in particular. Also, as discussed, let's remove any extraneous null checks as a result of the previous change so that we can understand the full impact and decide if we need to mitigate risk any further.

:+1: after that

PR #866, opened 2016-04-19
Fix for prompting multiple times to include node_modules. 

### LINKED ISSUE #860: Node_modules inclusion in the project asked 3 times
After installing NTVS 1.2 Alpha and opening existing solution with 1 NTVS project + 1 SQL Server project, the NTVS ask to include node_modules 3 times. I always select 'Don't include node_modules (recommended)'

**UPDATE**
Here the image of the dialog which appears
image

mjbvz:
Fixes users getting prompted multiple times to include `node_modules` in their project if an installed npm package includes a `typings` directory.

closes #860

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
@@ -40,7 +40,7 @@ namespace Microsoft.NodejsTools.Project {
             }
 
             if (Url.EndsWith(NodejsConstants.TypeScriptDeclarationExtension, StringComparison.OrdinalIgnoreCase)
-              && Url.StartsWith(Path.Combine(ProjectMgr.ProjectFolder, @"typings"), StringComparison.OrdinalIgnoreCase)) {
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/Project/NodejsFileNode.cs
```diff
@@ -39,7 +39,9 @@ public NodejsFileNode(NodejsProjectNode root, ProjectElement e)
                 return;
             }
 
-            if (Url.EndsWith(NodejsConstants.TypeScriptDeclarationExtension, StringComparison.OrdinalIgnoreCase) && Url.IndexOf(@"\typings\", StringComparison.OrdinalIgnoreCase) >= 0) {
+            if (Url.EndsWith(NodejsConstants.TypeScriptDeclarationExtension, StringComparison.OrdinalIgnoreCase)
+              && Url.IndexOf(@"\typings\", StringComparison.OrdinalIgnoreCase) >= 0
+              && Url.IndexOf(@"\node_modules\", StringComparison.OrdinalIgnoreCase) < 0) {
```
mjbvz: It may be better to check if the url starts with `$PROJECT_ROOT/typings` instead of this approach.

DinoV:
LGTM

Issue #870, opened 2016-04-19
Reload Project leaking `NodejsProjectNode`


mjbvz:
**Repo**
- Create simple new Express 4 app.
- Take memory snapshot.
- Unload the project.
- Reload the project.
- Take another snapshot

**Expected**
Only a single `NodejsProjectNodes` should be loaded since this is a fairly heavy object (not directly, but in what it owns).

**Actual**
Two copies of the `NodejsProjectNode` now exist. Repeat to create more instance.

mjbvz:
One potential cycle:

`NodeProjectNode -> NpmController -> FileSystemWatcher -> NpmController -> NodeProjectNode` where `FileSystemWatcher` is never released.

paulvanbrenk:
This doesn't seem to be a problem anymore

PR #871, opened 2016-04-20
Fixing memory cycles for `NodeModuleNodes` and `NpmControllers` on project unload

mjbvz:
Working on #870.

This change fixes the leaking of `NodeModuleNodes` and `NpmControllers` when NTVS projects are unloaded. There were two types of fix made for the problem:
* File system event handler registrations that were keeping `NodeModuleNodes` alive.
* Manually cutting links in a few places, either using `WeakReference` or by nulling out values or manually removing nodes from a hierarchy.

The second type of fix is more of a hack until we can figure out why `NodejsProjectNode` is leaking.

### DIFF
--- a/Nodejs/Product/Npm/SPI/NpmController.cs
+++ b/Nodejs/Product/Npm/SPI/NpmController.cs
@@ -33,8 +33,8 @@ namespace Microsoft.NodejsTools.Npm.SPI {
         private IGlobalPackages _globalPackage;
         private readonly object _lock = new object();
 
-        private FileSystemWatcher _localWatcher;
-        private FileSystemWatcher _globalWatcher;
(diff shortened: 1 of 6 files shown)


Nodejs/Product/Npm/SPI/NpmController.cs
```diff
@@ -33,8 +33,8 @@ internal class NpmController : AbstractNpmLogSource, INpmController {
         private IGlobalPackages _globalPackage;
         private readonly object _lock = new object();
 
-        private readonly FileSystemWatcher _localWatcher;
-        private readonly FileSystemWatcher _globalWatcher;
+        private FileSystemWatcher _localWatcher;
```
mjbvz: I think I can revert these changes now, they were more to help with debugging to leak.

PR #876, opened 2016-04-20
Run Repl code in strict mode

### LINKED ISSUE #487: Does 'use strict' work? Not seeing any effect in interpreter window.
To reproduce in NTVS interpreter window:

  Install (1) node.js 4.x.x (2) NTVS.Dev.2015-09-17.VS.2015.msi. Run 2015 and open NTVS interpreter window,

     > 'use strict';
    'use strict'
    > let A = [1, 3];
    SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode

mjbvz:
Issue #487

**Bug**
Repl currently cannot use some es6 features such as block scoping becaues it is not in strict mode.

**Fix**
I think we probably want to run all repl code in strict mode. Switch to use vm instead of eval and make sure all statements are in strict mode.

Also made `visualstudio_nodejs_repl.js` use strict.

Closes #487

### DIFF
--- a/Nodejs/Product/Nodejs/visualstudio_nodejs_repl.js
+++ b/Nodejs/Product/Nodejs/visualstudio_nodejs_repl.js
@@ -13,6 +13,7 @@
 //    permissions and limitations under the License.
 //
 //*********************************************************//
+"use strict";
 
 if (process.argv.length <= 2) {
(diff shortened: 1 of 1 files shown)


mousetraps:
:+1:

PR #879, opened 2016-04-21
Set correct main in AzureWebWorker 

### LINKED ISSUE #619: Azure Cloud Service with Node sets Incorrect Startup file
1. Create a new Azure Cloud Service
2. Add a Node.js Worker Role
3. Notice that package.json specifies main as "server.js" - it should be "worker.js"
4. Notice that  ServiceDefinition.csdef specifies \server.js" as the entry point - it should be \worker.js":

<ProgramEntryPoint commandLine="node.cmd .\server.js" setReadyOnProcessStart="true" />

mjbvz:
Use `worker` as `main` instead of `server` for AzureWebWorker template.

closes #619


mousetraps:
:+1:

PR #880, opened 2016-04-22
Fix dispose pattern in a few more cases

mjbvz:
Using the VS static analyzer to find objects that are not being disposed properly and fixing some of these.

Related to #870. Still trying to track down who exactly is holding onto the NodeJsProject node after it is unloaded

### DIFF
--- a/Nodejs/Product/Nodejs/Intellisense/BufferParser.cs
+++ b/Nodejs/Product/Nodejs/Intellisense/BufferParser.cs
@@ -16,21 +16,15 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
-using System.IO;
-using System.IO.Compression;
 using System.Linq;
 using System.Threading;
-using System.Windows;
-using System.Windows.Threading;
 using Microsoft.NodejsTools.Analysis;
 using Microsoft.VisualStudio.Text;
-using Microsoft.VisualStudio.Text.Editor;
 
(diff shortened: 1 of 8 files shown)


Nodejs/Product/Nodejs/Intellisense/BufferParser.cs
```diff
@@ -51,18 +45,6 @@ class BufferParser {
...
-                    buffer.ChangedLowPriority -= BufferChangedLowPriority;
-                    buffer.Properties.RemoveProperty(typeof(BufferParser));
-                    if (_document != null) {
-                        _document.EncodingChanged -= EncodingChanged;
-                        _document = null;
-                    }
-                }
-                _timer.Dispose();
-            }
-
```
mjbvz: This method was just never used

Nodejs/Product/Nodejs/Intellisense/VsProjectAnalyzer.cs
```diff
@@ -1327,9 +1327,16 @@ CollectingErrorSink errorSink
             }
 
             if (_analysisQueue != null) {
-                _analysisQueue.Stop();
+                _analysisQueue.Dispose();
```
mjbvz: This calls stop, and we want to ensure other held objects are released as well.

mjbvz:
Closing in favor of #881

PR #881, opened 2016-04-22
Stop Leaking NodeJsFolderNodes

mjbvz:
Branch of #880 to continue addressing #870 

The NodeJsProjectNode is still being leaked here, but this change cuts its connections to NodeJsFileNode so that these are no longer leaked, reducing the amount of leakage about 75% (more on more complex projects).

This is accomplished by:
* Implementing dispose patterns with #880
* Using WeakRefs in a few cases.
* Nulling out property fields on Dispose

Really, the last two are more hacks and could be avoided if we could release NodeJsProjectNode, but the reference graph around these objects is a big ball of craziness. Manually cutting the connections seems like the correct thing to do at this stage.


mousetraps:
Too bad we haven't been able to trace this down much further, but for the sake of readability and mergeability w/ common, let's replace weak refs with something cleaner.

mjbvz:
Done. We just null out or clear those in the destructor now.

mousetraps:
:+1:

PR #888, opened 2016-04-25
Fix an argument exception in `OnTestContainersChanged`

mjbvz:
**Bug**
When unloading a project with the Test Explorer window open, an exception is thrown in `OnTestContainersChanged` because `TryGetProjectPath` returns null

**Fix**
Add two additional checks to handle the case where the project has been unloaded but the test explorer is still referencing it.

Also auto corrected a delegate invocation in the body of the function.


mousetraps:
:+1:

mousetraps:
Also, in the future, can you please file issues as well so it's easier to track when we put out release notes?

PR #889, opened 2016-04-25
Preserve stack when rethrowing exceptions in C# files

mjbvz:
Two small changes to make sure we preserve the stack when catching and rethrowing an exception in our c# code.


mousetraps:
:+1:

Issue #854, opened 2016-04-13
OOM Exception after project load


nearwood:
See #820. I updated to VS2015 Update 2. Now I eventually receive a OOM error after opening the project, even if I just open a project and let VS sit for ~5 mins.

`System.OutOfMemoryException was unhandled
Message: An unhandled exception of type 'System.OutOfMemoryException' occurred in Microsoft.NodejsTools.Analysis.dll`

Event log:

```
Log Name:      Application
Source:        .NET Runtime
Date:          04/13/16 09:47:32
Event ID:      1026
Task Category: None
Level:         Error
Keywords:      Classic
User:          N/A
Computer:      xxxx
Description:
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.OutOfMemoryException
   at Microsoft.NodejsTools.Analysis.AnalysisSetDetails.AnalysisHashSet.EnsureSize(Int32)
   at Microsoft.NodejsTools.Analysis.AnalysisSetDetails.AnalysisHashSet.CheckGrow()
   at Microsoft.NodejsTools.Analysis.AnalysisSetDetails.AnalysisHashSet.AddOne(Microsoft.NodejsTools.Analysis.AnalysisProxy)
   at Microsoft.NodejsTools.Analysis.AnalysisSetDetails.AnalysisHashSet.Union(System.Collections.Generic.IEnumerable`1<Microsoft.NodejsTools.Analysis.AnalysisProxy>, Boolean ByRef)
   at Microsoft.NodejsTools.Analysis.AnalysisSet.Create(System.Collections.Generic.IEnumerable`1<Microsoft.NodejsTools.Analysis.AnalysisProxy>)
   at Microsoft.NodejsTools.Analysis.Analyzer.DependentKeyValue.AddTypes(Microsoft.NodejsTools.Analysis.ProjectEntry, Microsoft.NodejsTools.Analysis.JsAnalyzer, Microsoft.NodejsTools.Analysis.IAnalysisSet, Microsoft.NodejsTools.Analysis.IAnalysisSet, Boolean)
   at Microsoft.NodejsTools.Analysis.Values.ExpandoValue.SetIndex(Microsoft.NodejsTools.Parsing.Node, Microsoft.NodejsTools.Analysis.AnalysisUnit, Microsoft.NodejsTools.Analysis.IAnalysisSet, Microsoft.NodejsTools.Analysis.IAnalysisSet)
[...]

PR #890, opened 2016-04-26
#53 Add environment setup script to streamline build instructions

### LINKED ISSUE #53: Streamline NTVS build instructions
There are a few things that could be automated using a powershell script.
http://nodejstools.codeplex.com/wikipage?title=Build%20Instructions%20for%20NTVS

mousetraps:
Also...
* refactors vs target check 
* moves runtests to root level (reserving Tools for required 3rd party dependencies)
* minor fixes to runtests to apply to more env configurations

close #53


EnvironmentSetup.ps1
```diff
@@ -0,0 +1,91 @@
...
+    $targetVSTestHostVersion = "$($version.number).1.0"
+    if ((-not $currentVSTestHost) -or ($currentVSTestHost.VersionInfo.FileVersion -ne $targetVSTestHostVersion)) {
+        Write-Warning "VSTestHost already installed. Overriding VSTestHost version $($currentVSTestHost.VersionInfo.FileVersion) with target VSTestHostVersion $targetVSTestHostVersion"
+        $shouldInstallVSTestHost = $true
+        break
+    }
+}
+
+if ($shouldInstallVSTestHost) {
+    Start-Process msiexec -ArgumentList "/i $(Join-Path $rootDir $vsTestHostLocation)"
```
mjbvz: Side note: the vstesthost.msi seems to be causing problems on the most recent win10 builds (see #854.) I've run into this problem as well when setting up new VMs.

Do you know if there is an updated msi we can use instead?

mjbvz:
Looks great! This should make getting started as a contributor much easier. Merging in now.

Make sure to update the wiki as well: https://github.com/Microsoft/nodejstools/wiki/Build-Instructions

Issue #891, opened 2016-04-26
Have issues with building from source


kant2002:
I try to build latest NTVS to try fixes, but having problem building by instructions.

I try to run regular build by running
```
msbuild NodejsTools.sln /p:Configuration=Release /p:VsTarget=14.0 /p:VisualStudioVersion=14.0
```

This setup is give me following error:
```
error CS1762 : Warning as error : A reference was created to embedded interop assembly 'c:\Program Files (x86)\Microsoft.NET\Primary Interop Assemblies\stdole.dll' because of an indirect reference to that as
sembly created by assembly 'c:\Program Files (x86)\Microsoft Visual Studio 15.0\Common7\IDE\PublicAssemblies\Microsoft.VisualStudio.CommandBars.dll'. Consider changing the 'Embed Interop Types' property on eit
her assembly. [<dir>\nodejstools\Nodejs\Product\InteractiveWindow\InteractiveWindow.csproj]
```

When building from VS that not an issue since project does not set WarningAsErrors setting.

kant2002:
For the the fix was change property in the [InteractiveWindow.csproj:73](https://github.com/Microsoft/nodejstools/blob/master/Nodejs/Product/InteractiveWindow/InteractiveWindow.csproj#L73) and  [InteractiveWindow.csproj:76](https://github.com/Microsoft/nodejstools/blob/master/Nodejs/Product/InteractiveWindow/InteractiveWindow.csproj#L76) from `True` to `False`. But that 's override settings introduced in the @8757515299cd6090502dfb4eb4706182eb716f64

After that change I was able to build the NTVS, but not sure that this is 100% correct fix.

PR #893, opened 2016-04-26
Fix issues with building from source

kant2002:
This remove warning which prevent to perform build of MSI, See #891

### DIFF
--- a/Nodejs/Product/InteractiveWindow/InteractiveWindow.csproj
+++ b/Nodejs/Product/InteractiveWindow/InteractiveWindow.csproj
@@ -70,10 +70,10 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="envdte, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
-      <EmbedInteropTypes>True</EmbedInteropTypes>
+      <EmbedInteropTypes>False</EmbedInteropTypes>
     </Reference>
     <Reference Include="stdole, Version=7.0.3300.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
-      <EmbedInteropTypes>True</EmbedInteropTypes>
+      <EmbedInteropTypes>False</EmbedInteropTypes>
     </Reference>
     <Reference Include="Microsoft.VisualStudio.CommandBars, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
       <EmbedInteropTypes>False</EmbedInteropTypes>


mjbvz:
@mousetraps Can you please take a look at this one.

mousetraps:
This change has a good chance of breaking installations of NTVS on machines that do not have the VS SDK installed because required interop types are no longer being embedded. Instead, I believe the proper and less risky fix is to embed interop types on the Microsoft.VisualStudio.CommandBars reference instead. Could you try this and see how it goes?

Additionally, I can't reproduce this behavior at all. Could you specify some more details about your machine configuration? VS version / update #, etc? Which VS command prompt are you using? And presumably you're building from master?

kant2002:
@mousetraps 
1. I build from master.
2. I have installed VS2013, VS2015 and VS15 Preview.
3. I use Git Shell form GitHub for Windows (which is basically Powershell + some adjustments)

I have to make MSbuild from VS2015 to path, in order to make a build.

If I understand correctly somehow references from VSN15 (Next) picked up instead of VS2015.

I run build like that:
`msbuild NodejsTools.sln /p:Configuration=Release /p:VsTarget=14.0 /p:VisualStudioVersion=14.0`

What I notice is just now is that in each project envdte used from VS15 folder, not VS2015.
In the project reference properties
Path = `C:\Program Files (x86)\Microsoft Visual Studio 15.0\Common7\IDE\PublicAssemblies\envdte.dll`
Instead of 

Path = `C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\PublicAssemblies\envdte.dll`

That's I see in MSBuild logs and inside VS 2015.

mousetraps:
Okay, thanks for the additional info. I'll see if I can reproduce your environment when i get the chance - also did you try the fix suggested above?

kant2002:
@mousetraps I try instead of that fix, just change `EmbedInteropTypes` to `True` for Microsoft.VisualStudio.CommandBars, but still no luck.

```
error CS1762 : Warning as error : A reference was created to embedded interop assembly 
'c:\Program Files (x86)\Microsoft Visual Studio 15.0\Common7\IDE\PublicAssemblies\envdte.dll' 
because of an indirect reference to that assembly created by assembly 
'c:\Program Files (x86)\Microsoft Visual Studio 14.0\VSSDK\VisualStudioIntegration\Common\Assemblies\v4.0\Microsoft.VisualStudio.Shell.14.0.dll'. 
Consider changing the 'Embed Interop Types' property on either assembly.
```

kant2002:
Sorry for bothering, I do full clean and build working.

PR #899, opened 2016-04-26
Fix Debugger display of JS strings that contain Date strings

### LINKED ISSUE #106: String var containing time, not displayed correctly.

mjbvz:
Addresses #106.

**Bug**
Json.net will automatically try to parse strings that look like date strings. If one of these is recieved as the value in a Node debug message, we end up displaying the wrong value in the debugger.

**Fix**
Make sure we always use DateTimeParse.None when dealing with messages from the debugger. If we do need to parse a date string, we should handle those cases explicitly.

closes #106


mousetraps:
:+1:

PR #900, opened 2016-04-27
Fix some errors in runtests

mjbvz:
#890 introduced some regressions with `runtests.ps1`. A few small fixes:

* Use updated variable in place of `selectedVsVersion` to select test settings file.
* Set `nodejstoolsroot ` to root of NTVS
* Forward args to `vstest.console.exe`

### DIFF
--- a/RunTests.ps1
+++ b/RunTests.ps1
@@ -21,8 +21,7 @@
 [CmdletBinding()]
 param(
     [string[]] $vstarget,
-	[Parameter(Position=0)][string[]] $args
-
+    [Parameter(Position=0)][string[]] $args
 )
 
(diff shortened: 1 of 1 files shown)


RunTests.ps1
```diff
@@ -42,11 +43,13 @@ if (-not $target_versions) {
     throw "Could not find any valid versions of Visual Studio installed"
 }
 
+$nodejstoolsroot = $PSScriptRoot
```
mousetraps: oops, looks like we're duplicating logic - either use $rootDir or $nodejstoolsroot

mousetraps:
minor comment and :+1: after that!

PR #901, opened 2016-04-27
Remove "Test" prefix from test method names

mjbvz:
About half of our tests use the prefix `Test` for their method names. Test methods are already marked as tests using the `TestMethod` attribute. Also naming the test methods `Test*` makes finding tests in the test explorer more difficult

Quick find and replace to eliminate the `Test` prefixes. This gives all of our test methods a consistent naming convetion.


mousetraps:
nit: for the edge cases we could use *Test as a the suffix rather than prefix so that it's still easy to search for.

mousetraps:
:+1:

PR #902, opened 2016-04-27
Fix two Ignored Problematic json tests

mjbvz:
Fixes two ignored `package.json` tests. Both tests should throw exceptions.


mousetraps:
Let's just remove these two tests. In v4.5 of newtonsoft.json they were testing that parsing worked. The fact that they throw an exception now isn't particularly significant since we're not going to take any action if they stop throwing an exception (other than changing the test back again :wink:)

mjbvz:
Done

Issue #906, opened 2016-04-29
Loading existing project extremely slow


kant2002:
I try to load existing TypeScript project using NTVS built from master, and it load over 5 minutes on my project. Maybe this is related to #831. but for me cause was in the slow calls to `NodejsProjectNode.ShouldAcquireTypingsAutomatically`.

mjbvz:
Fix by #907

PR #907, opened 2016-04-29
Speedup calls to ShouldAcquireTypingsAutomatically by caching values.

kant2002:
Related to #906

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
@@ -49,9 +49,9 @@ namespace Microsoft.NodejsTools.Project {
         internal readonly RequireCompletionCache _requireCompletionCache = new RequireCompletionCache();
         private string _intermediateOutputPath;
         private readonly Dictionary<NodejsProjectImageName, int> _imageIndexFromNameDictionary = new Dictionary<NodejsProjectImageName, int>();
-        private bool? shouldAcquireTypingsAutomatically;
 
 #if DEV14
+        private bool? shouldAcquireTypingsAutomatically;
         private TypingsAcquisition _typingsAcquirer;
 #endif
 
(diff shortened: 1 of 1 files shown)


kant2002:
@mjbvz and @mousetraps Could you review that, since this is is significantly speedup loading time for me.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -49,6 +49,7 @@ class NodejsProjectNode : CommonProjectNode, VsWebSite.VSWebSite, INodePackageMo
         internal readonly RequireCompletionCache _requireCompletionCache = new RequireCompletionCache();
         private string _intermediateOutputPath;
         private readonly Dictionary<NodejsProjectImageName, int> _imageIndexFromNameDictionary = new Dictionary<NodejsProjectImageName, int>();
+        private bool? shouldAcquireTypingsAutomatically;
```
mjbvz: Guard this with a `#if DEV14`. Otherwise the compiler will complain about the unused variable when compiling on Dev15.

mjbvz:
The change looks good overall, just adding a few comments.

The original reason for not caching the value is that `IsTypeScriptProject` may be set after a project has already been created (for example, when you add a TypeScriptFile to a previously Javascript only project). Your fix looks like the right one, but just to handle that case, in this file, find where `SetProjectProperty(NodejsConstants.EnableTypeScript, true)` is called and manually set `shouldAcquireTypingsAutomatically` to `false` there: https://github.com/kant2002/nodejstools/blob/cache-ShouldAcquireTypingsAutomatically/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs#L311

After that, we'll merge this in and get the fix out in the next alpha.

Thanks for looking into this.

kant2002:
@mjbvz Address the feedback, and comment about `SetProjectProperty(NodejsConstants.EnableTypeScript, true)`

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -304,6 +310,10 @@ public NodejsProjectNode(NodejsProjectPackage package)
                 // enable TypeScript on the project automatically...
                 SetProjectProperty(NodejsConstants.EnableTypeScript, "true");
                 SetProjectProperty(NodejsConstants.TypeScriptSourceMap, "true");
+#if DEV14
+                // Reset cached value, so it will be recalculated later.
+                this.shouldAcquireTypingsAutomatically = null;
```
mjbvz: Because we know that `EnableTypeScript` is true, we should just set this to false here.

After than, looks good to me.

kant2002:
@mjbvz done.

mjbvz:
Thanks for following up on this. The fix will be in the next 1.2 alpha

Issue #909, opened 2016-04-29
IndexOutOfRangeException when starting solution with 3 Node.js projects


nearwood:
Hello again,

```
Microsoft Visual Studio Professional 2015
Version 14.0.25123.00 Update 2
Microsoft .NET Framework
Version 4.6.01055
...
JavaScript Language Service   2.0
JavaScript Project System   2.0
Node.js Tools   1.2.40329.01
Node.js Tools - Profiling   1.2.40329.01
TypeScript   1.8.30.0
VSTestHost   1.0.1
```

I was running 1.2 Alpha and ran into a System.IndexOutOfRangeException:

```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.IndexOutOfRangeException
   at Microsoft.NodejsTools.SourceMapping.SourceMap..ctor(System.IO.TextReader)
   at Microsoft.NodejsTools.SourceMapping.SourceMapper.TryGetMapInfo(System.String)
   at Microsoft.NodejsTools.SourceMapping.SourceMapper.MapToOriginal(System.String)
   at Microsoft.NodejsTools.SourceMapping.SourceMapper.GetOriginalFileName(System.String, System.Nullable`1<Int32>, System.Nullable`1<Int32>)
   at Microsoft.NodejsTools.Debugger.NodeDebugger.GetOrAddModule(Microsoft.NodejsTools.Debugger.NodeModule, Microsoft.NodejsTools.Debugger.NodeModule ByRef, Microsoft.NodejsTools.Debugger.NodeStackFrame)
   at Microsoft.NodejsTools.Debugger.NodeDebugger.AddModules(System.Collections.Generic.IEnumerable`1<Microsoft.NodejsTools.Debugger.NodeModule>)
   at Microsoft.NodejsTools.Debugger.NodeDebugger.OnCompileScriptEvent(System.Object, Microsoft.NodejsTools.Debugger.Communication.CompileScriptEventArgs)
   at Microsoft.NodejsTools.Debugger.Communication.DebuggerClient.HandleEventMessage(Newtonsoft.Json.Linq.JObject)
   at Microsoft.NodejsTools.Debugger.Communication.DebuggerClient.OnOutputMessage(System.Object, Microsoft.NodejsTools.Debugger.Communication.MessageEventArgs)
[...]

PR #910, opened 2016-04-29
#909 Fix OutOfRangeException

nearwood:
Change the way index bounds are checked for source map segments in SourceMap.cs.

Before:
```
if (info.Length >= SourcesIndex ) {
  sourceIndex += info[SourcesIndex];
}
```

This throws when, for example, `info.Length == 1 && SourcesIndex == 1`.

Now:
```
if (SourcesIndex < info.Length) {
  sourceIndex += info[SourcesIndex];
}
```

The same logic was applied to the other 3 index checks.


mjbvz:
Thanks, this looks like the correct fix and the code is easier to read now too.

PR #914, opened 2016-05-03
Fixes for Debugger when attaching to exited process

### LINKED ISSUE #912: Crash When Quickly Closing Console Window Before Debugging
**Repo**
1. Start with blank JavaScript Node console project
2. Run using debugger
3. Quickly close the console window. Not sure of the exact timing but it has to be almost right after the console appears.

**Expected**
Debugger exits

**Actual**
Assertion thrown, crashes VS.

The exact cause of the crash is `AD7Engine::Send` with a failed assertion on the line `EngineUtils.RequireOk(eventObject.GetAttributes(out attributes));` Clearly, things are not ok.

mjbvz:
#912 

**Bug**
We can currently crash or hang when attaching to a node process that exits quickly.

**Fixes**
This change adds a few fixes:

* In `Send`, don't treat RPC failure hresults as true failures that should throw exceptions.
* Fix infinite hang by timing out correctly.
* Handle  `InvalidOperationException` when trying to access `TcpClient.Stream`

closes #912

### DIFF
--- a/Nodejs/Product/Nodejs/Debugger/Communication/DebuggerConnection.cs
+++ b/Nodejs/Product/Nodejs/Debugger/Communication/DebuggerConnection.cs
@@ -295,8 +295,8 @@ namespace Microsoft.NodejsTools.Debugger.Communication {
                 }
             } catch (SocketException) {
             } catch (IOException) {
-            } catch (InvalidOperationException) {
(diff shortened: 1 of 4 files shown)


mousetraps:
:+1: but let's also file an issue to revisit so that we don't have to debug Aggregate exceptions

PR #918, opened 2016-05-03
Fix Repl window null deref when main NTVS package has not been loaded yet

### LINKED ISSUE #917: Null Dereference when Using Interactive Window Before Main NodeJs Package is loaded
**Repo**

1. Launch VS (restored state should have Node.js interactive window open already)
1. Quickly type something into the interactive window
1. Press enter.

**Expected**
Repl stuff

**Actual**
Null deref in GetStartupProject:

```
>	Microsoft.NodejsTools.dll!Microsoft.VisualStudioTools.CommonPackage.GetStartupProject(System.IServiceProvider serviceProvider) Line 178	C#
 	Microsoft.NodejsTools.dll!Microsoft.NodejsTools.Repl.VsNodejsReplSite.GetStartupProject() Line 26	C#
 	Microsoft.NodejsTools.dll!Microsoft.NodejsTools.Repl.NodejsReplEvaluator.GetNodeExePath() Line 228	C#
 	Microsoft.NodejsTools.dll!Microsoft.NodejsTools.Repl.NodejsReplEvaluator.Connect() Line 180	C#
 	Microsoft.NodejsTools.dll!Microsoft.NodejsTools.Repl.NodejsReplEvaluator.EnsureConnected() Line 169	C#
 	Microsoft.NodejsTools.dll!Microsoft.NodejsTools.Repl.NodejsReplEvaluator.ExecuteText(string text) Line 135	C#
 	Microsoft.NodejsTools.InteractiveWindow.dll!Microsoft.NodejsTools.Repl.ReplWindow.ExecuteActiveCode.AnonymousMethod__196_0() Line 2330	C#
 	Microsoft.NodejsTools.InteractiveWindow.dll!Microsoft.NodejsTools.Repl.ReplWindow.UIThread(System.Action action) Line 3272	C#
 	Microsoft.NodejsTools.InteractiveWindow.dll!Microsoft.NodejsTools.Repl.ReplWindow.ExecuteActiveCode() Line 2299	C#
 	Microsoft.NodejsTools.InteractiveWindow.dll!Microsoft.NodejsTools.Repl.ReplWindow.Submit() Line 1943	C#
 	Microsoft.NodejsTools.InteractiveWindow.dll!Microsoft.NodejsTools.Repl.ReplWindow.PreEditorCommandFilterExec(ref System.Guid pguidCmdGroup, uint nCmdID, uint nCmdexecopt, System.IntPtr pvaIn, System.IntPtr pvaOut) Line 1567	C#
 	Microsoft.NodejsTools.InteractiveWindow.dll!Microsoft.NodejsTools.Repl.ReplWindow.CommandFilter.Exec(ref System.Guid pguidCmdGroup, uint nCmdID, uint nCmdexecopt, System.IntPtr pvaIn, System.IntPtr pvaOut) Line 1344	C#
```

The root cause is that `NodejsPackage.Instance` is null at this point.

This crashes the interactive window and requires a VS restart.

mjbvz:
#917

**Bug**
If the Repl window is used before the main NTVS package has been loaded, `NodeJsPacakge.Instance` will be null. This causes a null deref problem in some call paths (see #917)

**Fix**
Add a few null checks to these call paths.

closes #917

### DIFF
--- a/Common/Product/SharedProject/CommonPackage.cs
+++ b/Common/Product/SharedProject/CommonPackage.cs
@@ -130,6 +130,7 @@ namespace Microsoft.VisualStudioTools {
         /// <returns></returns>
         public static IWpfTextView GetActiveTextView(System.IServiceProvider serviceProvider) {
             if (serviceProvider == null) {
+                Debug.Assert(false, "No service provider");
                 return null;
             }
 
@@ -179,6 +180,7 @@ namespace Microsoft.VisualStudioTools {
 
         internal static CommonProjectNode GetStartupProject(System.IServiceProvider serviceProvider) {
             if (serviceProvider == null) {
+                Debug.Assert(false, "No service provider");
                 return null;
             }
             var buildMgr = (IVsSolutionBuildManager)serviceProvider.GetService(typeof(IVsSolutionBuildManager));
--- a/Common/Product/SharedProject/CommonPackage.cs
+++ b/Common/Product/SharedProject/CommonPackage.cs
@@ -129,6 +129,10 @@ namespace Microsoft.VisualStudioTools {
         /// </summary>
         /// <returns></returns>
         public static IWpfTextView GetActiveTextView(System.IServiceProvider serviceProvider) {
+            if (serviceProvider == null) {
+                return null;
+            }
+
             var monitorSelection = (IVsMonitorSelection)serviceProvider.GetService(typeof(SVsShellMonitorSelection));
             if (monitorSelection == null) {
                 return null;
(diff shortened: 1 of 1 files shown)


mjbvz:
Actually, the change should not hurt anything, but it also does not solve a normal use case bug.

During development, I was messing around with the Repl window and this ended up persisting some bad data about the Repl Window to the registry under the `ActiveRepls` key. This seemed to have messed up loading, causing this bug to be hit. In normal user flows, I think that `Node.Instance` should always be set at these points in the code.

Common/Product/SharedProject/CommonPackage.cs
```diff
@@ -129,6 +129,10 @@ public abstract class CommonPackage : Package, IOleComponent {
         /// </summary>
         /// <returns></returns>
         public static IWpfTextView GetActiveTextView(System.IServiceProvider serviceProvider) {
+            if (serviceProvider == null) {
+                return null;
+            }
```
mousetraps: for both null checks, it's okay to fail gracefully, but let's go ahead and also add a debug assert so we know something's up when the issue occurs.

mousetraps:
minor comment and :+1: after that

PR #919, opened 2016-05-04
Unregister options page callback in NodejsProjectNode on destruction

mjbvz:
**Bug**
`NodeJsProjectNode` registers a callback for `GeneralOptionsPage::ShowBrowserAndNodeLabelsChanged` but never unregisters.

**Fix**
Unregister this event in the destructor of `NodeJsProjectNode`


mousetraps:
:+1:

PR #920, opened 2016-05-04
Remove Unused Functions and Fields in NodeJs Code

mjbvz:
Just a quick pass through the code to remove unused private functions and fields identified by the analyzer.

Removing these helps us better understand internal dependencies and helps make the code more maintainable.


Nodejs/Product/Nodejs/CodeWindowManager.cs
```diff
@@ -20,10 +20,14 @@
 
 namespace Microsoft.NodejsTools {
     class CodeWindowManager : IVsCodeWindowManager, IVsCodeWindowEvents {
+#if FALSE
         private readonly IVsEditorAdaptersFactoryService _adapterService;
-
```
mjbvz: This property is only referenced in `#if FALSE` blocks. Rather than remove those as part of this change, I just move this field into the block as well. A separate, more targeted change can remove all the `#if FALSE` blocks.

mousetraps:
:+1:

PR #925, opened 2016-05-05
Fixing some NodeJsProjectNode refCount Issues

mjbvz:
Port of Microsoft/PTVS#1262 to try to address core memory leak issues:

* Marshal incrementing refcount and not decrementing it in two places.
* Automation object dispose.

In my testing, this did not actually fix the leak but does fix a few real issues with the current code. #881 still helps out with this.

### DIFF
--- a/Common/Product/SharedProject/ProjectNode.cs
+++ b/Common/Product/SharedProject/ProjectNode.cs
@@ -1601,6 +1601,12 @@ namespace Microsoft.VisualStudioTools.Project {
                 taskProvider.Tasks.Clear();
             }
 
+            var autoObject = GetAutomationObject() as Automation.OAProject;
+            if (autoObject != null) {
+                autoObject.Dispose();
+            }
+            this.configProvider = null;
+
             try {
                 // Walk the tree and close all nodes.
                 // This has to be done before the project closes, since we want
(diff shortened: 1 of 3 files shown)


Common/Product/SharedProject/ProjectNode.cs
```diff
@@ -2790,9 +2796,14 @@ internal enum EventTriggering {
...
-            EnvDTE.Project automationObject = GetAutomationObject() as EnvDTE.Project;
+            var solutionBuild = (IVsSolutionBuildManager)GetService(typeof(SVsSolutionBuildManager));
+            IVsProjectCfg[] cfg = new IVsProjectCfg[1];
+            ErrorHandler.ThrowOnFailure(
+                solutionBuild.FindActiveProjectCfg(IntPtr.Zero, IntPtr.Zero, GetOuterHierarchy(), cfg));
 
-            SetConfiguration(Utilities.GetActiveConfigurationName(automationObject));
+            string name;
+            ErrorHandler.ThrowOnFailure(cfg[0].get_CanonicalName(out name));
+            SetConfiguration(name);
```
mousetraps: could you ensure this matches exactly the PR on the python side (spacing etc.) so that we resolve conflicts more easily when we sync w/ common?

mousetraps:
:+1: - you mentioned it doesn't fully solve the issue... is this the case even after GC (as per dino's comment)?

mjbvz:
@mousetraps Yes, I could be overlooking something but I've reloaded the projects multiple times and waited a little while, and still the the number of NodeJsProjects leaking.

Common/Product/SharedProject/ProjectNode.cs
mjbvz: Ok. The new content should be copy paste but I'll make sure the surrounding stuff matches too.

PR #927, opened 2016-05-05
Delete DropDownBarClient

mjbvz:
The `DropDownBarClient` code is all `#if FALSE` commented out and slowly accumulating errors. If we need to restore this at any point, it's in source control.

#924 Deletes the only other commented out references to this class.


mousetraps:
:+1:

PR #942, opened 2016-05-12
Update README.md links to use aka.ms links

mjbvz:
Just updating links to use aka.ms links so we don't have to update these through github for new releases.


mjbvz:
@mousetraps I'm not seeing http://aka.ms/ntvslateststable go to the correct redirect yet. Was this the correct link?

mjbvz:
Oh, I see that we actually took http://aka.ms/lateststable . Let's change that to be **ntvslateststable** for consistency.

mousetraps:
:+1:  and just changed the aka.ms link

Issue #869, opened 2016-04-19
Performance - Creating Hierarchy Nodes for every file in `node_modules`


mjbvz:
We currently create `FolderNodes` for every file in `node_modules` during project initialization. For projects with many installed packages, this results in extra start up time and extra memory usage since several thousand extra nodes end up being created.

I think that the `FolderNodes` are actually required if we want to show the contents of `node_modules` (even if the folder is excluded from the actual project). However, if possible, we should create the node_modules folder child nodes lazily when the parent is expanded.

mjbvz:
Just noting down some thoughts on this.

The easiest sub problem to solve first seems to be the NPM `DependencyNode`. We end up creating between several hundred to several thousand of these nodes for no real reason during project loading, which consumes memory but mostly just slows down project initialization and slows down NPM module update operations. I'm not sure how helpful browsing the entire NPM dependency tree really is, but let's assume we want to keep this behavior.

Here are my thoughts on making the children lazy instead:

1. `DependenyNode` inherits from `HierarchyNode`. Currently, `ReploadHeirachy` is what populates the `DependenyNode` children. It would be more clear to have  `DependenyNode` know how to populate its own children. Move the `ReploadHeirachy` logic to `DependenyNode` and make `ReploadHeirachy` call into this to preserve the existing behavior.

2. In `HierarchyNode`, add a overridable method to determine if a node should be expandable. Expandability is currently based on if the node has any children. But in the case of `DependencyNode`, we know that children should exists from `Package`, but do not actually want to create the children before they are needed. Add this base method to `HierachyNode` and override it on `DependenyNode` using the `Package` information.

3. In `HierarchyNode`, add a way to hook into an `OnExpand` type event that is invoked when the node is expanded  (either by the user or programmatically) or is loaded in an already expanded state. 

4. In `DependenyNode`, use `OnExpand` to populate the children for the first time or update its children. This would just invoke the old `ReploadHeirachy` logic moved in step 1, but not recursively like `ReploadHeirachy ` currently does. 

5. Remove the old calls to `ReploadHeirachy` introduced in step 1.

6. 🌈

paulvanbrenk:
Looks like the issue is that the scanning of the file system happens on the UI thread... which for obvious reasons is terrible. Further we don't need any of these node until the uses enables show all files... so going to change this to move the scanning to a low-priority bg thread.

paulvanbrenk:
[...]

PR #940, opened 2016-05-12
Mocha path

ericpizon:
-supports configuring mocha path thru mocha.json. When mocha path is configured, it will override the default, which is in project directory.

PR #944, opened 2016-05-18
Limit Depth of created dependecy nodes

mjbvz:
**Bug**
Right now, we end up parsing the entire node dependency tree and creating objects to represent every dependency in it. This uses a lot of memory and a lot of time during boot. Here's a simple, new Express 4 app for example:

image

The plus side of this extra computation is that we allow users to explore the entire module dependency tree in VS directly.

**Fix**
All these operations are negatively impacting user experience, especially when loading projects with lots of dependencies (such as the eslint source.) I also am not convinced that browsing the dependencies of your dependencies is actually useful.

This fix limits the depth to which we create dependency nodes to 1 (only root dependencies are shown). Memory usage and boot time are much improved:

image

The change does remove some existing behavior, so if anyone feels strongly about this, we can discuss this further. The alternative is something like #869.

### DIFF
--- a/Nodejs/Product/Npm/SPI/NodeModules.cs
+++ b/Nodejs/Product/Npm/SPI/NodeModules.cs
@@ -22,13 +22,18 @@ using System.Linq;
 
 namespace Microsoft.NodejsTools.Npm.SPI {
     internal class NodeModules : AbstractNodeModules {
-        private Dictionary<string, ModuleInfo> _allModules;
         private static readonly string[] _ignoredDirectories = { @"\.bin", @"\.staging" };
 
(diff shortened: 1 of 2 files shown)


mousetraps:
Let's discuss ways to mitigate the performance issue, but cutting the feature altogether is a little premature (especially after events like left-pad, which has put people on edge about their project composition) and also will create inconsistency with other project types. Additionally, this will impact the Node language service, which currently depends on these depth calculations.

mjbvz:
Holding off on this change because it may impact the analyzer is fine, but we should not keep this feature just because removing it reduce functionality. Instead, do people find this feature helpful and does the functionality fit with our longer term vision? Opening large Node projects is currently a terrible experience, and this fix makes things significant better. 

Beyond all the various leaks and other memory reduction fixes already in 1.2 Alpha 2, there is no good story to mitigate this problem. #940 could work, but that's a risky change and may re-introduce problems similar to #815.

PR #967, opened 2016-05-24
Fix potential leak/null reference when switching projects

mjbvz:
**Bug**
We can end up hitting a null dereference or leaking a node when switching projects for a project with a symlink watcher. This happens on `CommonProjectNode` because we null out the `ItemNode` before destroying the symlink watcher.

**Fix**
Reorder the operations. We should destory the symlink watcher and then null out the nodes.


mousetraps:
:+1:

PR #968, opened 2016-05-25
Fix hang when sending debug commands after Node program runs to completion

### LINKED ISSUE #935: Hang on Remove Breakpoint after program completes
**Repo**

1. Create new empty app with main file with one line:
```js
1. console.log('Hello world!');
```

2. Set breakpoint on that line.
3. Run app
4. Continue on breakpoint
5. App stops after at last line of program but is still being debugged.
6. Click to remove breakpoint on line 1

**Expected**
Breakpoint is removed

**Actual**
Infinite hang.

On Node 5.x, this operation is canceled out after a few seconds and we can continue on, but with Node 6.1.0 we hang forever. The hang is on this line: https://github.com/Microsoft/nodejstools/blob/master/Nodejs/Product/Nodejs/Debugger/DebugEngine/AD7BoundBreakpoint.cs#L52

mjbvz:
Issue #935

##### Bug
Run a node program to completion. We pause on the last line before exiting. Node is still running at this point. Running a debug command at this point (such as pause or remove breakpoint) can freeze VS since we never receive responses back from the Node debugger.

##### Fix
This change has two fixes:

* Add an explicit timeout when removing breakpoints. We have an identical timeout when adding breakpoints so it probably makes sense to add this.
* On our low level send calls, register the cancellation token to cancel the promise.

A better fix would be just to never send requests to node after we run to the last line, but it is not clear how we can detect this without a wrapper. The debug protocol does not seem to emit any events about this and Node is still running.

Also deleted a bunch of unused NodejsTools extension functions while I was in the file.

closes #935

### DIFF
--- a/Nodejs/Product/Nodejs/Debugger/Extensions.cs
+++ b/Nodejs/Product/Nodejs/Debugger/Extensions.cs
@@ -15,163 +15,11 @@
 //*********************************************************//
 
 using System;
-using System.Diagnostics;
-using System.IO;
-using System.Net.Sockets;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.NodejsTools.Debugger {
     static class Extensions {
-        /// <summary>
-        /// Reads a string from the socket which is encoded as:
-        ///     U, byte count, bytes 
-        ///     A, byte count, ASCII
-        ///     
-        /// Which supports either UTF-8 or ASCII strings.
-        /// </summary>
-        internal static string ReadString(this Socket socket) {
-            byte[] cmd_buffer = new byte[4];
-            if (socket.Receive(cmd_buffer, 1, SocketFlags.None) == 1) {
-                if (cmd_buffer[0] == 'N') {
-                    // null string
-                    return null;
-                }
-                bool isUnicode = cmd_buffer[0] == 'U';
-
-                if (socket.Receive(cmd_buffer) == 4) {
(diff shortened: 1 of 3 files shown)


Nodejs/Product/Nodejs/Debugger/Extensions.cs
```diff
@@ -182,18 +30,12 @@ static class Extensions {
...
-            int count = await streamReader.ReadBlockAsync(buffer, 0, length).ConfigureAwait(false);
-            if (count == 0) {
-                var errorMessage = string.Format("Unable to read {0} bytes from stream.", length);
-                throw new InvalidDataException(errorMessage);
-            }
-
-            // Get UTF-8 string
-            byte[] bytes = streamReader.CurrentEncoding.GetBytes(buffer, 0, count);
-            return Encoding.UTF8.GetString(bytes);
+        internal static async Task WaitAsync(this Task task, TimeSpan timeout, CancellationToken token = default(CancellationToken)) {
```
mjbvz: Unfortunately C# doesn't understand the signature:

```cs
internal static async T WaitAsync<T>(this T task, TimeSpan timeout, CancellationToken token = default(CancellationToken)) where T : Task {
    ...
}
```
Which would have allowed overloading for both `Task` and `Task<T>`. Let me know if there a better way to do this so we don't have to duplicate this code.

mousetraps:
:+1:

Issue #974, opened 2016-06-01
Double clicking on a Mocha test opens JavaScript source file in external application


jcansdale:
##### Expected Behavior 

Double clicking on test in Test Explorer should open source file in Visual Studio.

##### Actual Behavior 

Double clicking on Mocha test opens JavaScript source file in external application (Notepad2 in my case).

- NTVS Version: 1.2 Alpha
- Visual Studio Version: 2015
- Node.js Version: v6.1.0

##### Steps to Reproduce
1. Create new Mocha test
2. Double click on test in Test Explorer
![test_explorer](https://cloud.githubusercontent.com/assets/11719160/15723581/fba72164-283a-11e6-97fa-b42779d3ebe2.png)

mjbvz:
Merged in #979 and #982 that should resolve this. Let me know if that is not the case.

Thanks.

PR #976, opened 2016-06-01
Fix for double clicking on Mocha tests opens external editor.

jcansdale:
Issue   #974

##### Bug

Jumping to line: 0, column: 0 causes JavaScript file to open in external editor.

##### Fix

Changed default to line: 1, column: 1.


jcansdale:
It looks like I did a pull request on my 'master' branch, not realizing this would be a moving target (rather than a snapshot). I'll reissue the pull request on a separate branch that fixes just this issue. I'm familiar with subversion but git is new to me. Thanks for your patience. ;)

PR #978, opened 2016-06-03
Fix for Issue #974. Double clicking on Mocha tests opens external editor.

jcansdale:
Issue #974

##### Bug

Double clicking in Mocha test causes JavaScript file to open in external editor.

##### Fix

Jumping to line: 0, column: 0 seems to cause this. Changed default to line: 1, column: 1.

### DIFF
--- a/Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
+++ b/Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
@@ -20,8 +20,8 @@ var find_tests = function (testFileList, discoverResultFile, projectFolder) {
                         test: t.fullTitle(),
                         suite: suite.fullTitle(),
                         file: testFile,
-                        line: 0,
-                        column: 0
+                        line: 1,
+                        column: 1
                     });
                 });
             }


mjbvz:
Thanks for reporting this issue and finding a fix so quickly. The change looks good overall.

One nitpicky question, should it be `line 1, col 0` instead `line 1, col 1`? I don't think the current solution will cause any problem in most cases, but make sure to test opening a blank file this way to see if that does the correct thing as well.

jcansdale:
> One nitpicky question, should it be line 1, col 0 instead line 1, col 1?

Good question! It looks like 'NodejsTestInfo.SourceColumn' where it gets stored isn't actually used anywhere. Visual Studio uses 1 based for lines and columns. The v8 debugger and TypeScript package on NPM uses 0 based for both. I guess we can assume it should be consistent.

Maybe code on the JavaScript side should work with 0 based lines/columns and only code on the C#/Visual Studio side should work with 1 based? Otherwise I can see this issue creeping in again and again when new frameworks are added.

jcansdale:
This PR is obsolete if you choose to accept PR #982 instead (which would be my preference).

mjbvz:
Merged in #982

Issue #977, opened 2016-06-02
exportrunner.js / find_tests throws ReferenceError if global.dbg is not defined


jcansdale:
I've been writing some unit tests for exportrunner.js and have come across a number of issues.

1) If the 'dbg' global variable is not defined, the find_tests function throws "ReferenceError: dbg is not defined" rather than falling back to using default line and column numbers.

2) The default line and column numbers are 0 rather than 1. This would cause issues when the test was opened/navigated to.

3) The function is returning 0 based column numbers rather than 1 based as the editor expects.

4) Requiring Node to be started with the "--expose_debug_as=dbg" pollutes the global namespace and makes it a little more tricky to test. The debug context can be alternatively accessed via the following:
var debug = vm.runInDebugContext('Debug');

Here's a snippet that shows the v8 debugger using 0 based column numbers (so 1 needs to be added):

var column0Func = vm.runInNewContext("(function func\n(a,b,c) {})");
var debug = vm.runInDebugContext('Debug');
var column0FuncDetails = debug.findFunctionSourceLocation(column0Func);
assert.equal(column0FuncDetails.column, 0, "columns are 0 based");

Here is the pertinent code:
https://github.com/Microsoft/nodejstools/blob/89007991be6d08f844ce6465a4aa3a110bff5116/Nodejs/Product/Nodejs/TestFrameworks/ExportRunner/exportrunner.js#L18

- NTVS Version: 1.2 Alpha

None of these cause serious issues, but I noticed them and thought I'd mention it. I'll issue a pull request later. :)

PR #979, opened 2016-06-03
Fixed "ReferenceError: dbg is not defined" error when 'dbg' global va…

jcansdale:
Issue #977

##### Bug

1) If the 'dbg' global variable is not defined, the find_tests function throws "ReferenceError: dbg is not defined" rather than falling back to using default line and column numbers.

2) The default line and column numbers are 0 rather than 1. This would cause issues when the test was opened/navigated to (open .js file in external editor, see #974).

3) The function is returning 0 based column numbers rather than 1 based as the editor expects.

4) Requiring Node to be started with the "--expose_debug_as=dbg" pollutes the global namespace and makes it a little more tricky to test. The debug context can be alternatively accessed via the following:
var debug = vm.runInDebugContext('Debug');

##### Fix

Changed to use vm.runInDebugContext('Debug') rather than 'dbg' global variable.
Return 1 based column and line numbers.
No need to start Node process with "--expose_debug_as=dbg" .

### DIFF
--- a/Nodejs/Product/Nodejs/TestFrameworks/ExportRunner/exportrunner.js
+++ b/Nodejs/Product/Nodejs/TestFrameworks/ExportRunner/exportrunner.js
@@ -1,7 +1,15 @@
 var fs = require('fs');
 var path = require('path');
+var vm = require('vm');
 
 var find_tests = function (testFileList, discoverResultFile) {
+    var debug;
+    try {
+        debug = vm.runInDebugContext('Debug');
+    } catch (ex) {
+        console.error("NTVS_ERROR:", ex);
+    }
+
     var testList = [];
     testFileList.split(';').forEach(function (testFile) {
         var testCases;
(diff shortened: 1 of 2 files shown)


mjbvz:
Thanks for finding all these issues and fixing them too! I've merged in the PR. The fix should be available in our next release.

Thanks again.

jcansdale:
No problem. Thanks for accepting my first pull request. :)

BTW, are there any JavaScript (or preferably TypeScript) tests that run as part of this project? I'd like to submit test tests as well.

mjbvz:
@jcansdale Good question. Unfortunately we do not have this sort of test currently.

The NTVS tests could be improved a lot, both in terms of quality and coverage. This is something that we really need to look into more, in order to improve our checkin confidence and speed. A subset of our C# tests are run with AppVeyor, and you can manually run additional tests to validate changes on your local machine: https://github.com/Microsoft/nodejstools/wiki/Running-Our-Tests The current NTVS tests mainly focus on Visual studio integration, and many are UI based or have dependencies on things like the file system 😢

It's far too much work to go back and fix all out our legacy tests in a single pass, so I think that introducing new tests that cover a small area well is the correct approach to improve the NTVS testing situation. I'd be happy to provide pointers or assistance to anyone interested in looking into this.

PR #981, opened 2016-06-06
ES6 IntelliSense UX improvements

mousetraps:
After auto-dts acquisition, the following InfoBar will appear in solution explorer to help educate the user and increase discoverability of the feature.
image

Clicking on the "typings folder" link will open a web page that includes more information about what's going on under the covers. 

Clicking on the "Customize..." button will open the relevant options page, which has been reconstructed to be conditional upon the selected IntelliSense mode. ES6 will use the new Salsa language service, and ES5 will use the old Node.js static analysis engine. 

**ES6 Mode**
image

**ES5 Mode**
image

**Additional Considerations**
* InfoBar design
  * Consistent with VS UI guidelines 
  * We do not want to block the user’s flow during project creation, so we’ve gone with an in-context infobar with an option to customize the experience rather than asking for confirmation. Anecdotally, most people are okay with this experience, and the action is also easy enough to undo (simply delete the folder) 
  * Also considered adding a “Don’t show again” button to the info bar, but it’s a small space and cluttered things up a bit much. 
  * It would be annoying for the infobar to show up every time typings are downloaded, so it will only show up when the folder is newly created. In every other case, it just prints out to the output window / status bar.  
  * The message chosen for the info bar was a balance between informativeness and length (to minimize chance of wrapping over to line #3)
  * Also considered infobar in editor rather than solution explorer, and the upside is that we wouldn’t be as limited with length of text, but the downside is that it can be quite jarring because the primary focus is on the editor, and also made things appear quite cluttered. 
  * Selected NewFolder glyp because it best conveys the thing that the user cares about – which is that a new folder was added to their project.
* Options design
  * When ES5 is selected, a warning will appear. Right now the warning appears below the combo box to inform the user that it’s been deprecated. The horizontal line and fixed height serve to make the reflow less jarring when a different IntelliSense mode is selected. 
  * Quick IntelliSense will be selected by default when moving from ES6->ES5 mode (rather than full intellisense). All other options are retained when switching between modes.
  * running tsd with --save is now *off* by default so that we don't unnecessarily clutter up the user's project. 

**What's missing?**
* currently all the links point to the same page - we should create fwlinks for each, so that we have a little more flexibility
* still need to update docs
* was intending to synchronize formatting options between TS and Node.js, but it looks like that's not nearly as simple as anticipated, and we may need to reconsider our approach here.


Nodejs/Product/Nodejs/Options/NodeLsIntellisenseOptionsControl.cs
```diff
@@ -0,0 +1,135 @@
...
+                        break;
+                }
+            }
+        }
+
+        internal int AnalysisLogMaximum {
+            get {
+                int max;
+                // The Max Value is described by 'Max' instead of 'Int32.MaxValue'
+                if (_analysisLogMax.Text == "Max") {
```
mjbvz: Why?

[nit] Also, extract `Max` to a constant if it is going to be used.

mjbvz: Ok, see this was moved from other file. Fine leaving it as is.

mjbvz:
Where any considerations made for typings acquisition being disabled in VS15? I'm fine merging this in before the VS15 experience is perfect since we don't have a release of that scheduled soon, but we should track this as a future work item. 

Also, should `#if DEV14` guards be used anywhere to explicitly prevent new logic or classes from being used in dev15?

Nodejs/Product/Nodejs/Options/TypingsInfoBar.cs
```diff
@@ -0,0 +1,134 @@
...
+                if (_instance == null) {
+                    lock (_infoBarLock) {
+                        if (_instance == null) {
+                            _instance = new TypingsInfoBar(NodejsPackage.Instance);
+                        }
+                    }
+                }
+                return _instance;
+            }
+        }
```
mjbvz: [nit] Prefer `Lazy` to encapsule this pattern.

Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -137,7 +137,11 @@ internal class TypingsAcquisition {
         }
 
         private static IEnumerable<string> TsdInstallArguments(IEnumerable<string> packages) {
-            return new[] { "install", }.Concat(packages).Concat(new[] { "--save" });
+            var arguments = new[] { "install", }.Concat(packages);
+            if (NodejsPackage.Instance.IntellisenseOptionsPage.SaveChangesToConfigFile) {
+                arguments = arguments.Concat(new[] { "--save" });
```
mjbvz: [super-nit] just return here instead of the assignment.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -145,10 +147,19 @@ public NodejsProjectNode(NodejsProjectPackage package)
         }
 
         private void TryToAcquireTypings(IEnumerable<string> packages) {
+            bool isNewTypingsFolder = !Directory.Exists(Path.Combine(this.ProjectHome, "typings"));
```
mjbvz: Shouldn't we also notify users in the case when we update an existing `typings` folder? That seems like more consistent behavior to me since it also results in an unexpected change to your local files.

Nodejs/Product/Nodejs/Options/SalsaLsIntellisenseOptionsControl.cs
```diff
@@ -0,0 +1,42 @@
...
+            page.SaveChangesToConfigFile = _saveChangesToConfigFile.Checked;
+        }
+
+        internal void SyncControlWithPageSettings(NodejsIntellisenseOptionsPage page) {
+            _enableAutomaticTypingsAcquisition.Checked = page.EnableAutomaticTypingsAcquisition;
+            _showTypingsInfoBar.Checked = page.ShowTypingsInfoBar;
+            _saveChangesToConfigFile.Checked = page.SaveChangesToConfigFile;
+        }
+
+        private void typingsLearnMoreLink_LinkClicked(object sender, LinkLabelLinkClickedEventArgs e) {
```
mjbvz: I don't see any references to `typingsLearnMore` but am not too familiar with WinForms. Is this function used?

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -145,10 +147,19 @@ public NodejsProjectNode(NodejsProjectPackage package)
...
             if (ShouldAcquireTypingsAutomatically && TypingsAcquirer != null) {
                 TypingsAcquirer
                     .AcquireTypings(packages, null /*redirector*/)
-                    .ContinueWith(x => x);
+                    .ContinueWith(x => {
+                        if (NodejsPackage.Instance.IntellisenseOptionsPage.ShowTypingsInfoBar &&
```
mjbvz: For the future: if the acquisition fails, do we also want to show a message of some kind? Typing acquisition failures will result in an unexpectedly degraded intellisense experience.

mjbvz:
The change looks great overall and should help users a lot.

I added a few nit comments and had a few questions/notes, but there were no major problems in the code. Let's just make sure we are on the same page regarding the VS15 and open any new work items for that before checking this in. We can create another WI for updating the documentation too, since that step shouldn't block this check in but must be done before the 1.2 release.

Also, if you are still looking into synchronizing the TS and Node formatting, please do that in another PR so we can keep this one more focused.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
mousetraps: See reasoning above. This is by-design. In the majority of cases:
a) users will not have an existing typings folder and will only care that a new folder has been added to the project
b) users that have a typings folder in a javascript project will be okay with stuff being downloaded to it

And even the small subset of users that already have a typings folder, but are not okay with stuff being downloaded to it will be educated enough to discover the option.

As mentioned earlier, a suitable next step here is to add an option to use tsd or typings, but that can come in a future PR.

Nodejs/Product/Nodejs/Options/SalsaLsIntellisenseOptionsControl.cs
mousetraps: GitHub collapsed a lot of code. It's being used in the *Designer.cs file to back the "Learn More about typings" link.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
mousetraps: Yep, that's a great idea - at least for when internet connection is lacking or something like that. If we do it lack of availability, i'd be concerned with how frequently it will fail to acquire typings for a newly installed package. We should chat with @bowdenk7, who's been delving into Salsa UX in general.

PR #987, opened 2016-06-08
Reorder TryDeactivateSymLinkWatcher in call to prevent leak/crash

mjbvz:
This is the same fix as #967, reordering function calls to make sure we deactive the symlinkwatcher before destroying the FolderNode. The current order results in an error when deleting folders from a project.

### DIFF
--- a/Common/Product/SharedProject/HierarchyNode.cs
+++ b/Common/Product/SharedProject/HierarchyNode.cs
@@ -1567,7 +1567,6 @@ namespace Microsoft.VisualStudioTools.Project {
             firstChild = null;
             lastChild = null;
             nextSibling = null;
-            itemNode = null;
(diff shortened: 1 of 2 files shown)


Common/Product/SharedProject/HierarchyNode.cs
```diff
@@ -1567,7 +1567,6 @@ internal abstract class HierarchyNode :
             firstChild = null;
             lastChild = null;
             nextSibling = null;
-            itemNode = null;
```
mjbvz: This change was added in order to address our memory leaks #871 but is no longer required. For 1.2, it may cause additional problems so I think reverting this line is the correct solution

mousetraps:
:+1:

PR #993, opened 2016-06-08
Gate 1.2 Beta installer to VS 2015 update 3 or later

### LINKED ISSUE #990: Gate 1.2 Beta  installer to VS 2015 update 3 or later
NTVS 1.2 Beta will require VS 2015 Update 3 or later in order to pick up some important Typescript fixes.

We should gate the installer to this version, similar to what we did for the 1.2 Alpha and gating to VS 2015 update 2 or later: #806 

The beta may actually have to be gated to the Update 3 RC, while the final release should be gated to Update 3 final

mjbvz:
This prevents installing NTVS 1.2 Beta on anything before VS 2015 Update 3 RC. For the final 1.2 release, we will want to change the gate to target the final version of Update 3 instead. This is being tracked by #994 

closes #990

### DIFF
--- a/Common/Setup/LaunchConditions.wxs
+++ b/Common/Setup/LaunchConditions.wxs
@@ -74,10 +74,10 @@
   </Fragment>
 
   <Fragment>
-    <Property Id="VS_IS_2015_PRE_UPDATE_2" Secure="yes">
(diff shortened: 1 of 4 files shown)


mousetraps:
The fixes that we're waiting on are not until RTM, so no need to gate just yet.

PR #995, opened 2016-06-08
Synchronize Node.js <-> TS settings

mousetraps:
Synchronize Node.js <-> TS settings so that Node.js options are applied to Salsa. In particular this change:

* on startup, populate Node.js options with TS -> Node.js editor settings (general, scrollbars, and tabs)
* on Node.js editor settings changed, populate TS options
* use TS properties as backing for Node.js formatting settings

limitations:
* On TS options changed, will not sync back to Node.js
* On TS + Node.js options changed, will use Node.js options

### DIFF
--- a/Nodejs/Product/Nodejs/LanguagePreferences.cs
+++ b/Nodejs/Product/Nodejs/LanguagePreferences.cs
@@ -14,14 +14,15 @@
 //
 //*********************************************************//
 
+using System;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.TextManager.Interop;
 
 namespace Microsoft.NodejsTools {
-    class LanguagePreferences : IVsTextManagerEvents2 {
-        LANGPREFERENCES _preferences;
+    class LanguagePreferences : IVsTextManagerEvents4 {
+        internal LANGPREFERENCES3 _preferences;
 
-        public LanguagePreferences(LANGPREFERENCES preferences) {
+        public LanguagePreferences(LANGPREFERENCES3 preferences) {
             _preferences = preferences;
         }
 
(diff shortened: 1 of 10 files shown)


Nodejs/Product/Nodejs/LanguagePreferences.cs
```diff
@@ -47,12 +48,18 @@ class LanguagePreferences : IVsTextManagerEvents2 {
...
         }
 
-        public int OnUserPreferencesChanged2(VIEWPREFERENCES2[] viewPrefs, FRAMEPREFERENCES2[] framePrefs, LANGPREFERENCES2[] langPrefs, FONTCOLORPREFERENCES2[] colorPrefs) {
-            if (langPrefs != null && langPrefs.Length > 0 && langPrefs[0].guidLang == this._preferences.guidLang) {
-                _preferences.IndentStyle = langPrefs[0].IndentStyle;
-                _preferences.fAutoListMembers = langPrefs[0].fAutoListMembers;
-                _preferences.fAutoListParams = langPrefs[0].fAutoListParams;
-                _preferences.fHideAdvancedAutoListMembers = langPrefs[0].fHideAdvancedAutoListMembers;
+        public int OnUserPreferencesChanged4(VIEWPREFERENCES3[] pViewPrefs, LANGPREFERENCES3[] pLangPrefs, FONTCOLORPREFERENCES2[] pColorPrefs) {
+            IVsTextManager4 textMgr = (IVsTextManager4)NodejsPackage.Instance.GetService(typeof(SVsTextManager));
```
mjbvz: Can `textMgr` ever be null?

Nodejs/Product/Nodejs/Options/NodejsFormattingDialogPage.cs
```diff
@@ -0,0 +1,60 @@
...
+//*********************************************************//
+
+using System;
+using System.Runtime.InteropServices;
+using Microsoft.VisualStudio.Settings;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.NodejsTools.Options {
+    public class NodejsFormattingDialogPage : NodejsDialogPage {
+        ISettingsManager _settingsManager;
```
mjbvz: [nit] `private readonly`?

mjbvz:
Look good overall. Let's get the CI passing and make sure to test this on a machine without the VS sdk installed and then merge this in for the 1.2 beta.

Nodejs/Product/Nodejs/LanguagePreferences.cs
mousetraps: We don't null check anywhere, and I wouldn't expect it to ever be null (except maybe inside of some of the mock test cases?)

PR #996, opened 2016-06-08
Fix Interactive window .npm commands sometimes failing in VS 2015 update 3

mjbvz:
##### Bug
If `dteProject.FullName` is empty when we execute an `.npm` command in the interactive window, an exception is thrown by `Path.GetDirectoryName(dteProject.FullName)` and we end up not doing anything. I believe this is caused by the typescript virtual project or something similar showing up in the list of projects that we iterate through.

##### fix
Check if `dteProject.FullName` is empty or null before calling `GetDirectoryName`


mousetraps:
:+1:

PR #1000, opened 2016-06-08
Remove empty email field from template package.json

mjbvz:
**Bug**
Our templates currently generate invalid `package.json` (according to the VS `package.json` schema checker) because the `email` field of `author` is always empty. This has been annoying me for a while.

**Fix**
Just remove the email field from the template entirely since it is optional.


mjbvz:
Here's to the next #1000. Candy for everyone 🍬 🍫 🍭

mousetraps:
:+1:

PR #1015, opened 2016-06-10
Fix smart indent settings crash

### LINKED ISSUE #1014: Invalid Cast Exception in Interactive Window in NTVS 1.2 Beta
##### Steps to Reproduce
1. Open VS 2015
2. Open Node.js interactive window
3. Type:  `a.` in the interactive window
4. Put cursor on the `$` in `a$.`
5. Press return 

##### Expected Behavior 
We get a nice smart indent on the next line.

##### Actual Behavior 
💥 with `InvalidCast` exception on `SmartIndent.cs` in the `GetDesiredIndentation` function on the line where we call `dte.get_properties("TextEditor", "Node.js")`. `No such interface supported`.

- NTVS Version: NTVS 1.2 Beta
- Visual Studio Version: VS 2015 Update 3

mjbvz:
Issue #1014

##### Bug
Crash in `SmartIndent.cs` in the interactive window. I believe that the root cause is #995, which changed how we store our settings.

##### Fix
Switch to using the NodeJs language preferences from typescript instead. This ensures we are using the same logic to get the language settings as we use in other places.

closes #1014

### DIFF
--- a/Nodejs/Product/Nodejs/NodejsPackage.cs
+++ b/Nodejs/Product/Nodejs/NodejsPackage.cs
@@ -251,6 +251,11 @@ namespace Microsoft.NodejsTools {
             Environment.SetEnvironmentVariable(NodejsConstants.NodeToolsProcessIdEnvironmentVariable, Process.GetCurrentProcess().Id.ToString());
         }
 
+        public static LANGPREFERENCES3[] GetNodejsLanguagePreferencesFromTypeScript() {
+            IVsTextManager4 textMgr = (IVsTextManager4)Instance.GetService(typeof(SVsTextManager));
+            return GetNodejsLanguagePreferencesFromTypeScript(textMgr);
+        }
+
         private static LANGPREFERENCES3[] GetNodejsLanguagePreferencesFromTypeScript(IVsTextManager4 textMgr) {
             var langPrefs = new LANGPREFERENCES3[1];
             langPrefs[0].guidLang = Guids.TypeScriptLanguageInfo;
--- a/Nodejs/Product/Nodejs/NodejsPackage.cs
+++ b/Nodejs/Product/Nodejs/NodejsPackage.cs
@@ -251,12 +251,7 @@ namespace Microsoft.NodejsTools {
             Environment.SetEnvironmentVariable(NodejsConstants.NodeToolsProcessIdEnvironmentVariable, Process.GetCurrentProcess().Id.ToString());
         }
 
-        public static LANGPREFERENCES3[] GetNodejsLanguagePreferencesFromTypeScript() {
-            IVsTextManager4 textMgr = (IVsTextManager4)Instance.GetService(typeof(SVsTextManager));
-            return GetNodejsLanguagePreferencesFromTypeScript(textMgr);
-        }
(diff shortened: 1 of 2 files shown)


Nodejs/Product/Nodejs/SmartIndent.cs
```diff
@@ -46,19 +48,17 @@ sealed class SmartIndent : ISmartIndent {
         #region ISmartIndent Members
 
         public int? GetDesiredIndentation(VisualStudio.Text.ITextSnapshotLine line) {
-            var dte = (EnvDTE.DTE)NodejsPackage.GetGlobalService(typeof(EnvDTE.DTE));
-
-            var props = dte.get_Properties("TextEditor", "Node.js");
-            switch ((EnvDTE._vsIndentStyle)(int)props.Item("IndentStyle").Value) {
+            var langPrefs = NodejsPackage.GetNodejsLanguagePreferencesFromTypeScript();
+            switch ((EnvDTE._vsIndentStyle)langPrefs[0].IndentStyle) {
```
mjbvz: I'll see if I can get rid of this gross casting now that we are not directly using EnvDTE.

Nodejs/Product/Nodejs/SmartIndent.cs
```diff
@@ -23,6 +23,8 @@
 using Microsoft.VisualStudio.Text.Editor.OptionsExtensionMethods;
 using Microsoft.VisualStudio.Text.Tagging;
 using Microsoft.VisualStudio.Utilities;
+using Microsoft.VisualStudio.TextManager.Interop;
```
mjbvz: Neither of these is required now, will remove.

Nodejs/Product/Nodejs/NodejsPackage.cs
```diff
@@ -251,6 +251,11 @@ internal sealed partial class NodejsPackage : CommonPackage {
             Environment.SetEnvironmentVariable(NodejsConstants.NodeToolsProcessIdEnvironmentVariable, Process.GetCurrentProcess().Id.ToString());
         }
 
+        public static LANGPREFERENCES3[] GetNodejsLanguagePreferencesFromTypeScript() {
+            IVsTextManager4 textMgr = (IVsTextManager4)Instance.GetService(typeof(SVsTextManager));
+            return GetNodejsLanguagePreferencesFromTypeScript(textMgr);
+        }
```
mousetraps: This is unnecessary - we should be able to use NodejsPackage.Instance.LangPrefs to access this property.

Nodejs/Product/Nodejs/SmartIndent.cs
```diff
@@ -22,7 +22,7 @@
 using Microsoft.VisualStudio.Text.Editor;
 using Microsoft.VisualStudio.Text.Editor.OptionsExtensionMethods;
 using Microsoft.VisualStudio.Text.Tagging;
-using Microsoft.VisualStudio.Utilities;
+using EnvDTE;
```
mousetraps: sort usings

mousetraps:
minor comments and :+1: after that

PR #1025, opened 2016-06-13
Download typings on es5 -> Es6 switch

### LINKED ISSUE #1021: Switching from ES5->ES6 Mode should attempt to download typings
##### Expected Behavior 
IntelliSense is available after switching from ES5-> ES6 Mode

##### Actual Behavior 
IntelliSense is not available because typings haven't been downloaded.

mjbvz:
##### Bug
Typings not downloaded on es5 -> Es6 intellisense switch in NTVS 1.2

##### Fix
Hook into the intellisense change event to download them.

Closes #1021

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
@@ -147,24 +147,30 @@ namespace Microsoft.NodejsTools.Project {
         }
 
         private void TryToAcquireTypings(IEnumerable<string> packages) {
-            bool isNewTypingsFolder = !Directory.Exists(Path.Combine(this.ProjectHome, "typings"));
-            if (ShouldAcquireTypingsAutomatically && TypingsAcquirer != null) {
-                TypingsAcquirer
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -591,6 +581,7 @@ public NodejsProjectNode(NodejsProjectPackage package)
                 }
             }
             _analyzer = analyzer;
+            TryToAcquireCurrentTypings();
```
mjbvz: This function is internally gated to only download the typings if IntelliSense is ES6 and a number of other conditions are met.

mousetraps: I know it's not part of this change, but since the operation gets called enough times (OnIdleNodeModules), we should change the internal gating of TryToAcquireTypings to check for ShouldAcquireTypings immediately rather than after it's determined the npm packages to install. Additionally, while you're at it... let's move the isNewTypingsFolder bool into the if condition in TryToAcquireTypings

mousetraps:
small comment and :+1: after that

Issue #1022, opened 2016-06-12
Why does 6-11-2016 Dev Build for VS15 require VS2015?


alanpurple:
Why need VS2015?
Which was not true for 1.2 alpha2

alanpurple:
and yet, even after I installed VS2015 community with updates, still same message blocked me from installing

mjbvz:
Does the message mention something about upgrading to update 2 of VS 2015? We changed the gating recently but this could have broken VS15: #1009

mousetraps:
This should be fixed in the next dev build.

PR #1009, opened 2016-06-09
Restore gating for Pre Update 2 for NTVS 1.2

mjbvz:
Adds back in gating removed by #1005. This change instead uses the PTVS gating logic.

Tested on RTM machine and on saw gating, plus tested on update 2 machine and saw that it was not gated.

closes #1003

PR #1026, opened 2016-06-13
Move all of VS_IS_2015_PRE_UPDATE_2 into VS14 guard

mjbvz:
Issue #1022

**Bug**
I believe that on machines with only VS15, the gating introduced by #1009 may actually incorreclty resolve to true, preventing installs.

**Fix**
Move the if dev14 guard to the outer scope.


mousetraps:
:+1:

PR #1034, opened 2016-06-14
Always install node typings and gate alert to when typings actually created

### LINKED ISSUE #1032: After switching from ES5 -> ES6, typings InfoBar appears, but install doesn't complete
1. Start VS with ES5 mode on
2. Create new Node.js Console app project
3. Switch from ES6 -> ES5

Expected: Node.d.ts downloaded
Actual: Infobar appears, but typings folder is not there.

mjbvz:
Two small fixes:
* Ensure we always install the `node` typings. It currently is only installed when loading a project for the first time.
* Only show the typings warning when we actually create a typings folder. The first fix should prevent us hitting this in most cases, but its worth fixing so we don't hit this in the future.

closes #1032


mousetraps:
:+1:

PR #1046, opened 2016-06-14
Fix _reference.js template files for starter express apps

mjbvz:
These files look to be corrupted somehow, both the line endings and the content: see #1036. Just recreate them from their old contents.

Closes #1036

PR #1050, opened 2016-06-14
Fix _reference.js files again

### LINKED ISSUE #1036: _references.js in Starter Node.js Express 3 App has line ending differences
image

mjbvz:
It looks like #1046 only fixed the `_reference.js` files on my local machine. When I synced to master to pick up the change, the files were corrupted again.

This time, I didn't use VS for editing and force converted everything to utf-8, while also adding a `.gitattribute` to force js files to use text diffs (these corrupted files currently get treated as binary files: http://stackoverflow.com/questions/6855712/why-does-git-treat-this-text-file-as-a-binary-file.)

Closes #1036


mousetraps:
:+1:

PR #1053, opened 2016-06-14
Delete ExternalTools directory on Npm Cache Clear

### LINKED ISSUE #1044: Npm Options delete cache button should delete AppData typings install
That way we give the user a somewhat gracefully degrading experience if their install gets messed up for some reason.

mjbvz:
This change updates the npm cache clear button to also delete our ExternalTools directory. This directory contains node packages used internally by NTVS, so we want to provide a way to delete it if users get in a bad state.

Also extracted the localappdata paths for NTVS to constants.

closes #1044


Nodejs/Product/Nodejs/Options/NodejsNpmOptionsControl.cs
```diff
@@ -42,20 +36,26 @@ public partial class NodejsNpmOptionsControl : UserControl {
...
-                Directory.Delete(_npmCachePath, true);                    
-                _cacheClearedSuccessfully.Visible = true;
+                Directory.Delete(cachePath, true);
+               return true;
             } catch (DirectoryNotFoundException) {
                 // Directory has already been deleted. Do nothing.
-                _cacheClearedSuccessfully.Visible = true;
+                return true;
             } catch (IOException exception) {
                 // files are in use or path is too long
```
mousetraps: Huh, just now realizing this case might be pretty likely in the case of `typings`... what's the max path length of typings?

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsControl.cs
```diff
@@ -42,20 +36,26 @@ public partial class NodejsNpmOptionsControl : UserControl {
         }
 
         private void ClearCacheButton_Click(object sender, EventArgs e) {
+            bool didClearNpmCache = DeleteCacheDirectory("npm cache", NodejsConstants.NpmCachePath);
+            bool didClearTools = DeleteCacheDirectory("NTVS external tools", NodejsConstants.ExternalToolsPath);
```
mousetraps: nit: NTVS -> Node.js

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsControl.cs
```diff
@@ -42,20 +36,26 @@ public partial class NodejsNpmOptionsControl : UserControl {
...
+
+            _cacheClearedSuccessfully.Visible = (didClearNpmCache && didClearTools);
```
mousetraps: nit: remove parens

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsControl.cs
```diff
@@ -42,20 +36,26 @@ public partial class NodejsNpmOptionsControl : UserControl {
...
+        }
+
+        private static bool DeleteCacheDirectory(string displayName, string cachePath) {
```
mousetraps: nit: -> TryDeleteCacheDirectory

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsControl.cs
```diff
@@ -42,20 +36,26 @@ public partial class NodejsNpmOptionsControl : UserControl {
...
                 MessageBox.Show(
-                           string.Format("Cannot clear npm cache. {0}", exception.Message),
-                           "Cannot Clear npm Cache",
-                           MessageBoxButtons.OK,
-                           MessageBoxIcon.Information
-                       );
+                    string.Format("Cannot clear {0}. {1}", displayName, exception.Message),
+                    string.Format("Cannot Clear {0}", displayName),
+                    MessageBoxButtons.OK,
+                    MessageBoxIcon.Information);
```
mousetraps: If neither are able do be deleted successfully, we're going to display two message boxes in a row, which is a pretty poor user experience. Maybe we should just point to the Node.js Tools directory as a whole?

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsControl.cs
```diff
@@ -64,12 +64,12 @@ public partial class NodejsNpmOptionsControl : UserControl {
                 }
 
                 MessageBox.Show(
-                           string.Format("Cannot clear npm cache. Try manually deleting the directory: {0}", _npmCachePath),
-                           "Cannot Clear npm Cache",
-                           MessageBoxButtons.OK,
-                           MessageBoxIcon.Information
-                       );
+                    string.Format("Cannot clear {0}. Try manually deleting the directory: {1}", displayName, cachePath),
+                    string.Format("Cannot Clear {0}", displayName),
```
mousetraps: let's extract these strings to SR

mousetraps:
Left some comments... we may want to reconsider our approach here if we think that we are significantly more likely to fail to delete typings than we are to delete the existing cache directory.

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsControl.cs
mjbvz: Sounds like a good idea. Will make this change so we just tell users to delete the external tools dir. It's not used for anything else is it?

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsControl.cs
mjbvz: Here was the script I used:

```ps
C:\Users\matb.REDMOND\AppData\Local\Microsoft\Node.js Tools> Get-ChildItem -Recurse . | % { $_.fullname } | sort -Property Length
```

Longest path is 150 characters with Node 6 and my relatively short user name.

mjbvz: On < Node 6, the max path was 273 :sad:

I changed the logic to use `robocopy` to nuke the directory instead. This was tested with Node 6 and Node 0.12, and both seem to work fine now.

For reference, another way to delete long path files is with the [Windows `DeleteFile` function](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363915%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396):

> In the ANSI version of this function, the name is limited to MAX_PATH characters. To extend this limit to 32,767 wide characters, call the Unicode version of the function and prepend "\\\\?\" to the path.

Nodejs/Product/Nodejs/Options/NodejsNpmOptionsControl.cs
```diff
@@ -42,34 +37,57 @@ public partial class NodejsNpmOptionsControl : UserControl {
...
+            }
+
             try {
-                Directory.Delete(_npmCachePath, true);                    
-                _cacheClearedSuccessfully.Visible = true;
-            } catch (DirectoryNotFoundException) {
-                // Directory has already been deleted. Do nothing.
-                _cacheClearedSuccessfully.Visible = true;
-            } catch (IOException exception) {
+                // To handle long paths, nuke the directory contents with robocopy
```
mousetraps: nit: nuke -> exterminate 🤖

mousetraps:
:+1:

Issue #934, opened 2016-05-09
Breakpoints Sometimes not resolving with Node 6.1.0


mjbvz:
**Repo**

1. Install Node 6.1.0
2. Create new app.
3. Create new file (seems to repo more in larger projects and files besides the main file).
3. Set some breakpoints this new file.
4. `require` the file from main.
5. Try debugging.

**Expected**
Debugger stops at breakpoints

**Actual**
Breakpoints are not resolved.

Breakpoints were already less stable with Node 6.0.0, but with 6.1.0 I'm not seeing them resolved at all. Node 5.x branches work fine.

https://github.com/Microsoft/vscode/issues/6029

NoelAbrahams:
Any idea when this is likely to be resolved? This issue is blocking our adoption of Node 6.1 so would appreciate a fix soon. Thanks.

mjbvz:
I'm looking into this, but right now it is not clear what changed on the Node side of things or what the root cause of the problem is.

mjbvz:
Quick update: The root cause seems to be a change on the node side of things, not in NTVS directly , but it is possible we were relying on bugged behavior previously. The issue is nondeterministic and more common on larger projects. I believe it also exists in Node 6.0.

I'm seeing several different behaviors:
* Sometimes everything works.
* Sometimes breakpoints are not resolved.
* Sometimes breakpoints are resolved but then cause node to emit `Internal error: illegal access"` when hit and we call `backtrace`.

https://github.com/node-inspector/node-inspector/issues/864

mjbvz:
Update on some of the earlier leads:

The path differences don't seem to matter. I tried using a case insensitive regex to match the file name when setting the initial breakpoint, and using a file path with no case differences except for the drive letter. These seem unrelated to the breakpoint being hit or not.

Also, I see the backtrack trigger `Internal error: illegal access` both when breakpoints are hit and when they are not. Setting the breakpoint inside of the express router seems to be a reliable trigger:

```js
router.get('/', function (req, res) {
    res.render('index', { title: 'Express' }); // set breakpoint here
});
```

mousetraps:
This fix is available in v1.2 beta. Please try it out and let us know what you think! https://aka.ms/ntvslatest

NoelAbrahams:
We've been trying out 1.2 beta with Node 6.2 and the breakpoints seem to be working so far. Thanks.

PR #1054, opened 2016-06-14
Fix Node 6 Debugging

mjbvz:
Issue #934

##### Bug
Initial breakpoints currently sometimes do not resolve when using Node 6.x. In most cases, you can remove the breakpoint and add it back about 30 seconds after the program has started in order to get it to resolve.

##### Fix
I believe something with Node 6 changed the casing for the names of scripts in the v8-debug protocol. 

There were two fixes here:

* Use a case insensitive regular expression to match the script names when setting a breakpoint (I believe this is similar to what VS code does)
* Fix one case where we were relying on a case sensitive comparision to update a breakpoint.

This is not quite tested to be get merged in just yet, but wanted to see if there were any initial concerns or feedback with the approch here.

### DIFF
--- a/Nodejs/Product/Nodejs/Debugger/Commands/SetBreakpointCommand.cs
+++ b/Nodejs/Product/Nodejs/Debugger/Commands/SetBreakpointCommand.cs
@@ -61,12 +61,9 @@ namespace Microsoft.NodejsTools.Debugger.Commands {
             if (_module != null) {
                 _arguments["type"] = "scriptId";
                 _arguments["target"] = _module.Id;
-            } else if (remote) {
+            } else {
                 _arguments["type"] = "scriptRegExp";
                 _arguments["target"] = GetCaseInsensitiveRegex(_position.FileName);
-            } else {
-                _arguments["type"] = "script";
-                _arguments["target"] = _position.FileName;
             }
 
             if (!NodeBreakpointBinding.GetEngineEnabled(_breakpoint.Enabled, _breakpoint.BreakOn, 0)) {
--- a/Nodejs/Product/Nodejs/Debugger/Commands/SetBreakpointCommand.cs
+++ b/Nodejs/Product/Nodejs/Debugger/Commands/SetBreakpointCommand.cs
@@ -126,40 +126,24 @@ namespace Microsoft.NodejsTools.Debugger.Commands {
 
         private static string CreateRemoteScriptRegExp(string filePath) {
             string fileName = Path.GetFileName(filePath) ?? string.Empty;
-            bool trailing = fileName != filePath;
-
-            fileName = Regex.Escape(fileName);
-
-            var builder = new StringBuilder();
-            if (trailing) {
-                builder.Append(pathSeperatorCharacterGroup);
-            } else {
-                builder.Append('^');
-            }
-
-            AddCaseInsensitiveStringToRegExp(fileName, builder);
-
-            builder.Append("$");
-            return builder.ToString();
+            string start = fileName != filePath ? pathSeperatorCharacterGroup : "^";
+            return string.Format("{0}{1}$", start, CreateCaseInsensitiveRegExpFromString(fileName));
         }
 
         private static string CreateLocalScriptRegExp(string filePath) {
-            string fileName = Regex.Escape(filePath);
-
-            var builder = new StringBuilder();
-            builder.Append('^');
-            AddCaseInsensitiveStringToRegExp(fileName, builder);
-            builder.Append("$");
-            return builder.ToString();
+            return string.Format("^{0}$", CreateCaseInsensitiveRegExpFromString(filePath));
         }
 
         /// <summary>
-        /// Add a case-insensitive string to a string buffer containing a regular expression.
+        /// Convert a string into a case-insensitive regular expression.
         /// 
         /// This is a work around for the fact that we cannot pass a regex case insensitive modifier to the Node (V8) engine.
         /// </summary>
-        private static void AddCaseInsensitiveStringToRegExp(string str, StringBuilder builder) {
-            foreach (var ch in str) {
+        private static string CreateCaseInsensitiveRegExpFromString(string str) {
+            string escapedString = Regex.Escape(str);
+
+            var builder = new StringBuilder();
+            foreach (var ch in escapedString) {
                 string upper = ch.ToString(CultureInfo.InvariantCulture).ToUpper();
                 string lower = ch.ToString(CultureInfo.InvariantCulture).ToLower();
                 if (upper != lower) {
@@ -171,6 +155,7 @@ namespace Microsoft.NodejsTools.Debugger.Commands {
                     builder.Append(upper);
                 }
             }
+            return builder.ToString();
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Nodejs/Product/Nodejs/Debugger/NodeDebugger.cs
```diff
@@ -530,15 +530,13 @@ public NodeDebugger(Uri debuggerEndpointUri, int id)
             foreach (NodeModule module in modules) {
                 NodeModule newModule;
                 if (GetOrAddModule(module, out newModule)) {
-                    if (newModule.FileName != newModule.JavaScriptFileName) {
```
mjbvz: I'm a bit more unsure what the original intention of this line was. I'm going to try stepping through it more with various node versions to see, but I never saw the old expression evaluate to `true` in my testing. 

My main concern is that the current approach may just be brute forcing breakpoints into working with Node 6, instead of addressing the fundamental issues. That could risk breaking both old and future node versions.

Nodejs/Product/Nodejs/Debugger/Commands/SetBreakpointCommand.cs
```diff
@@ -61,12 +61,9 @@ public SetBreakpointCommand(int id, NodeModule module, NodeBreakpoint breakpoint
             if (_module != null) {
                 _arguments["type"] = "scriptId";
                 _arguments["target"] = _module.Id;
-            } else if (remote) {
+            } else {
                 _arguments["type"] = "scriptRegExp";
                 _arguments["target"] = GetCaseInsensitiveRegex(_position.FileName);
```
mjbvz: Now that I look more closely at `GetCaseInsensitiveRegex`, it only seems to use the filename in the regexp. That's odd,  since, in my understanding, with a project like:

```
a ->
    index.js
b ->
   index.js
```

You could not set a breakpoint in both of the `index.js` files.

Not sure this is proper fix given that, but perhaps a different regex can be used instead for local files.

mjbvz: Split the logic to avoid introducing regressions for the remote case. Updated the script to convert the entire path to a regexp instead of just the file name This also resolved the phantom breakpoint issue I was seeing earlier.

Testing on various node versions < 6.0 now.

Nodejs/Product/Nodejs/Debugger/NodeDebugger.cs
mousetraps: Hmm - did you try loading a TS project? This would probably evaluate to true in those scenarios.

I suspect there might be a better way of resolving the issue, but I don't think it'll *break* anything to add this (other than having a perf impact) - basically what you're doing is attempting to bind the breakpoints a little later than we normally do (after scripts have been loaded), in which case I don't understand how the only thing that's changed with 6.x is the case sensitivity, and I think it would be worth digging into it a little further to understand specifically what changed or what existing issue has manifested itself (and also if there's a way to log this on our end so that it's easier to debug in the future.)

Also worth looking into when VSCode binds breakpoints, and if there's a separate codepath for JS vs. source-mapped files.

Nodejs/Product/Nodejs/Debugger/Commands/SetBreakpointCommand.cs
```diff
@@ -25,6 +25,8 @@
 
 namespace Microsoft.NodejsTools.Debugger.Commands {
     sealed class SetBreakpointCommand : DebuggerCommand {
+        private static string pathSeperatorCharacterGroup = string.Format("[{0}{1}]", Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
```
mousetraps: nit: prepend with underscore

Nodejs/Tests/Core/Debugger/Commands/SetBreakpointCommandTests.cs
```diff
@@ -24,6 +26,37 @@ namespace NodejsTests.Debugger.Commands {
     [TestClass]
     public class SetBreakpointCommandTests {
         [TestMethod, Priority(0), TestCategory("Debugging")]
+        public void CreateLocalScriptRegExpShouldCreateCaseInSensitiveRegularExpression() {
+            const string drive = "c:";
+            var pathParts = new[] { "nOdE", "IS", "awsome.js" };
```
mousetraps: awsome? Is that like 1/7 less awesome? :wink:

Nodejs/Tests/Core/Debugger/Commands/SetBreakpointCommandTests.cs
```diff
@@ -114,8 +147,8 @@ public class SetBreakpointCommandTests {
             Assert.AreEqual(commandId, setBreakpointCommand.Id);
             Assert.AreEqual(
                 string.Format(
-                    "{{\"command\":\"setbreakpoint\",\"seq\":{0},\"type\":\"request\",\"arguments\":{{\"line\":{1},\"column\":{2},\"type\":\"script\",\"target\":\"{3}\",\"ignoreCount\":1}}}}",
-                    commandId, line, column, fileName.Replace(@"\", @"\\")),
+                    "{{\"command\":\"setbreakpoint\",\"seq\":{0},\"type\":\"request\",\"arguments\":{{\"line\":{1},\"column\":{2},\"type\":\"scriptRegExp\",\"target\":\"{3}\",\"ignoreCount\":1}}}}",
+                    commandId, line, column, SetBreakpointCommand.CreateLocalScriptRegExp(fileName).Replace(@"\", @"\\")),
                 setBreakpointCommand.ToString());
         }
```
mousetraps: If you haven't already - can you also take a look through existing UI tests and see if we need to change anything there?

Nodejs/Product/Nodejs/Debugger/NodeDebugger.cs
mjbvz: Here's my rough understanding of the problem.

First is the case issue. Without the correct case, we fail to set breakpoints properly at all. Fix this, and we run into another issue: Node 6 does not seem to handle the way we set breakpoints during initial program startup. This manifests itself as a race condition, which explains why see the bug more often in the more complex express apps and more often in external files.

The root cause of the second issue could be somewhere in our code still, but the casing fix alone is was not enough to consistently fix the bug for me. This change works by making us reset the breakpoints when we get an "afterCompile" event for a file, brute forcing the breakpoint into existence.

Nodejs/Tests/Core/Debugger/Commands/SetBreakpointCommandTests.cs
mjbvz: Not to my knowledge. The user experience should be unchanged and I don't see any new failures on the UI tests. Do you have an specific concerns?

Nodejs/Tests/Core/Debugger/Commands/SetBreakpointCommandTests.cs
mjbvz: This test was actually a meta test designed to test the ability of the programmer to spell...

Should be fixed now.

mousetraps:
:+1:

Issue #1051, opened 2016-06-14
Remove Express 3 Templates


mjbvz:
We want to encourage new users to use modern, more well documented and supported packages when they create a new project from the template. Therefore, we should remove the Express 3 templates in favor of the Express 4 based templates.

This will not prevent existing projects that use the Express 3 templates from working, and users will still be able to install and use Express 3 just fine.

This work is part of #1037

mjbvz:
@mousetraps mentioned that we would want to update the "starter app" templates to express 4 as well. The starter apps offer a bit more html and some common client side libraries (bootstrap, jquery, ...) to work with.

mjbvz:
Fixed by #1055. Tracking express 3 -> express 4 update work with #1058

PR #1055, opened 2016-06-15
Delete Basic and Azure JavaScript  Express 3 Templates

mjbvz:
Part of #1051

This change just deletes the existing (non-starter app) JavaScript Express 3 templates. Later changes will update the other existing Express 3 templates to use Express 4, but that work can happen in stages.


mjbvz:
@mousetraps Do we specifically reference Express 3 and these templates in any of our published marketing materials or documentation? 

I'll update this page once this goes in, with note about NTVS 1.2: https://github.com/Microsoft/nodejstools/wiki/Projects#project-types

mousetraps:
:+1:, though there are merge conflicts

Issue #1037, opened 2016-06-14
29 Errors in Starter Node.js Express 3 App


mousetraps:
It says 45 because I'm still on update 2 and running into #928. 29 of them are unexpected. 
![image](https://cloud.githubusercontent.com/assets/762848/16027600/0c61b424-318c-11e6-9cac-62e09d35f54f.png)

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	TS1127	Invalid character.	ExpressApp23	c:\Users\Itani\documents\visual studio 2015\Projects\ExpressApp23\ExpressApp23\public\javascripts\_references.js	4	Active
Error	TS5055	Cannot write file 'c:/Users/Itani/documents/visual studio 2015/Projects/ExpressApp23/ExpressApp23/app.js' because it would overwrite input file.	ExpressApp23		1	Active
Error	TS5055	Cannot write file 'c:/Users/Itani/documents/visual studio 2015/Projects/ExpressApp23/ExpressApp23/public/javascripts/_references.js' because it would overwrite input file.	ExpressApp23		1	Active
Error	TS5055	Cannot write file 'c:/Users/Itani/documents/visual studio 2015/Projects/ExpressApp23/ExpressApp23/public/javascripts/bootstrap.js' because it would overwrite input file.	ExpressApp23		1	Active
Error	TS5055	Cannot write file 'c:/Users/Itani/documents/visual studio 2015/Projects/ExpressApp23/ExpressApp23/public/javascripts/bootstrap.min.js' because it would overwrite input file.	ExpressApp23		1	Active
Error	TS5055	Cannot write file 'c:/Users/Itani/documents/visual studio 2015/Projects/ExpressApp23/ExpressApp23/public/javascripts/jquery-1.10.2.intellisense.js' because it would overwrite input file.	ExpressApp23		1	Active
Error	TS5055	Cannot write file 'c:/Users/Itani/documents/visual studio 2015/Projects/ExpressApp23/ExpressApp23/public/javascripts/jquery-1.10.2.js' because it would overwrite input file.	ExpressApp23		1	Active
Error	TS5055	Cannot write file 'c:/Users/Itani/documents/visual studio 2015/Projects/ExpressApp23/ExpressApp23/public/javascripts/jquery-1.10.2.min.js' because it would overwrite input file.	ExpressApp23		1	Active
[...]

PR #1056, opened 2016-06-15
Add `typings.json` to Express 4 Template and Auto Install Typings from `typings.json`

mjbvz:
Issue #1037

##### Bug
This fix addresses the two `serve-static` related errors we're  seeing in #1037. The root cause is that the `express.d.ts` file references other packages not installed by default. For our Express templates, this results in two errors  and a pretty bad first experiance.

##### Fix
Two parts to the fix:

* Update typings acquisition to automatically attempt to install typings from a top level `typings.json` file, if such a file exists.
* Add a `typings.json` file to the Express4 template.

I'll add similar `typings.json` files to the other express 4 based JavaScript templates, if this is the approch we decide we want to go with.


Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -78,37 +78,49 @@ internal class TypingsAcquisition {
 
         public Task<bool> AcquireTypings(IEnumerable<string> packages, Redirector redirector) {
             return typingsToolGlobalWorkSemaphore.WaitAsync().ContinueWith(async _ => {
-                var typingsToAquire = GetNewTypingsToAcquire(packages);
-                var success = await DownloadTypings(typingsToAquire, redirector);
-                if (success) {
-                    _acquiredTypingsPackageNames.Value.UnionWith(typingsToAquire);
+                try {
+                    return await DownloadTypingsForProject(redirector) && await DownloadTypingsForPackages(packages, redirector);
```
mousetraps: Let's flip these two around so that typings.json has the "final say" because if someone has a typings.json file, they've explicitly chosen to lock their dependency versions.

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -910,6 +910,7 @@
       <IncludeInVSIX>true</IncludeInVSIX>
       <VSIXSubPath>ItemTemplates</VSIXSubPath>
     </Content>
+    <ZipProject Include="ProjectTemplates\Express4App\typings.json" />
```
mousetraps: nit: I know VS chose this location, but can we just move this closer to the rest of the Express4App project template?

Nodejs/Product/Nodejs/ProjectTemplates/Express4App/typings.json
```diff
@@ -0,0 +1,16 @@
...
+  "globalDependencies": {
+    "body-parser": "registry:dt/body-parser#0.0.0+20160317120654",
+    "cookie-parser": "registry:dt/cookie-parser#1.3.4+20160316155526",
+    "debug": "registry:dt/debug#0.0.0+20160317120654",
+    "express": "registry:dt/express#4.0.0+20160317120654",
+    "express-serve-static-core": "registry:dt/express-serve-static-core#0.0.0+20160602151406",
+    "jade": "registry:dt/jade#0.0.0+20160316155526",
+    "mime": "registry:dt/mime#0.0.0+20160316155526",
+    "morgan": "registry:dt/morgan#1.7.0+20160524142355",
+    "node": "registry:dt/node#6.0.0+20160613154055",
```
mousetraps: let's not lock down the version for node

mousetraps:
A few comments and :+1: after that!

Issue #668, opened 2016-01-24
ES6 Intellisense in NIW


macfarmw:
Thanks for working on ES6 syntax. This is a big help to adopting it more rapidly.

The ES6 Intellisense works nicely in the code editor but not in the NIW. The NIW highlights this line with several red squiggles.

    var x = `The answer is ${3+2}`;

This statement runs in the NIW without any error but it will be confusing to many developers to get error highlights on valid ES6 code.

mousetraps:
Thanks for the feedback!

@jramsay @paulvanbrenk @bowdenk7 - just a heads up that we should start thinking about what it would take to enable Salsa in the Interactive Window as well and also in the upcoming folder/file based project system stuff. One complexity with the Interactive Window is that it can be opened outside of a project context altogether.

gdennie:
Similar issue with code editor for the following line in App.jsx:

var {Routes, Route} = require('react-router');

'{' and '=' flagged as 'expected indentifier' and 'syntax error' respectively.

mjbvz:
We're working on this now. Closing this issue in favor of #1090

PR #1059, opened 2016-06-16
#916 #668 No IntelliSense in Interactive Window with ES6 Mode

### LINKED ISSUE #916: No Intellisense in interactive Window With ES6 Intellisense
Interactive widow should show intellisense with ES6 preview Intellisense.

Here's the current behavior with Full intellisense:

image

and with ES6 Preview:

image

mousetraps:
* Initialize loose analyzer with Quick IntelliSense if Preview mode is on.
* Don't display errors when ES6 Mode is on
* Don't update status bar for implicit projects

close #916, address part of #668

### DIFF
--- a/Nodejs/Product/Nodejs/NodejsPackage.cs
+++ b/Nodejs/Product/Nodejs/NodejsPackage.cs
@@ -637,7 +637,7 @@ namespace Microsoft.NodejsTools {
         internal VsProjectAnalyzer DefaultAnalyzer {
             get {
                 if (_analyzer == null) {
-                    _analyzer = new VsProjectAnalyzer(IntellisenseOptionsPage.AnalysisLevel, IntellisenseOptionsPage.SaveToDisk);
+                    _analyzer = CreateLooseVsProjectAnalyzer();
                     LogLooseFileAnalysisLevel();
                     _analyzer.MaxLogLength = IntellisenseOptionsPage.AnalysisLogMax;
                     IntellisenseOptionsPage.AnalysisLevelChanged += IntellisenseOptionsPageAnalysisLevelChanged;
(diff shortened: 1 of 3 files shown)


Nodejs/Product/Nodejs/NodejsPackage.cs
```diff
@@ -661,6 +661,21 @@ internal sealed partial class NodejsPackage : CommonPackage {
             LogLooseFileAnalysisLevel();
         }
 
+        private VsProjectAnalyzer CreateLooseVsProjectAnalyzer() {
```
mjbvz: Are there any non-repl flows through `GetAnalyzer` that could be regressed by changing `DefaultAnalyzer`?

Nodejs/Product/Nodejs/NodejsPackage.cs
```diff
@@ -661,6 +661,21 @@ internal sealed partial class NodejsPackage : CommonPackage {
...
+            // it will throw to JSLS. This means that currently, loose files will be handled by the JSLS editor,
+            // rather than the NodeLS editor (which means that the loose project analyzer will not be invoked
+            // in these scenarios.)
+            //
+            // Because Salsa doesn't support the stitching together inputs from the surrounding text view,
+            // we do not throw to Salsa from the REPL window. However, in order to provide a workable editing
+            // experience within the REPL context, we initialize the loose analyzer with Quick IntelliSense
+            // during ES6 mode.
+            var analysisLevel = IntellisenseOptionsPage.AnalysisLevel == AnalysisLevel.Preview ? AnalysisLevel.NodeLsMedium : IntellisenseOptionsPage.AnalysisLevel;
+            return new VsProjectAnalyzer(analysisLevel, false);
```
mjbvz: With this change, we are taking an explicit dependency back on our old AnalysisEngine, whereas we currently are almost in a state where we could delete all that code, simplifying the maintenance and development of NTVS considerably, therefore improving product quality in the long run.

I don't have any problem with making this change for 1.2, but want to ensure we have a plan in place to fix the underlying issues that are preventing us from using a modern intellisense engine in the repl window. Is this the case?

mjbvz:
LGTM. Merge when ready, but please respond to the more general comments/questions I had about this change and planning.

Nodejs/Product/Nodejs/NodejsPackage.cs
mousetraps: Just added #668 to our On Radar milestone so it doesn't fall through the cracks. We won't be addressing it for Dev14, but certainly for Dev15

Nodejs/Product/Nodejs/NodejsPackage.cs
mousetraps: The two usage scenarios are:
* opening loose files, which now just throws to JSLS
* interactive window

PR #1060, opened 2016-06-16
Fix function declarations in interactive window

### LINKED ISSUE #1030: Cannot define function in Interactive Window
**Expected:**
image

**Actual:**
image

mjbvz:
#####  Bug
Function declarations are broken in interactive window. The root cause is that we are wrapping them in parens, creating function expressions instead.

##### Fix
Don't wrap function declarations in parens for evaluations.

Closes #1030

### DIFF
--- a/Nodejs/Product/Nodejs/visualstudio_nodejs_repl.js
+++ b/Nodejs/Product/Nodejs/visualstudio_nodejs_repl.js
@@ -108,7 +108,9 @@ function processRequest(command) {
                 var obj;
                 // function f() { } should return undefined, wrapping it in parens
                 // produces a value, so ignore it if we have a function definition.
-                if (!/^\s*function/.test(command["code"])) {
+                if (/^\s*function[\s\(\*]/.test(command["code"])) {
+                    obj = vm.runInContext(evalPrefix + command["code"], context);
+                } else {
                     try {
                         // object literals are ambigious, so first try w/ parens.
                         obj = vm.runInContext(evalPrefix + '(' + command["code"] + ')', context);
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/visualstudio_nodejs_repl.js
```diff
@@ -108,7 +108,9 @@ function processRequest(command) {
                 var obj;
                 // function f() { } should return undefined, wrapping it in parens
                 // produces a value, so ignore it if we have a function definition.
-                if (!/^\s*function/.test(command["code"])) {
```
mjbvz: The old regexp would match any var starting with `function`. We only want to not apply wrapping for function declarations that use the function keyword explicitly. If I recall, function keyword in this context can either be  followed by either:

* Space and then identifier
* Paren for argument list (actually, this would make it an function expression but JS has a weird ambiguity about function declarations vs function expressions).
* A `*` for a generator.

Nodejs/Product/Nodejs/visualstudio_nodejs_repl.js
```diff
@@ -120,8 +122,6 @@ function processRequest(command) {
                             throw err;
                         }
                     }
-                } else {
-                    obj = vm.runInContext(evalPrefix + '(' + command["code"] + ')', context);
```
mousetraps: Looks like the parens were introduced in #876 - do you remember if there was a reason for them?

mousetraps:
:+1:

Nodejs/Product/Nodejs/visualstudio_nodejs_repl.js
mjbvz: Methinks that month-ago @mjbvz was suffering from a bad case of the copy-pastes. 

Questions to ask with any introduced regression like this: why wasn't this caught sooner and how can we prevent this same mistake from happening again. Adding tests is an important step in addressing both of those questions, so I've opened #1063 to track this work.

PR #1064, opened 2016-06-17
Remove duplicate Vstman Template manifest entries

mjbvz:
##### Bug
There appear to be duplicate entries in the `vstman` template manifest file.

#### Fix
Remove these duplicate entries.


mjbvz:
I'll also check `Templates/VS_Nodejs.item.vstman` for duplicates and remove these.

PR #1066, opened 2016-06-17
Fix corrupted CloudService template files

### LINKED ISSUE #1065: Corrupted CloudService vstemplate files
Example file: `Nodejs\Product\Nodejs\ProjectTemplates\CloudService\CloudService.vstemplate`

image

Pretty sure it's not supposed to look that way. GitHub seems to render the file just fine: https://github.com/Microsoft/nodejstools/blob/master/Nodejs/Product/Nodejs/ProjectTemplates/CloudService/CloudService.vstemplate

Will do a quick pass to see if other templates files are effected as well. We saw this same issue previously with #1036

mjbvz:
##### Bug
Windows is displaying corrupted file contents for the CloudService templates. These files do not show as corrupted on GitHub.

##### Fix
Revert all these files to their previous content and explicitly convert them to utf-8.

Closes #1065


mjbvz:
I did few simple letter searchs to see if any more files were showing up with similar corruption. This change fixes all the files identified, but there could be others still.

mousetraps:
:+1:

PR #1071, opened 2016-06-18
fixes of obvious code mistakes

DamirAinullin:
fixes of some simple obvious code mistakes and warnings.

### DIFF
--- a/Nodejs/Product/Profiling/Profiling/StandaloneTargetView.cs
+++ b/Nodejs/Product/Profiling/Profiling/StandaloneTargetView.cs
@@ -152,7 +152,7 @@ namespace Microsoft.NodejsTools.Profiling {
                     WorkingDirectory.IndexOfAny(Path.GetInvalidPathChars()) == -1 &&
                     (Path.IsPathRooted(ScriptPath) || Path.IsPathRooted(WorkingDirectory)) &&
                     File.Exists(Path.Combine(WorkingDirectory, ScriptPath)) &&
-                    (string.IsNullOrEmpty(WorkingDirectory) || Directory.Exists(WorkingDirectory)) &&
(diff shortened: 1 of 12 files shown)


Nodejs/Product/Profiling/Profiling/StandaloneTargetView.cs
```diff
@@ -152,7 +152,7 @@ public StandaloneTargetView(StandaloneTarget template)
                     WorkingDirectory.IndexOfAny(Path.GetInvalidPathChars()) == -1 &&
                     (Path.IsPathRooted(ScriptPath) || Path.IsPathRooted(WorkingDirectory)) &&
                     File.Exists(Path.Combine(WorkingDirectory, ScriptPath)) &&
-                    (string.IsNullOrEmpty(WorkingDirectory) || Directory.Exists(WorkingDirectory)) &&
+                    (WorkingDirectory == string.Empty || Directory.Exists(WorkingDirectory)) &&
```
mjbvz: This changes the behavior and some analyzers will complain about comparing to `string.Empty` directly. Is there a specific reason for this change?

mjbvz:
Thanks for the fixes. Just one small comment to take a look at before we merge this in.

What tool did you use to find these?

Nodejs/Product/Profiling/Profiling/StandaloneTargetView.cs
DamirAinullin: We use WorkingDirectory at the line 152, before check it for null. So I can suppose that check is not needed. Maybe NullReferenceException is possible  at the line 152. Maybe we should check it for null before using.

DamirAinullin:
Thanks for positive response. I just try to help your project. I use static analyser PVS-Studio. It is a wonderful tool.

mjbvz:
Thanks for details and thanks again for finding these. LGTM, merging this in now.

PR #1081, opened 2016-06-21
Add Microsoft Code of Conduct link in readme

mjbvz:
Adds links to Microsoft code of conduct to README


mousetraps:
Let's actually add this to the bottom of the contributing section. :+1: after that!

Issue #1080, opened 2016-06-21
Interactive Window busted in Node.js v0.10.x and below


mousetraps:
**Expected**
Enter statements into interactive window and they get executed properly

**Actual**
`TypeError: needs a 'context' argument.`

![image](https://cloud.githubusercontent.com/assets/762848/16239619/64e84536-379a-11e6-886b-367391e1b9d8.png)

mousetraps:
@avitalb is looking into this

mousetraps:
Looks like it's only broken in v0.10.x and below, probably as a result of using new APIs in #876

mjbvz:
See this old Node documentation for `vm`: https://nodejs.org/docs/v0.10.40/api/vm.html

I suspect we need to change the context creation line to `context = vm.createInContext(context)`

PR #1082, opened 2016-06-21
fixed Issue #1080 - fixed interactive window type error due to contex…

avitalb:
Issue #1080

##### Bug
Interactive Window type error due to context not being returned in node 0.10x

##### Fix
Context is now being returned


mjbvz:
👍looks good to me

PR #1084, opened 2016-06-22
Adds new line character to test framework not found output messages

### LINKED ISSUE #1042: Should add newline in Output pane messages, otherwise the user must scroll in order to read important information
image

mjbvz:
This change manually adds new lines to the error messages that are displayed when a test framework cannot be found. The mitigation steps in those cases are usually pushed off the screen.

##### Testing
Manually verified messages display correctly.

Closes #1042


mousetraps:
:+1:

Issue #1083, opened 2016-06-21
.targets file not copied to proper location in EnvironmentSetup.ps1


avitalb:
##### Expected Behavior 
copied to "C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\Node.js Tools\Microsoft.NodejsTools.targets"

##### Actual Behavior 
copied to "C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\14.0\Node.js Tools\Microsoft.NodejsTools.targets"

mjbvz:
Fixed by #1088

PR #1088, opened 2016-06-22
fixed Issue #1083: EnvironmentSetup.ps1 copies .targets file to folder named after just the version number instead of 'v' and then the version number. 

avitalb:
Issue #1083 

##### Bug
EnvironmentSetup.ps1 copied the .targets file to a folder named after the version number instead of a folder named v($version.number).

##### Fix
EnvironmentSetup.ps1 now copies .targets file to a folder named v$($version.number).


mjbvz:
👍 LGTM

PR #1091, opened 2016-06-22
Fix Dispose on PackageJsonFileNode

mjbvz:
#####  Bug
`PackageJsonFileNode` file system watcher not being disposed of properly.

#####  Fix
Dispose of this object

##### Testing
Manual testing of this path with debugger to confirm this is hit


mousetraps:
:+1: for RTM

PR #1095, opened 2016-06-23
Fix survey news showing multiple times and when users opt out

### LINKED ISSUE #1086: Multiple pages of NTVS News opens up on load
Hi,

**Version**: NTVS 1.2 Beta / VS 2015 Update 2

**Actual behaviour**

When I open up the solution 9 tabs showing NTVS News (URL: http://ntvs.azurewebsites.net/news/1010/) open up.

**Expected behaviour**

I would like to see a grand total of 0 tabs of NTVS News opening on launch.

mjbvz:
Issue #1086

##### Bug
Not honoring never setting for Survey news options. The never case is actually the worst one since it can open multiple survey news pages, one for each project.

##### Fix
Make sure we do honor this option and never automatically display survey news when enabled.

##### Testing
Manually verified bad behavior before and that this change fixes the bad behavior.

Closes #1086


mousetraps:
:+1:

PR #1097, opened 2016-06-23
Add better logging when node is not installed

### LINKED ISSUE #1069: Poor initial experience when Node.js is not installed
##### Expected Behavior 
We warn that Node.js is not installed (preferably in the document well, rather than in a modal dialog)

##### Actual Behavior 
IntelliSense doesn't work because typings haven't been downloaded because typings hasn't been installed because npm hasn't been installed because node hasn't been installed :wink:, and the only indication is this awesome error in Bower/npm window "Error - Cannot load global packages."

If you try debugging, you'll get a modal dialog, but it's a confusing experience until then.

mjbvz:
##### Bug
Poor initial experiance when launching NTVS without node installed.

##### Fix
Adds a few more logging points:
* Log if npm cannot be found
* Log if typings tool install fails
* Log if typings tool cannot be found.

All these are logged to the output window.

##### Testing
Manually tested by messing up node path and then correcting node path to make sure the errors are not shown

image

Closes #1069


mjbvz:
At some point, I'd like to revist #729. Adding strings is a pain right now and it's far too easy to introduce fun bugs through string name mismatches.

mousetraps:
:+1: this is a good start, but I'm not entirely sure if this is really a place users will look so we should continue to refine the design.

Issue #1104, opened 2016-06-24
Visual Studio Update 3 Install gating check is unreliable


mousetraps:
It works on some machines, but not on others. We should look into more reliable ways to determine the VS version.

PR #1106, opened 2016-06-27
#1104 VS Update 3 gating check is unreliable

mousetraps:
Issue #1104
* update way we check VS version - instead use VS install path to check assembly version
* change way we spawn U3 gating dialog - for some reason variables weren't being initialized properly at the EULA dialog, so in the interim, simply launch the dialog in a similar fashion that we do with other dialogs

##### Testing
* verify that the version check works properly by changing up the target assembly version
* verify the check works properly on a couple machines
* verify dialog doesn't pop up when changing / uninstalling NTVS
* verify dialog fires on both advanced and regular installs
* verify dialog appears at most once throughout flow


mousetraps:
Merging this in as I'd like to kick off a build with these bits. Will revert it it proves to be problematic after further testing.

Issue #1058, opened 2016-06-16
Update JavaScript Starter Express App Templates to Use Express 4


mjbvz:
Same reasoning as with #1051, but unlike with that issue, we currently do not offer any Express 4 based options. 

Here are the templates that need to be updated to use Express 4:

- [ ] AddWebSiteAzureExpressApp
- [ ] AddWebSiteAzureStarterExpressApp
- [ ] AddWebSiteExpressAppExpressApp
- [ ] AddWebSiteStarterExpressApp
- [ ] AzureStarterExpressApp
- [ ] StarterExpressApp

The typescript templates are out of scope for this issue.

@mousetraps, I'm trying to cut down the work here if possible, so could you help me understand the main purpose of the `AddWebSite*` templates vs their non `AddWebSite` counterparts? Can we combine the templates in some way, or at least reuse files between them?

mjbvz:
Closing in favor of removal of the Started Express Templates with #1270. These started templates are out of date and provide a poor example of how to organize a modern JavaScript project. Projects based on the started express templates will continue to work just fine.

PR #1113, opened 2016-06-29
Prototype Starter Express App template Express 3 to 4 change

mjbvz:
Related to #1058

This prototypes moving the basic StarterApp from Express3 to Express4. The update is based on our existing Express 4 template.

I'd like to get feedback on this change before doing the work to migrate the other 5 Starter Express based templates.

If we can cut down the number of templates, that would also help us a lot.


mjbvz:
Closing in favor of #1270. A later PR will ensure the existing basic express templates are up to date.

PR #1117, opened 2016-06-30
Adds Telemetry for IntelliSense Levels

### LINKED ISSUE #1096: Add Telemetry for Which IntelliSense Is Currently Selected
With NTVS 1.2, we would like to understand if user ever switch from the default ES6 IntelliSense mode back to ES5 IntelliSense. We really want to move users off of ES5 IntelliSense as much as possible, and having this data would help us identify potential issues that are keeping users on ES5 and help us plan out the longer term removal of the ES5 IntelliSense option.

mjbvz:
##### Bug
We would like to understand what IntelliSense level users are on and when they change IntelliSense levels. This will help us identify adoption of the ES6 analyzer as well as any potential issues keeping people on the old analyzer. It also helps us plan the phase out of the old analyzer

##### Fix
Adds two telemetry points:

* IntelliSenseLevelChanged - General, IntelliSense level has changed event. Sent when user changes the level through the options.
* IntelliSenseActivatedForProject - Logged for each project when the project is first loaded or when the intellisense level is changed.

These events are both implemented using helper extension methods on `NodejsPackage.Instance.TelemetryLogger`. This unifies and hides the actual logging logic. I also updated `LogProjectImported` to use this style of logging.

Closes #1096

### DIFF
--- a/Nodejs/Common/Telemetry/TelemetryEvents.cs
+++ b/Nodejs/Common/Telemetry/TelemetryEvents.cs
@@ -20,7 +20,7 @@ namespace Microsoft.NodejsTools.Telemetry {
     /// </summary>
     internal static class TelemetryEvents {
         public const string ProjectImported = "ProjectImported";
-        public const string IntelliSenseActivatedForProject = "IntelliSenseActivatedForProject";
-        public const string IntelliSenseLevelChanged = "IntelliSenseLevelChanged";
+        public const string AnalysisActivatedForProject = "AnalysisActivatedForProject";
+        public const string AnalysisLevelChanged = "AnalysisLevelChanged";
     }
 }
--- a/Nodejs/Common/Telemetry/TelemetryEvents.cs
+++ b/Nodejs/Common/Telemetry/TelemetryEvents.cs
@@ -20,5 +20,7 @@ namespace Microsoft.NodejsTools.Telemetry {
     /// </summary>
     internal static class TelemetryEvents {
         public const string ProjectImported = "ProjectImported";
+        public const string IntelliSenseActivatedForProject = "IntelliSenseActivatedForProject";
+        public const string IntelliSenseLevelChanged = "IntelliSenseLevelChanged";
     }
 }
--- a/Nodejs/Product/Nodejs/NodejsPackage.cs
+++ b/Nodejs/Product/Nodejs/NodejsPackage.cs
@@ -241,6 +241,8 @@ namespace Microsoft.NodejsTools {
             MakeDebuggerContextAvailable();
 
             IntellisenseOptionsPage.AnalysisLogMaximumChanged += IntellisenseOptionsPage_AnalysisLogMaximumChanged;
+            IntellisenseOptionsPage.AnalysisLevelChanged += IntellisenseOptionsPageAnalysisLevelChanged;
+            IntellisenseOptionsPage.SaveToDiskChanged += IntellisenseOptionsPageSaveToDiskChanged;
 
             InitializeLogging();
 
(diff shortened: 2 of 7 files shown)


Nodejs/Common/Telemetry/TelemetryEvents.cs
```diff
@@ -20,5 +20,7 @@ namespace Microsoft.NodejsTools.Telemetry {
     /// </summary>
     internal static class TelemetryEvents {
         public const string ProjectImported = "ProjectImported";
+        public const string IntelliSenseActivatedForProject = "IntelliSenseActivatedForProject";
```
mjbvz: Should this be AnalysisLevel instead? We seem to use IntelliSense level or IntelliSense mode in the UI and documentation, but analysis level in the code.

mousetraps: Yep AnalysisLevel would be better for consistency. For technical discussions, we typically make a distinction between "IntelliSense" - the VS-branded user experience - and the actual engine that's providing the code intelligence.

Nodejs/Product/Nodejs/NodejsPackage.cs
```diff
@@ -241,6 +241,8 @@ internal sealed partial class NodejsPackage : CommonPackage {
             MakeDebuggerContextAvailable();
 
             IntellisenseOptionsPage.AnalysisLogMaximumChanged += IntellisenseOptionsPage_AnalysisLogMaximumChanged;
+            IntellisenseOptionsPage.AnalysisLevelChanged += IntellisenseOptionsPageAnalysisLevelChanged;
+            IntellisenseOptionsPage.SaveToDiskChanged += IntellisenseOptionsPageSaveToDiskChanged;
```
mousetraps: Can you verify that LoadSettingsFromStorage executes before these event subscriptions? Otherwise the TelemetryLogger will log the IntelliSenseLevelChanged event every time the package is loaded.

Nodejs/Product/Nodejs/Telemetry/NodejsTelemetryExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+//
+//*********************************************************//
+
+using System;
+
+namespace Microsoft.NodejsTools.Telemetry {
+    /// <summary>
+    /// Extensions for logging telemetry events.
+    /// </summary>
+    internal static class NodejsTelemetryExtensions {
```
mousetraps: Thanks for refactoring this - makes it clearer what events we're logging.

Nodejs/Product/Nodejs/Telemetry/NodejsTelemetryExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+        public static void LogIntelliSenseActivatedForProject(this ITelemetryLogger logger, Guid projectGuid, Options.AnalysisLevel newAnalysisLevel) {
+            logger.ReportEvent(
+                TelemetryEvents.IntelliSenseActivatedForProject,
+                new DataPointCollection(
+                    new DataPoint(TelemetryProperties.ProjectGuid, projectGuid.ToString("B")),
+                    new DataPoint(TelemetryProperties.IntelliSenseLevel, newAnalysisLevel.ToString())));
+        }
```
mousetraps: The IntelliSense setting is a user setting, and I wouldn't expect it to differ between projects. Is there any reason we opt to log on a per-project basis rather than per instance of NodejsPackage, as we do with IntelliSenseLevelChanged

Nodejs/Product/Nodejs/Telemetry/NodejsTelemetryExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        public static void LogIntelliSenseLevelChanged(this ITelemetryLogger logger, Options.AnalysisLevel newAnalysisLevel) {
+            logger.ReportEvent(
+                TelemetryEvents.IntelliSenseLevelChanged,
+                TelemetryProperties.IntelliSenseLevel,
+                newAnalysisLevel.ToString());
+        }
```
mousetraps: More generally, how did you test this? Also once this is in, we should add our query to the dashboard so that we can verify that we're not missing anything important.

mousetraps:
Left a few comments and :+1: after that!

Nodejs/Product/Nodejs/Telemetry/NodejsTelemetryExtensions.cs
mjbvz: I agree that we should not see any different per-project at a specific point. What this point allows us to do is track intellisense changes by project or project session for correlation, and also more easily write queries that do something like: for 90% projects, users have ES6 enabled.

Logging per project allows all the same queries as if we were to log intellisense level per VS session, plus other interesting queries using project correlation.

Nodejs/Product/Nodejs/Telemetry/NodejsTelemetryExtensions.cs
mjbvz: I dropped our production telemetry dll in place with a NTVS debug build, and used the VS telemetry monitor to see the these showing up. 

I'll put together some graphs on this once this goes out.

Nodejs/Product/Nodejs/NodejsPackage.cs
mjbvz: Verified. The callback is only fired when the user goes in to the options page and changes the setting, not on first load.

Issue #1112, opened 2016-06-29
Investigate File > Open Folder experiences


mousetraps:
As we investigate the File > Open Folder experiences, we should keep in mind the following open Qs

**Extensibility**
- [ ] How do we register as an File > Open provider?

**Performance**
- [ ] Why does it take so long to open a folder?
- [ ] is there any way to mitigate this issue?

**Project System**
- [ ] how to determine File > Open Folder project context? 
- [ ] What areas does our current featureset depend on the existing project system? For each of these areas, how do we extend File > Open Folder to make it happen?
- [ ] How to determine Node.js project context?

**Editing / IntelliSense**
- [ ] how to supply Node.js-specific tsconfig.json w/o explictly adding it to the project?

**Telemetry**
- [ ] How to determine # Node.js projects?
- [ ] Any current telemetry?

**Debugging / Build**
- [ ] how do we set options / debug flags?
  - [ ] current options? 
  - [ ] any future options we should consider?
- [ ] msbuild as a "case study" - what would it take? 

**Roaming between VS / VSCode**
- [ ] what does that look like? 
- [ ] are there any existing efforts to bring them closer together?

paulvanbrenk:
implemented

PR #1120, opened 2016-07-01
Prototype: Port Open Folder Node Debug Provider to NTVS

mjbvz:
Issue #1112 

This change takes the existing node debugger provider embedded in the Open Folder team's codebase, and moves it into the NTVS code. This will allow us to maintain the debug provider going forward.

Testing
Tested on VS15. There seems to be a bug with the Open Folder where `ProviderPriority` is not respected, so the builtin copy of the node debug provider is always used over this new debug provider. I changed the file extension type to test this change, and saw this debug provider being used.


Nodejs/Product/Profiling/app.config
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
mjbvz: All these changes and files were generated by package manager when I added the new deps.

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -122,6 +122,22 @@
     <Reference Include="Microsoft.VisualStudio.TextManager.Interop.12.0, Version=12.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
       <EmbedInteropTypes>True</EmbedInteropTypes>
     </Reference>
+    <Reference Include="Microsoft.VisualStudio.Validation, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
```
mjbvz: These may have to be scoped to dev15 only.

mjbvz:
Looks like the nuget path provided ( https://mseng.pkgs.visualstudio.com/DefaultCollection/_packaging/VSIDEProj-TestSigned-Release/nuget/v3/index.json ) is Microsoft internal, which I believe is why AppVeyor failed. I'll reach out the the Open Folder team to confirm this.

Also, AppVeyor does not seem to be building dev15. Will see if I can add that.

mjbvz:
Opened #1121 to track dev15 AppVeyor work.

PR #1124, opened 2016-07-06
Fix Typings output Encoding

### LINKED ISSUE #1123: Typings Tool Output to VS Console is Garbled
With NTVS 1.2 RC, open a project that will download typings. The following is printed:

image

There seems to be an encoding issue with the output of the typings tool.

mjbvz:
**Bug**
The output from the typings command is sometimes garbled. This seems to be an encoding issue.

**Fix**
Force the output to be treated as utf-8. Here's the fixed output:

image

Closes #1123


mousetraps:
:+1:  - let's pull this in for RTM

PR #1126, opened 2016-07-06
Remove GlobalModulesNode

### LINKED ISSUE #1125: Don't Populate Global Modules Node In Project Explorer
We currently populate a npm node with a list of global modules for every NTVS project. For users with a few globals installed and/or lots of NTVS projects in a solution, this can lead to perf issues with both memory and CPU, since the global modules node includes the entire dependency tree of modules. Browsing global modules is also not a very common use case.

Simply removing the global modules node from the npm explorer would improve perf without significantly reducing functionality.

mjbvz:
##### Bug
We currently populate a npm node with a list of global modules for every NTVS project. For users with a few globals installed and/or lots of NTVS projects in a solution, this can lead to perf issues with both memory and CPU, since the global modules node includes the entire dependency tree of modules.

##### Fix
Simply remove this node and the entire `GlobalModulesNode` class

Closes #1125


Nodejs/Product/Nodejs/Project/DependencyNodeProperties.cs
```diff
@@ -143,14 +143,6 @@ public class DependencyNodeProperties : NodeProperties {
...
             get {
-                var node = DependencyNode as HierarchyNode;
-                while (null != node) {
-                    if (node is GlobalModulesNode) {
-                        return true;
-                    }
-
-                    node = node.Parent;
-                }
                 return false;
```
mjbvz: Since this expression is now constant, this creates a bunch of dead code paths that could be removed. I'm included to go the "you ain't gonna need it" route and delete these too, but wanted to check first

mousetraps: Agreed. We should remove it unless we can think of a particular case in which we would want to light up that codepath at some point in the future (personally, I can't think of a good reason given the typical usage of global modules.) My only concern is that those changes might be a tad more difficult to properly code review and I don't want to block on that for RTM. What do you think about adding a TODO for now so we can kick off a dev build, and then just follow it up with another PR?

Nodejs/Product/Nodejs/Project/GlobalModulesNode.cs
```diff
@@ -1,101 +0,0 @@
...
-//*********************************************************//
-
-using System;
-using System.Linq;
-using Microsoft.NodejsTools.Npm;
-using Microsoft.VisualStudio;
-using Microsoft.VisualStudioTools.Project;
-
-namespace Microsoft.NodejsTools.Project {
-    internal class GlobalModulesNode : AbstractNpmNode {
```
mousetraps: In addition to this, please remove usages of NpmBinCommand (which we use to locate the global modules directory) because that's been a source of completely un-reproducible perf issues and we've only ever been able to fix half the problem.

mjbvz: @mousetraps This sounds like a good idea. One concern though: the Npm management window currently depends on the global modules data and therefore on the NpmBinCommand:

image

Removing the NpmBinCommand means we should probably also remove the logic for supporting globals in the Npm window. Does that sound correct and do we want to do this? I agree that as it stands, this change only addresses one part of the global module memory usage problem.

mjbvz:
Merging this in since it addresses some of the memory usage problems. Following up with another PR that will fully remove the npm bin command and global support from the NPM window as discussed.

PR #1134, opened 2016-07-08
Remove npm bin Command and Supporting Global Package Logic

mjbvz:
Follow up on  #1126, which removed the NPM Global Modules node from the VS UI

##### Bug
Removing the Global Modules UI elements reduced memory usage, but we still build up an internal global modules hierarchy. The Npm bin command has also been a source of problems for us in the past.

##### Fix
This change removes all tracking of Global modules from NTVS. You can still install global modules using npm commands, but the Npm UI no longer supports installing globals modules.

This removes the need for the NPM bin command which in-turn eliminated the need for wide swaths of NTVS code related to global packages.

##### Testing
This change mainly effects the NPM UI, so I ran through common scenarios for that to make sure they continue to work. Most of the code paths touched are not hit in our normal flows, but I did a few sanity checks to make sure our main npm interaction points still work as expected.


mjbvz:
Looks like AppVeyor failed with "Error uploading artifact to Google Storage." at the end of the task. Build and tests seem to have passed thought

Nodejs/Product/Nodejs/NpmUI/NpmOutputViewModel.cs
```diff
@@ -385,30 +375,20 @@ private class QueuedNpmCommandInfo : EventArgs {
 
             public QueuedNpmCommandInfo(
                 string name,
-                string version) {
+                string version,
+                DependencyType depType) {
```
mousetraps: remove depType

Nodejs/Product/Npm/SPI/NpmCommander.cs
```diff
@@ -178,24 +174,16 @@ private class NmpCommandRunner : IDisposable {
             return type;
         }
 
-        private async Task<bool> UninstallPackageAsync(string packageName, bool global) {
-            return await DoCommandExecute(true,
+        public Task<bool> UninstallPackageAsync(string packageName) {
```
mousetraps: why public?

Nodejs/Product/Npm/SPI/NpmCommander.cs
```diff
@@ -211,23 +199,15 @@ private class NmpCommandRunner : IDisposable {
             return await UpdatePackagesAsync(new List<IPackage>());
         }
 
-        private async Task<bool> UpdatePackagesAsync(IEnumerable<IPackage> packages, bool global) {
-            return await DoCommandExecute(true,
+        public Task<bool> UpdatePackagesAsync(IEnumerable<IPackage> packages) {
+            return DoCommandExecute(true,
```
mousetraps: why modify the await behavior here?

Nodejs/Product/Npm/SPI/NpmCommander.cs
mjbvz: Git doesn't make it clear, but this is the result of merging the three `UninstallPackageAsync` functions.

First I deleted `public UninstallGlobalPackageAsync`. This left `private UninstallPackageAsync` function with a single caller that only forwards its arguments. So then I merged the `public UninstallPackageAsync` with the `private UninstallPackageAsync`.

Nodejs/Product/Npm/SPI/NpmController.cs
```diff
@@ -100,18 +95,19 @@ internal class NpmController : AbstractNpmLogSource, INpmController {
...
 #if DEBUG
                     Debug.Fail(ex.ToString());
 #endif
                 }
-            });
+            }
         }
 
-        public async Task RefreshAsync() {
+        private void RefreshImplementation() {
```
mousetraps: Ahh, I see the reason behind changing the behavior. I would prefer not to take this additional refactoring for this particular change, but rather in a separate PR

Nodejs/Tests/NpmTests/MaxPathTests.cs
```diff
@@ -1,73 +0,0 @@
...
-
-            Assert.IsNotNull(controller.GlobalPackages, "Cannot retrieve global packages after uninstall");
-            Assert.IsFalse(controller.GlobalPackages.Modules.Contains("yo"), "Global package failed to uninstall");
-        }
-
-        void controller_OutputLogged(object sender, NpmLogEventArgs e) {
-            Debug.WriteLine(e.LogText);
-        }
-    }
-}
```
mousetraps: Do we have any other MAX_PATH tests? If not, we should just change this to work for locally installed dependencies

Nodejs/Product/Npm/SPI/NpmCommander.cs
```diff
@@ -178,24 +174,16 @@ private class NmpCommandRunner : IDisposable {
             return type;
         }
 
-        private async Task<bool> UninstallPackageAsync(string packageName, bool global) {
```
mousetraps: we may want to consider keeping the global behavior in NpmCommander, as there's a clear enough abstraction. I don't have a strong preference either way.

Nodejs/Product/Npm/SPI/NpmCommander.cs
mjbvz: In my understanding, because we already have a task, a function with a single line return statement does not need to await because that essentially just rewraps the same task. I can revert these change if we want to be more cautious though.

Nodejs/Product/Npm/SPI/NpmController.cs
mjbvz: The main reason for this change was not refactoring, but because when I removed `NpmBinCommand`, `RefreshAsync` can no longer be an async method. It's a compiler error to mark it `async` since it does not have any async behavior.

I chose to rename this method and remove the async behavior, but an alternative could possibly be making `RefreshImplementation` return an async void or similar. What do you think?

Nodejs/Tests/NpmTests/MaxPathTests.cs
mjbvz: Good point. I've restored the second test and made it a local install instead.

Nodejs/Product/Npm/SPI/NpmController.cs
mousetraps: Ahh, gotcha - yeah, nevermind. Now that I take a closer look, let's leave it as is... the behavior should be exactly the same because an async method that returns void and doesn't have await operators would run synchronously anyways.

mousetraps:
:+1:

PR #1136, opened 2016-07-08
Generic Project Characteristics Telemetry

### LINKED ISSUE #1115: Add telemetry to detect project characteristics that help drive performance improvements
This would include general characteristics like # of files in project, file size, etc. that would help us drive perf improvements in the right areas.

mjbvz:
Issue #1115 

##### Bug
We would like to know understand general characteristics of the types of NTVS projects users work with. This will help us better test and tune NTVS for real world usage.

##### Fix
Add a telemetry point for uploading generic, non-identifying information about NTVS project characteristics. This currently uploads the number of files included in the project for a given file type (such as *.js or *.d.ts) Users can opt out of this by disabling visual studio telemetry.

closes #1115

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
@@ -570,10 +570,9 @@ namespace Microsoft.NodejsTools.Project {
         }
 
         /// <summary>
-        /// Log generic 
+        /// Log generic characteristics about the project, including number of files of each type in the project.
         /// </summary>
         private void LogProjectInformation() {
-            // Compute number of included files of each type in the project.
             var fileTypeInfo = new Dictionary<string, int>();
             foreach (var node in this.DiskNodes) {
                 if (node.Value?.ItemNode?.IsExcluded ?? true) {
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
@@ -564,6 +564,38 @@ namespace Microsoft.NodejsTools.Project {
                 // scan for files which were loaded from cached analysis but no longer
                 // exist and remove them.
                 _analyzer.ReloadComplete();
+
+                Task.Factory.StartNew(LogProjectInformation);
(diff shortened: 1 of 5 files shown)


Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -564,6 +564,38 @@ public NodejsProjectNode(NodejsProjectPackage package)
                 // scan for files which were loaded from cached analysis but no longer
                 // exist and remove them.
                 _analyzer.ReloadComplete();
+
+                Task.Factory.StartNew(LogProjectInformation);
+            }
+        }
+
+        /// <summary>
+        /// Log generic
```
mjbvz: Good comment 😄 
Will fill out the rest.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -564,6 +564,37 @@ public NodejsProjectNode(NodejsProjectPackage package)
...
+                Task.Factory.StartNew(LogProjectInformation);
+            }
+        }
+
+        /// <summary>
+        /// Log generic characteristics about the project, including number of files of each type in the project.
+        /// </summary>
+        private void LogProjectInformation() {
+            var fileTypeInfo = new Dictionary<string, int>();
+            foreach (var node in this.DiskNodes) {
```
mjbvz: Thinking about this more, I'm not confident that iterating here is safe. If the dictionary is changed on another thread, the iterator will be invalidated. I haven't been able to repo that in my testing, but it seems like it should be possible to trigger an exception from that.

Even moving this logic to the main thread may not be safe since it's not clear to me if any background threads could end up updating `DiskNodes`.

mjbvz:
I've thought about this change more and don't have confidence that it adds enough value for us to push it into RTM. I have concerns about thread safety, and this new code path will end up being hit every project open.

mousetraps:
Just took a look myself, and agreed.

mjbvz:
Closing this for now. We may decide to revisit this in the future but its not clear that this adds enough value given the risks of the change.

PR #1143, opened 2016-07-15
Create Wrapper Script for Installing Typings

### LINKED ISSUE #1140: Execute typings separately for each package
##### Expected Behavior 

Typings should be installed for all packages, not failing when some are missing.

##### Actual Behavior 

If Typings is provided multiple packages as command-line arguments, it fails and stops as soon as any of those is missing. Since many npm's do not have typings in the registry, it is quite common for it to fail and then it doesn't attempt to install the remaining listed packages.

- NTVS Version: 1.2.40627.01
- Visual Studio Version: 14.0.25422.01 Update 3
- Node.js Version: 6.3.0
- Typings Version: 1.0.5

##### Steps to Reproduce
1. Add an npm that has no typings info in the registry
2. Add an npm that does have typings info in the registry
3. Note that typings info for the former will not be installed, because it fails on the latter.

mjbvz:
Issue #1140
##### Bug
If a single package in the current project does not have a typings file, we fail to install any typings. This results in a poor IntelliSense experience.

##### Fix
Create a node script that wraps `typings-core`. This script will continue installation of all packages, even if some packages do not have typings files.

##### Testing
Tested on Node 6 so far with a few different scenarios, including projects where the typings for an installed package do not exist. Still have to test on earlier versions of node.

Closes #1140


Nodejs/Product/Nodejs/TypingsAquisitionTool/index.js
```diff
@@ -0,0 +1 @@
+
```
mjbvz: Seems that npm requires some sort of index/main file, so I just created this empty one.

mjbvz:
Did a basic test pass with Node 0.12. My main concern was whether local module installs would behave, but it seems to work just fine.

.gitignore
```diff
@@ -1,5 +1,4 @@
 # Build Folders (you can keep bin if you'd like, to store dlls and pdbs)
-[Bb]in/
```
mjbvz: These were ignoring the `bin` directory in the new local npm module. I thought it was better to remove them rather than override include the `bin` directory. No other new files seem to have been added from this

Nodejs/Product/Nodejs/TypingsAquisitionTool/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "ntvs-typings-aquisition-tool",
+  "author": "Microsoft",
+  "version": "1.0.0",
+  "description": "Internal tool that aquires typings to power Node Js Tools For Visual Studio IntelliSense",
+  "private": true,
+  "main": "index.js",
+  "dependencies": {
+    "minimist": "^1.2.0",
+    "typings-core": "^1.3.1"
```
mjbvz: Will lock down the version of typings-core for publish

.gitignore
mousetraps: I would prefer not to include bin by default, as this will likely be an issue if certain projects are built that aren't hooked up into our default project targets

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -735,6 +735,18 @@
     </ConditionalEmbeddedResource>
   </ItemGroup>
   <ItemGroup>
+    <Content Include="TypingsAquisitionTool\index.js">
```
mousetraps: spelling - Acquisition

Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -29,14 +29,30 @@
 
 namespace Microsoft.NodejsTools {
     internal class TypingsAcquisition {
-        private const string TypingsTool = "typings";
-        private const string TypingsToolVersion = "1.0.5"; // Lock to stable version of the 'typings' tool with known behavior.
```
mousetraps: we'll want to lock down to a specific version once we publish to npm

Nodejs/Product/Nodejs/TypingsAcquisition.cs
```diff
@@ -29,14 +29,30 @@
...
+            get {
+                string codeBase = System.Reflection.Assembly.GetExecutingAssembly().CodeBase;
+                var uri = new UriBuilder(codeBase);
+                var path = Uri.UnescapeDataString(uri.Path);
+                var root = Path.GetDirectoryName(path);
+
+                return Path.Combine(
+                    root,
+                    "TypingsAquisitionTool");
+            }
```
mousetraps: Any packages installed to this directory likely won't get cleaned up during uninstall, so as discussed, let's move to just publishing the package to npm

Nodejs/Product/Nodejs/TypingsAquisitionTool/bin/install_typings
```diff
@@ -0,0 +1,75 @@
...
+var typings = require('typings-core');
+var events = require('events');
+var minimist = require('minimist');
+
+/**
+ * Create a promise that runs one or more promises sequentially.
+ */
+var runAll = function (promises) {
+    return promises.reduce(function (p1, p2) {
+        return p1.then(function (_) {
```
mousetraps: why the underscore?

Nodejs/Product/Nodejs/TypingsAquisitionTool/bin/install_typings
```diff
@@ -0,0 +1,75 @@
...
+var options = {
+    save: argv.save,
+    global: true,
+    emitter: emitter,
+    cwd: argv.cwd || process.cwd()
+};
+
+var packagesToInstall = argv._;
+
+if (!packagesToInstall.length) { // top level package install
```
mousetraps: nit: comment on new line

Nodejs/Product/Nodejs/TypingsAquisitionTool/index.js
mousetraps: As discussed, let's move the core logic into index.js

Nodejs/Product/Nodejs/TypingsAquisitionTool/package.json
mousetraps: and minimist too

Nodejs/Product/Nodejs/TypingsAquisitionTool/package.json
```diff
@@ -0,0 +1,15 @@
...
+  },
```
mousetraps: licensing?

Nodejs/Product/Nodejs/TypingsAquisitionTool/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "ntvs-typings-aquisition-tool",
+  "author": "Microsoft",
+  "version": "1.0.0",
+  "description": "Internal tool that aquires typings to power Node Js Tools For Visual Studio IntelliSense",
```
mousetraps: nit: Node Js -> Node.js

.gitignore
mjbvz: I'll just exclude top level /bin dirs then? Or revert the gitignore  change and force add the node bin files?

Nodejs/Product/Nodejs/TypingsAcquisition.cs
mjbvz: Note that this is how typings is already installed so moving this new package to npm makes no difference, we're just installing from a local source now. The tool is installed alongside the sqllite npm cache though, so if the uninstaller handles cleaning up directory, we probably should also get this cleaned up

Nodejs/Product/Nodejs/TypingsAquisitionTool/package.json
mjbvz: What should it be? Apache?

Nodejs/Product/Nodejs/TypingsAquisitionTool/bin/install_typings
mjbvz: My scala is showing there. It's helpful to indicate that a parameter is ignored sometimes.

Nodejs/Product/Nodejs/TypingsAcquisitionTool/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "ntvs-typings-acquisition-tool",
+  "author": "Microsoft",
+  "version": "1.0.0",
+  "description": "Internal tool that aquires typings to power Node.js Tools For Visual Studio IntelliSense",
+  "license": "Apache 2",
```
mousetraps: Apache 2.0

mousetraps:
:+1:

PR #1149, opened 2016-07-19
Update scripts/node.d.ts

mjbvz:
**Bug**
The current scripts/node/node.d.ts file we use for typescript projects is out of date.

**Fix**
Copy in the latest node.d.ts file.

**Testing**
Created new typescript project. Completions now match expected node 6 apis.


mjbvz:
Merged this in since this was just a update to a typings file. I'll take this into v1.2.x too

PR #1193, opened 2016-08-04
Fix breaking on plain old errors

### LINKED ISSUE #1192: throw new Error('...') exception does not get handled by VS while debugging
##### Expected Behavior 
When node.js exceptions are enabled in the VS debugger, the exception should be caught and displayed to the user.

##### Actual Behavior 
Error is not displayed 

##### Steps to Reproduce
1. Create new js console app.
2. Enable  node.js break on exceptions
3.  In `app.js`, add
```js
throw new Error('I am');
```

Interestingly, builtin subclasses of Error seem to work fine, as do non-error objects:

```js
// these all work
throw 'I am error';
throw {};
throw new EvalError('eval')
```

But this does not work:

```js
function SubError() { }
SubError.prototype = new Error

throw new SubError()
```

mjbvz:
closes #1192

**bug**
```js
throw new Error('I am');
```
is not handled, even with all node.js exceptions enabled

**fix**
Use correct name when enabling or disabling errors.

**testing**
Tested this case before and after fix. See error being handled correctly now.

image


Nodejs/Product/Nodejs/Debugger/ExceptionHandler.cs
```diff
@@ -84,117 +84,14 @@ sealed class ExceptionHandler {
...
-#endif
 
-            foreach (string exceptionType in breakNeverTypes) {
-                defaultExceptionTreatments[exceptionType] = ExceptionHitTreatment.BreakNever;
+            // Get exception names from in NodePackage.Debugger.cs
+            foreach (var attr in System.Attribute.GetCustomAttributes(typeof(NodejsPackage))) {
+                var debugAttr = attr as ProvideNodeDebugExceptionAttribute;
+                if (debugAttr != null && !string.IsNullOrEmpty(debugAttr.ExceptionName)) {
+                    defaultExceptionTreatments[debugAttr.ExceptionName] = ExceptionHitTreatment.BreakNever;
+                }
```
mjbvz: Rather than risk having this happen again, I refactored this code to remove the duplication. The core fix for the original problem was just the mismatch between using `Error()` one place and `Error` the other.

PR #1196, opened 2016-08-08
Use Signed Binary Path For WebRole.Dll in Lightweight install

mjbvz:
##### Bug
For VS15 lightweight install, we currently deploy an unsigned copy of the webrole.dll.

##### Fix
Switch to using a signed version of the dll from `SignedBinariesPath` when it is available. 

Drop: https://vsdrop.corp.microsoft.com/file/v1/Products/DevDiv/Microsoft/nodejstools/lightweight-install/20160808.3;NodejsTools.vsman


mousetraps:
:+1:

Nodejs/Setup/swix/Microsoft.VisualStudio.NodejsTools.Targets.swixproj
```diff
@@ -31,7 +31,9 @@
   </Target>
 
   <PropertyGroup> 
-    <PackagePreprocessorDefinitions>$(PackagePreprocessorDefinitions);BuildOutputRoot=$(BuildOutputRoot)</PackagePreprocessorDefinitions>
+    <WebRoleDllPath Condition="'$(SignedBinariesPath)' != ''">$(SignedBinariesPath)</WebRoleDllPath>
+    <WebRoleDllPath Condition="'$(SignedBinariesPath)' == ''">$(BuildOutputRoot)\Binaries\Nodejs\Debugger\RemoteDebug</WebRoleDllPath>
```
mousetraps: nit: including an extra backlash in folder definitions is helpful so that it's easier to append paths

Nodejs/Setup/swix/Microsoft.VisualStudio.NodejsTools.Targets.swixproj
```diff
@@ -31,7 +31,9 @@
   </Target>
```
mousetraps: no need to include it for this change, but think we take care of the vsix copy in BuildRelease, so this is no longer necessary

PR #1197, opened 2016-08-09
Fixed an issue where certain tests would be skipped since Mocha 3.0.0.

Andreyul:
### Steps to reproduce:
Create a new 'Blank Node.js Console Application' project.
Install npm package Mocha 3.0.2.
Add a test:
```
const assert = require('assert');

it('+ ', function () {
    assert(false);
});
```
Run this test.

### Expected:
Test should fail.

### Actual result:
Test is skipped. Output:
```
Test Name:  +
Test Outcome:   Passed
Result StandardOutput:  
1..0
# tests 0
# pass 0
# fail 0
```

### Explanation:
Since Mocha 3.0.0 method mocha.grep() accepts a regexp string. Before Mocha 3.0.0 string passed to mocha.grep() was escaped. Method mocha.fgrep() appeared in Mocha 3.0.0, it behaves exactly as mocha.grep() in Mocha 2.x.x.

### Fix
Check that mocha.fgrep() method exists, then call it. Otherwise call mocha.grep().

### DIFF
--- a/Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
+++ b/Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
@@ -65,7 +65,10 @@ var run_tests = function (testName, testFile, workingFolder, projectFolder) {
     var mocha = initializeMocha(Mocha, projectFolder);
 
     if (testName) {
-        mocha.grep(testName);
+        if (typeof mocha.fgrep === 'function')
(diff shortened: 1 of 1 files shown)


mjbvz:
@Andreyul Thanks for the contribution. This looks like the correct fix so I've gone ahead and merged it in. Look for the change in the next dev build. We'll also probably want to take this into the next, minor 1.2 release

PR #1201, opened 2016-08-09
Remove pkgdef template references

### LINKED ISSUE #1177: [VS15 lightweight] Duplicate Template Entries in the template tree
image

The root cause is that our template projects are usually deployed as zip files, but with Willow install, these are now unzipped on install. VS picks up the vstman files in each of these templates, creating the duplicate entries. Fixing this in a way that does not break normal installs or dev testing has been challenging. 

[Tracked internally as 240934](https://devdiv.visualstudio.com/DevDiv/_workitems?id=240934)

mjbvz:
**Bug**
Duplicate template entries showing up in VS15 #1177

**Fix**
The root cause seems to be our `*.pkgdef` file, which sets `ProjectTemplatesDir`. Set this to nullpath instead of the actual path. Also updated the dev pkgdef to point to the correct location for testing.

**Testing**
Tested VS14 dev build. Create a new VS15 drop and try the change out there. Manually copying over the pkgdef to a VS15 instance seemed to work fine:

image

closes #1177


mjbvz:
I'll kick off another dev15 build for NTVS and submit a PR against lab/bcb2 for testing.

PR #1203, opened 2016-08-12
Lightweight remote debug

### LINKED ISSUE #1179: [VS15 lightweight] Cannot Find RemoteDebugProxy folder
##### Repo
On VS 15 with lightweight install:

Go to `Tools -> Node.js Tool -> Remote Debug Proxy -> Open Containing Folder`

##### Expected
Opens folder

##### Actual
Message about how the folder cannot be found.

mjbvz:
Issue #1179 

##### Bug
With VS15 lightweight install, we do not include the RemoteDebug.js script. Also, we use the registry to find this script.

##### Fix
Copy `RemoteDebug.js` into the vsix at the same location as the msi. Also update the location of the file so that dev installs match the msi installs on dev14. 

Also removes the logic for getting `RemoteDebug.js` from the registry. I think we should treat this file like the other test adapter javascript files, and always get it from the extension.

##### Testing
Tested for regressions on dev14. Also tested to ensure the file exists on dev15. I still need to investigate if we also need to copy `webrole.dll` into the remotedebug file location.

Closes  #1179


Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -901,8 +901,9 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
-    <Content Include="Debugger\RemoteDebug\RemoteDebug.js">
+    <Content Include="RemoteDebug\RemoteDebug.js">
```
mousetraps: This makes more sense to exist in Debugger\*, so would prefer to set the VSIXPath than change the location on disk.

mousetraps:
Minor comment and :+1: after that

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -1684,27 +1685,18 @@
   </Target>
   <PropertyGroup>
     <PostBuildEvent>
-      copy /y $(OutDir)..\WebRole\Microsoft.NodejsTools.WebRole.dll $(OutDir)Debugger\RemoteDebug\
+      copy /y $(OutDir)..\WebRole\Microsoft.NodejsTools.WebRole.dll $(OutDir)RemoteDebug\
```
mousetraps: This won't copy the assembly to the VSIX (necessary for git deploy scenarios, where the file must be included in the project manually)

Nodejs/Product/Nodejs/Nodejs.csproj
mjbvz: The msi installs the files to just `RemoteDebug`. Should I change the msi instead?

Nodejs/Product/Nodejs/Nodejs.csproj
mjbvz: Yes, this change doesn't solve the missing webrole.dll problem yet. I'm still trying to figure out the best way to address it but may take that in with another change.

mjbvz:
Merging this in. It improves the remotedebug scenario but we still need to make sure webrole.dll is properly deployed in the remotedebug folder as well.

PR #1227, opened 2016-08-23
Delete Man Field from Npm package.json Parser

mjbvz:
The `Man` field for `package.json` files is only consumed by our tests, and therefore not very useful. Just remove the field and delete the tests that were testing it.


mousetraps:
:+1:

PR #1228, opened 2016-08-23
Remove unused functions in commonutils.cs

mjbvz:
These functions are not referenced anywhere in NTVS. Just deleting them.


mousetraps:
:-1: These are in `Common/*`, which means that they might break PTVS or other dependents of shared project if we remove them

mjbvz:
Missed that.  This file is probably the lowest priority of unused functionality from common :)

PR #1236, opened 2016-08-29
Fix Building NTVS with devenv /Build

### LINKED ISSUE #1230: Package failed to load in devenv
##### Expected Behavior

When I run my nodejs project build from devenv the project is loaded and built.
##### Actual Behavior

When I run my nodejs project build from **devenv** the 

> Package 'Node.js Project Package' failed to load. 

The project can be built when I use the Visual Studio GUI.
- NTVS Version: 1.2.40726.00
- Visual Studio Version: 14.0.25425.01 Update 3
- Node.js Version: v5.6.0
##### Steps to Reproduce
1. Create new **Blank Node.js Console Application** in Visual Studio
2. Build project successfully in Visual Studio GUI
3. Open **cmd** and change to solution dir.
4. See **devenv** failing with: `"C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\devenv.exe" NodejsConsoleApp1.sln /Build Debug /Out log.txt /Log activity.log`
5. Inspect **log.txt** and created **UpgradeLog.htm**

mjbvz:
Issue #1230

**Bug**
Crash when trying to build a NTVS project from the command line with `devenv /build`. The root cause is in the `UIThread` where we attempt to call `TaskScheduler.FromCurrentSynchronizationContext`. In non-ui flows, this throws an `InvalidOperationException`.

**Fix**
If `TaskScheduler.FromCurrentSynchronizationContext` fails, try setting the current syntronization context manually and retrying the call. This fixes the `UIThread`, but also fixes all other code paths (including some in the VS extension bases themseleves) that use `TaskScheduler.FromCurrentSynchronizationContext` or assume a `SyncronizationContext`  exists.

closes #1230

### DIFF
--- a/Common/Product/SharedProject/UIThread.cs
+++ b/Common/Product/SharedProject/UIThread.cs
@@ -26,12 +26,10 @@ namespace Microsoft.VisualStudioTools {
         private readonly Thread _uiThread;
 
         private UIThread() {
-            try {
-                _scheduler = TaskScheduler.FromCurrentSynchronizationContext();
-            } catch (InvalidOperationException) {
+            if (SynchronizationContext.Current == null) {
                 SynchronizationContext.SetSynchronizationContext(new SynchronizationContext());
-                _scheduler = TaskScheduler.FromCurrentSynchronizationContext();
             }
+            _scheduler = TaskScheduler.FromCurrentSynchronizationContext();
             _factory = new TaskFactory(_scheduler);
             _uiThread = Thread.CurrentThread;
         }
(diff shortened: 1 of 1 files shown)


Common/Product/SharedProject/UIThread.cs
```diff
@@ -26,7 +26,12 @@ class UIThread : UIThreadBase {
         private readonly Thread _uiThread;
 
         private UIThread() {
-            _scheduler = TaskScheduler.FromCurrentSynchronizationContext();
```
mjbvz: I believe we could instead try checking `SyncronizationContext.Current` hear instead of catching an exception from `TaskScheduler`. That may be more clear and scope the fix better.

PR #1253, opened 2016-09-08
Add SendFeedback menu item for NTVS

### LINKED ISSUE #1252: Add VS Menu Option to Send NTVS Feedback/Open Github Issue
#### Problem
If a user runs into a problem with NTVS or has an idea for improvement, it may not be clear how to let us know this. This will be even more true in VS15, where NTVS will be installed through the VS installer.

We already have the VS-wide send feedback mechanism, but GitHub is our preferred feedback channel

#### Proposal
Under `Tools -> Node.js Tools`, add a menu option for sending feedback or opening an issue on GitHub directly. This is similar to what VS code does in their `Help menu`. We could also look into getting into the VS `Help` menu instead, but I'm not sure if this is possible.

mjbvz:
Adds a simple menu item to send feedback about ntvs. See #1252 for details

image

Closes #1252


mjbvz:
May want to only enable this inside of node projects or move it back into the `Tools -> Node.js Tools` menu group instead.

Nodejs/Product/Nodejs/Commands/SendFeedbackCommand.cs
```diff
@@ -0,0 +1,32 @@
...
+//    distributed under the License is distributed on an "AS IS" BASIS, 
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or 
+//    implied. See the License for the specific language governing 
+//    permissions and limitations under the License.
+//
+//*********************************************************//
+
+using System;
+using Microsoft.VisualStudioTools;
+using System.Diagnostics;
```
mousetraps: nit:sort usings

Nodejs/Product/Nodejs/Commands/SendFeedbackCommand.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Microsoft.NodejsTools.Commands {
+    internal sealed class SendFeedbackCommand : Command {
```
mousetraps: Also worth adding an Open Documentation... command

Nodejs/Product/Nodejs/Commands/SendFeedbackCommand.cs
```diff
@@ -0,0 +1,32 @@
...
+        public override void DoCommand(object sender, EventArgs args) {
+            var uri = new Uri(@"https://aka.ms/ntvs-feedback");
```
mousetraps: fwlink instead of aka.ms - aka.ms is only for vanity urls / social media, which isn't necessary for the product. The nice thing is that the aka.ms tool now lets you easily switch the domain create fwlinks (which was the other reason we might use it in-product :smiley:)

Nodejs/Product/Nodejs/Commands/SendFeedbackCommand.cs
```diff
@@ -0,0 +1,32 @@
...
+            Process.Start(new ProcessStartInfo(uri.AbsoluteUri));
```
mousetraps: Process.Start("https://aka.ms/ntvs-feedback") is sufficient.

Nodejs/Product/Nodejs/NodejsTools.vsct
```diff
@@ -91,6 +91,13 @@
               <CommandFlag>DynamicVisibility</CommandFlag>
           If you do not want an image next to your command, remove the Icon node or set it to <Icon guid="guidOfficeIcon" id="msotcidNoIcon" /> -->
 
+       <Button guid="guidNodeToolsCmdSet" id="cmdidSendFeedback" priority="0x0500" type="Button">
+        <Parent guid="guidSHLMainMenu" id="IDG_VS_HELP_SUPPORT"/>
```
mousetraps: This should be in Tools > Node.js Tools - we don't like other ppl cluttering up common menus, so we shouldn't do the same. Additonally, when you do that, you'll also want to add a menu group to separate diagnostic info, survey/news, and send feedback

Nodejs/Product/Nodejs/NodejsTools.vsct
```diff
@@ -91,6 +91,13 @@
...
+        <Strings>
+          <ButtonText>Report Node.js Tools Issue or Send Feedback</ButtonText>
```
mousetraps: a couple things:
1. text should be shorter maybe even just "Report Issue or Send Feedback" (most products actually split it into two items "Report Issue" and "Request Feature", so I have a hunch that that's more effective for some reason, but I don't know from experience)
2. Add a "..." at the end - this signifies that the action does not "complete" once you click, and is consistent with our UI paradigms throughout VS

Nodejs/Product/Nodejs/NodejsTools.vsct
mjbvz: Here's how it now looks:

image

I also created a documentation command.

PR #1255, opened 2016-09-08
Remove ES5 IntelliSense Option

### LINKED ISSUE #1254: Remove ES5 IntelliSense Option For NTVS 1.3
**Problem**
Now that NTVS 1.2 has been out for a few months, we have not seen many users moving back to the old ES5 IntelliSense. However, we still support switching IntelliSense levels in the NTVS options. There are few reasons to ever do this.

**Proposal**
I propose we eliminate the option to switch the intellisense back to es5 entirely. Es5 IntelliSense is buggy, unstable, and does not support modern JavaScript. 

Longer term, we will also eliminate this ES5 IntelliSense code, but it still currently powers the Interactive window.

mjbvz:
As explained in #1254, there are few reasons for anyone to switch back to the old ES5 based intellisense. Our telemtry also shows that very very few people are switching back too. We want everyone on the modern, more well supported ES6/Typscript backed Intellisense.

This change removes the option to switch from the VS entirely. This locks all main editors in ES6/Typescript backed mode. We'll keep around the old analysis engine code a bit longer in order to power the Interactive window, but will eventually rip it out too.

Closes #1254


mjbvz:
This may have created other dead code paths that have not been pruned yet. I'll make quick sweep through the codebase to check and run the code analysis tools to catch some simple cases.

Nodejs/Product/Nodejs/Commands/DiagnosticsCommand.cs
```diff
@@ -24,6 +24,7 @@
 using Microsoft.NodejsTools.Logging;
 using Microsoft.VisualStudioTools;
 using Microsoft.VisualStudioTools.Project;
+using Microsoft.NodejsTools.Options;
```
mousetraps: nit:sort usings

Nodejs/Product/Nodejs/Commands/DiagnosticsCommand.cs
```diff
@@ -285,8 +286,6 @@ private class FileTypeInfo {
                     jsAnalyzer.DumpLog(writer);
                 }
             }
-
-            res.AppendLine(string.Format("IntelliSense Completion Only Tab or Enter to Commit: {0}", NodejsPackage.Instance.IntellisenseOptionsPage.OnlyTabOrEnterToCommit));
```
mousetraps: There's a bunch of other logic about the analyzer in this file that we should also remove unless it's also logging the IW window (or there's an easy way to make it do so).

Nodejs/Product/Nodejs/Intellisense/IntellisenseController.cs
```diff
@@ -648,8 +644,7 @@ internal sealed class IntellisenseController : IIntellisenseController, IOleComm
                         case VSConstants.VSStd2KCmdID.COMPLETEWORD:
                             ForceCompletions = true;
                             try {
-                                TriggerCompletionSession((VSConstants.VSStd2KCmdID)nCmdID == VSConstants.VSStd2KCmdID.COMPLETEWORD
-                                    && !NodejsPackage.Instance.IntellisenseOptionsPage.OnlyTabOrEnterToCommit);
+                                TriggerCompletionSession(false);
```
mousetraps: still should check for VSConstants.VSSt2KCmdID.COMPLETEWORD

Nodejs/Product/Nodejs/Intellisense/VsProjectAnalyzer.cs
```diff
@@ -1075,41 +1075,6 @@ CollectingErrorSink errorSink
         ) {
             // There are some scenarios during ES6 Mode where a textview is still opened using NodeLS.
             // In these cases, we do not properly parse ES6 syntax, and therefore do not want to display errors. 
-            if (NodejsPackage.Instance.IntellisenseOptionsPage.AnalysisLevel == AnalysisLevel.Preview) {
-                return;
-            }
```
mousetraps: nit: I would prefer to remove just this, and remove everything below in a separate, more focused change, when we remove the analyzer altogether

Nodejs/Product/Nodejs/NodejsProject.cs
```diff
@@ -544,7 +544,7 @@ out project
             var properties = GetExtensionObject(_innerVsHierarchy, selectionItemId).Properties;
             try {
                 var itemType = properties.Item("ItemType").Value;
-                ourEditor = (string)itemType == ProjectFileConstants.Compile && !IsES6IntellisensePreview() ? Guids.NodejsEditorFactory : Guid.Empty;
+                ourEditor = Guid.Empty;
             } catch (ArgumentException) {
                 // no item type, file is excluded from project.
                 ourEditor = Guids.NodejsEditorFactory;
```
mousetraps: Also remove the itemType assignment, otherwise we might still fall back to NodejsEditorFactory. With these changes, it may also be helpful to place a TODO or track in some other way because much of this logic is actually unnecessary since we are always using the default editor

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.resx
```diff
@@ -520,10 +298,10 @@
     <value>GrowAndShrink</value>
   </data>
   <data name="$this.Margin" type="System.Windows.Forms.Padding, System.Windows.Forms">
-    <value>6, 6, 0, 6</value>
+    <value>3, 3, 0, 3</value>
   </data>
   <data name="$this.Size" type="System.Drawing.Size, System.Drawing">
-    <value>735, 745</value>
+    <value>368, 260</value>
```
mousetraps: it's a little hard to visualize these changes - do you have a screenshot of the new UI?

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsPage.cs
```diff
@@ -25,20 +25,7 @@
 namespace Microsoft.NodejsTools.Options {
     [ComVisible(true)]
     public class NodejsIntellisenseOptionsPage : NodejsDialogPage {
-        private static readonly Version _typeScriptMinVersionForES6Preview = new Version("1.6");
```
mousetraps: So this tells me that we are *only* targeting Dev15 from here on out. Is that correct? If so, we need to balance that with the ability to get feedback on dev builds, so we should block in the installer, or provide a link to install the latest TS bits if it doesn't exist on startup

Nodejs/Product/Nodejs/Project/NodejsFolderNode.cs
```diff
@@ -100,14 +100,7 @@ class NodejsFolderNode : CommonFolderNode {
 
         public override string Caption {
             get {
-                var res = base.Caption;
-
-                if (NodejsPackage.Instance.GeneralOptionsPage.ShowBrowserAndNodeLabels &&
-                    NodejsPackage.Instance.IntellisenseOptionsPage.AnalysisLevel != AnalysisLevel.Preview &&
-                    _containsNodeOrBrowserFiles) {
-                    res = AppendLabel(res, ContentType);
```
mousetraps: You should also be able to remove the `UpdateContentType` method and other related logic, as that only applies to the old analyzer. This should also result in a nice perf boost :smiley:

Nodejs/Product/Nodejs/Telemetry/NodejsTelemetryExtensions.cs
```diff
@@ -29,13 +29,6 @@ internal static class NodejsTelemetryExtensions {
                     new DataPoint(TelemetryProperties.AnalysisLevel, newAnalysisLevel.ToString())));
         }
 
-        public static void LogAnalysisLevelChanged(this ITelemetryLogger logger, Options.AnalysisLevel newAnalysisLevel) {
-            logger.ReportEvent(
-                TelemetryEvents.AnalysisLevelChanged,
```
mousetraps: I don't see a corresponding change in TelemetryEvents. Is this intentional?

Nodejs/Product/Nodejs/NodejsPackage.cs
```diff
@@ -93,7 +93,9 @@ namespace Microsoft.NodejsTools {
     [ProvideLanguageEditorOptionPage(typeof(NodejsFormattingSpacingOptionsPage), NodejsConstants.Nodejs, "Formatting", "Spacing", "3042")]
     [ProvideLanguageEditorOptionPage(typeof(NodejsFormattingBracesOptionsPage), NodejsConstants.Nodejs, "Formatting", "Braces", "3043")]
     [ProvideLanguageEditorOptionPage(typeof(NodejsFormattingGeneralOptionsPage), NodejsConstants.Nodejs, "Formatting", "General", "3044")]
```
mousetraps: Eventually we should remove Node.js formatting options pages as well when TS uses JS, rather than TS to back their editor. Worth adding an item to track.

Nodejs/Product/Nodejs/Intellisense/IntellisenseController.cs
mjbvz: `OnlyTabOrEnterToCommit` defaults to true in my understanding, which would make this `(VSConstants.VSStd2KCmdID)nCmdID == VSConstants.VSStd2KCmdID.COMPLETEWORD && !true`. Is that not correct?

Nodejs/Product/Nodejs/Intellisense/VsProjectAnalyzer.cs
mjbvz: Do you mean remove the entire If block? This code is unreachable after this change if we keep the current logic.

Nodejs/Product/Nodejs/NodejsPackage.cs
mjbvz: Opened #1258 to track this.

Nodejs/Product/Nodejs/NodejsProject.cs
mjbvz: I've got another change that builds on this one to remove the node editors entirely (you can still reach them using `Open With`). I'll make sure this function is updated as part of that work.

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsPage.cs
mjbvz: We're only targetting people with TS installed, both VS2015 and VS15. Based on our telemetry, this is the vast majority of users (we don't actually have a breakdown of who is using the old editor because they don't have TS vs. they switched back manually, but the entire set of 1.2 users on our old IntelliSense engine is miniscule)

I agree we should provide a better experience so that people know they need TS thought. Right now, we just will prompt them on launch in `GetNodejsLanguagePreferencesFromTypeScript`. This does not check TS version however

Nodejs/Product/Nodejs/Options/NodejsIntellisenseOptionsControl.resx
mjbvz: Here's VS14:

image

In VS15, the page goes away entirely.

Nodejs/Product/Nodejs/Telemetry/NodejsTelemetryExtensions.cs
mjbvz: Good catch, will remove the now unused constants

Nodejs/Product/Nodejs/Intellisense/IntellisenseController.cs
mousetraps: yep, that's actually correct - parsed it wrong at first :smiley:

mjbvz:
Following up with some more targeted PRs that will address the remaining comments.

Issue #1271, opened 2016-09-12
Express Template Modernization


mjbvz:
This tracks work to modernize the Express Templates included in NTVS. The following templates are targeted:

* AzureExpress4App
* Express4App
* TypeScriptAzureExpressApp
* TypeScriptExpressApp

Planned work consists of the following:

- [x] Update package.json versions. Switch to using `^` for version numbers where possible so we don't default users in old versions unexpectedly
- [x] Update Jade to pug #895 
- [x] Switch `styl` to plain old css. The templates should make few assumptions about intended use and `styl` is not even one of the more popular style sheet languages. And a new user is more likely to have experience with css. And we do not provide any syntax highlighting or real assistance for editing styl files, but web tools does provide this for css files.
- [x] Update TypeScript express projects to Express 4.

Please let me know if you have any suggestions or feedback on this plan.

mjbvz:
All templates should be updated now. You can try the JavaScript templates out in the latest devbuild. The TypeScript template changes will get picked up by the next devbuild

PR #1275, opened 2016-09-12
Switch Jade to Pug in Express Templates

### LINKED ISSUE #895: Switch Jade to Pug
Jade had been renamed to Pug for legal reasons, so when a user creates an express project using our NTVS express templates, they end up seeing a warning about switching to use Pug instead of Jade.

We should:
- [ ] Update the project templates that use the latest version of Jade to use Pug instead. I think all of our templates use `"jade": "*"`
- [ ] Update any user facing strings.
- [ ] Update our documentation to make all this clear. 

Not sure if the `*.jade` file extensions should be changed too. It's all rather confusing.

mjbvz:
Part of #1271 

**Bug**
Jade has been renamed to pug. Our templates still install Jade and this generates a warning.

**Fix**
Switch from Jade to Pug. Rename all template file extensions and update express to use pug instead of jade. NTVS 1.3 already supports viewing *.pug files as if they were *.jade files.

closes #895


Nodejs/Product/Nodejs/ProjectTemplates/AzureExpress4App/typings.json
```diff
@@ -5,7 +5,6 @@
     "debug": "registry:dt/debug#0.0.0+20160317120654",
     "express": "registry:dt/express#4.0.0+20160317120654",
     "express-serve-static-core": "registry:dt/express-serve-static-core#0.0.0+20160602151406",
-    "jade": "registry:dt/jade#0.0.0+20160316155526",
```
mjbvz: This was removed because the pug type definitions only exist in the npm registry, not in dt. Multiple registries (and ones that are not `--global`) are not something our super simplified typings acquisition system can handle. VS15 should handle typings better.

For the majority of use cases, users are not invoking pug directly but just using it through express.

PR #1309, opened 2016-10-04
Fixes for Scoped NPM Modules

mjbvz:
**Bug**
There are a few bugs around scoped modules currently:
- If a scope module is listed in the package.json but not installed, the NPM node shows it as "not in package.json".
- In the same situation, only the module name itself is shown, not the scope plus name
- We may try to acquire typings for scoped modules. These do not exist and are not supported by typings.

**Fix**
Add a check to filter out scoped modules before attempting to install typings for them.

Thread the entire name of the module through the npm structures instead of just the package path. Use the name instead of the last element of the path for the UI.


mjbvz:
Not a fan of this work around in its current state. It addresses the problem but I'll see if I can clean it up more.

paulvanbrenk:
#1707 replaces this.

PR #1322, opened 2016-10-08
Fix build project in VS 15 Preview 5

### LINKED ISSUE #1321: Building broken in Preview
##### Expected Behavior 
The build success, and tests showing in the Test Explorer

##### Actual Behavior 
An exception occurred while test discoverer 'TestDiscoverer' was loading tests. Exception: The imported project "C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v15.0\WebApplications\Microsoft.WebApplication.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.  C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v15.0\Node.js Tools\Microsoft.NodejsTools.targets

- NTVS Version: Shipped with VS 15 Preview 5
- Visual Studio Version: VS 15 Preview 5
- Node.js Version: 

##### Steps to Reproduce
1. Create new project
2. Add 'Javascript Mocha Unit Test'
3. Press 'Run All' in Test Explorer

kant2002:
Issue #1321 

##### Bug
Test Explorer does not show the list of the tests

##### Fix
Don't expect that WebApplication.targets file present on the user machine

##### Testing

This should fix #1321, which cause Test Explorer be unusabe. Right now NTVS has dependency on WebApplications which not always the case with New installer Experience

### DIFF
--- a/Nodejs/Product/Nodejs/Microsoft.NodejsTools.targets
+++ b/Nodejs/Product/Nodejs/Microsoft.NodejsTools.targets
@@ -168,7 +168,7 @@
     <PublishPipelineCollectFilesCore>$(PublishPipelineCollectFilesCore);CollectNodejsFiles</PublishPipelineCollectFilesCore>
   </PropertyGroup>
 
-  <Import Project="$(VSToolsPath)\WebApplications\Microsoft.WebApplication.targets" Condition="'$(VSToolsPath)' != ''" />
+  <Import Project="$(VSToolsPath)\WebApplications\Microsoft.WebApplication.targets" Condition="'$(VSToolsPath)' != '' and Exists('$(VSToolsPath)\WebApplications\Microsoft.WebApplication.targets')" />
 
   <!--- Detects if we're deploying via web deploy, and sets _NodejsDeploymentMethod appropriately.
         Otherwise we'll assume we're deploying to cloud service. -->


mjbvz:
Thank you for looking into this. I believe this fix address a symptom of the problem, but the root cause may have already been addressed. 

In the current VS15 preview, `VSToolsPath` resolves to: `C:\Program Files (x86)\MSBuild\...` when an NTVS project is loaded in the test explorer. That's the old global msbuild path. With VS15, the webapplications targets file is instead installed to: `C:\Program Files (x86)\Microsoft Visual Studio\VS15Preview\MSBuild\...`

I recently made a fix for the test explorer (https://github.com/Microsoft/nodejstools/commit/2fd265812bcc01395eb90d74e376a8cdfd4daaa3#diff-65a0dc1fe91fc286ca7f115c2a094e77) that addresses `VSToolsPath` being set to the wrong location. This fix should be in VS15 RC.

Can you check on your machine and let me know if `Microsoft.WebApplication.targets` exists inside of the Visual Studio MSBuild directory? 

Thanks again.

kant2002:
@mjbvz  The file "C:\Program Files (x86)\Microsoft Visual Studio\VS15Preview\MSBuild\Microsoft\VisualStudio\v15.0\WebApplications\Microsoft.WebApplication.targets" is present if I understand correctly where it should goes. Please correct me if I should look in the other location. If this is correct location, then you could close this issue

mjbvz:
@kant2002 Yes, I think that file should be installed in dev15, but you'll need to wait for VS15 RC to pick up the NTVS fix that ensures we look in the right location.

I'm going to close this PR since the other fix should handle this.

Thanks.

PR #1334, opened 2016-10-12
modifications to mocha.js and some TestExecutor logic

ozyx:
Hi again.
This is kind of a big, messy PR but I'd like to get your feedback on the changes before I continue further. I've moved the logic for determining test pass/failure and sending results to `mocha.js`. It is much easier to determine if we are using mocha API's `runner` events. This works really nicely if we are running mocha tests, but my main concern is how to make `run_tests` work among all frameworks with this new logic. I'd like your feedback on these changes and if they are viable.

I've also changed some logic in `TestExecutor`. The code is currently in a working state, so only running one test per node process, but I've moved much of the setup logic to `RunTests`, and eventually launching/disposing the process logic will be moved there, as well. I've added some comments with my thought process-- if you need more info I can elaborate.

My biggest concern would be making `run_tests` work among all testing frameworks, any info on how I can make this work would be greatly appreciated. Using mocha's `runner` events is really great because I would be able to determine pass/fail easily, record start and end times and include much more data in the result than was available before. If `tape` and `ExportRunner` don't have a similar feature (I am unfamiliar with these frameworks) then I'm not sure how viable these changes will be.

Either way, when you have time to review I'd love to hear your opinion.
Thanks!

### DIFF
--- a/Common/Product/SharedProject/ProcessOutput.cs
+++ b/Common/Product/SharedProject/ProcessOutput.cs
@@ -424,7 +424,7 @@ namespace Microsoft.VisualStudioTools.Project {
                 if (_process.StartInfo.RedirectStandardInput) {
                     // Close standard input so that we don't get stuck trying to read input from the user.
                     try {
-                        _process.StandardInput.Close();
+                        //_process.StandardInput.Close();
                     } catch (InvalidOperationException) {
                         // StandardInput not available
                     }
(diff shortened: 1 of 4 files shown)


ozyx:
I understand that having `run_tests` return a `result` would be the best way to streamline all three frameworks, but how would I be able to return a result while at the same time using asynchronous events from mocha to determine test pass/failure? This is why I switched to having `mocha.js` log the result as JSON, but this approach is problematic when it comes to making all three frameworks operate in the same manner.

Not only that but in order to have `mocha.js` run the mocha tests in the correct order (no multiple calls to `before()` and `after()`), I would have to have `mocha.js` run all of the selected tests, instead of calling `mocha.grep` and running each test individually. This is why the runner events are so great, constructing a list of results this way would be rather easy.

Is there a way to block the `return` from happening until I know the tests are done running? I don't think this approach is a good approach for node.js but I'm kind of stumped.

You can merge this into the branch for now, I'll fix the nitpicky (still appreciated!!) stuff in the next PR.

ozyx:
Made the appropriate fixes as per your feedback and this PR should be good to merge now.

PR #1355, opened 2016-10-18
prepare run_tests to run multiple tests

ozyx:
Hello!

In this PR I've made some minor adjustments to allow for run_tests to be able to actually run a list of tests, instead of just one-at-a-time. 

There are some issues that have arisen/will arise moving forward and I'd like to open a dialogue to see your opinion on how I should go about this. 

Firstly, about `mocha.grep`... 
I see now the reason why mocha tests are being run with multiple calls to `before()` and `after()` hooks -- each call to `run_tests` also called `mocha.grep`, turning a singular test suite containing 10 tests into 10 test suites containing 1 test. As my ultimate goal is to make `mocha` tests run properly, I would have to either remove the `grep` call completely (this would mean ALL mocha tests in a file will always run, breaking "Run Selected Tests" functionality) or construct a long `grep` string containing all selected tests and pass that to the `grep` call. 

Second problem -- recording start and end times:
The approach by NTVS currently is to call `RecordStart` on the given test case, send `mocha` the `testName` and `testFile` and have `mocha` grep the one `testName` out of the provided `testFile`. Then once that `process` exits, it's safe to call `RecordEnd` and record the end time for the test, then repeat. 
However, I want to run a list of tests-- in order to do so, I have to provide `mocha` with the list up front so it can adjust its test suite accordingly. If I'm providing `mocha` with a list of tests to run up front, we can no longer record accurate Start and End times for those tests. I could try to record the times in `mocha.js`, but this adds another requirement for the `run_tests` interface, and might make changes more difficult down the road.

Any guidance is welcome and appreciated. Please let me know if you need me to elaborate on anything written here, hopefully I haven't rambled too much...

Thanks again

### DIFF
--- a/Nodejs/Product/Nodejs/TestFrameworks/Tape/tape.js
+++ b/Nodejs/Product/Nodejs/TestFrameworks/Tape/tape.js
@@ -2,6 +2,20 @@
 var EOL = require('os').EOL;
 var fs = require('fs');
 var path = require('path');
+var result = {
+    "title": "",
+    "passed": false,
+    "stdOut": "",
+    "stdErr": ""
+};
+
+process.stdout.write = function (string, encoding, fd) {
+    result.stdOut += string;
+}
+
+process.stderr.write = function (string, encoding, fd) {
+    result.stdErr += string;
+}
 
 function find_tests(testFileList, discoverResultFile, projectFolder) {
     var test = findTape(projectFolder);
@@ -37,8 +51,9 @@ function find_tests(testFileList, discoverResultFile, projectFolder) {
 };
 module.exports.find_tests = find_tests;
 
-function run_tests(testName, testFile, workingFolder, projectFolder) {
+function run_tests(testName, testFile, workingFolder, projectFolder, callback) {
     var testCases = loadTestCases(testFile);
+    result.title = testName;
     if (testCases === null) {
         return;
     }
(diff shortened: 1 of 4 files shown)


mjbvz:
@ozyx Thanks for the work. Good progress overall.

For Mocha, I took and look at their programmatic api and it unfortunately seems poorly designed for this use case. I think you are correct that you'll have to construct a grep string for mocha. Starting from a list of tests cases, the flow would probably look something like this:

* Group tests cases by file.
* For each file:
  * Create a new instance of mocha.
  * Create a grep string by joining together all the test names. This would be something like:
  ```js
  // https://developer.mozilla.org/en/docs/Web/JavaScript/Guide/Regular_Expressions
  function escapeRegExp(string){
     return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
   }
  
  var testGrepString = '^(' + testCases.map(function(testCase){
      return escapeRegExp(testCase.title)
  }).join('|') + ')$'; 
  ```
  * call `Mocha.grep`
  * call `Mocha.addFile`
  * call `Mocha.run`

This adds extra work because you now have to worry about managing a run of Mocha for each file, but  this is the only way I can see to properly support this.

---

For the timing, you'll want to move the execution time calculation from the C# code into test runner JavaScript implementations themselves. This would be another property on the `result` object.

In the case of Mocha, in the `test` event, set a `startTime` property on the result using `Date.now()`. Then in the `test end` event, compute the taken time using `startTime` and `Date.now()`. A similar approach should work for the `tape` and the `export` runners.

Let me know if that makes sense.

PR #1396, opened 2016-10-27
Don't Register as Language(VsTemplate)="JavaScript" in Vs15

mjbvz:
**Bug**
The uwp team contacted us about some of their tests breaking when NTVS is installed in VS15. The root cause seems to be that they attempt to retrieve their "JavaScript" file templates using a VS API, and this only ends up returning the NTVS template path.

**Fix**
Don't register as a "JavaScript" language template provider. Scoped to VS15 only in case this fix causes regressions.

**Testing**
Went through basic project creation, project management, and debugging scenarios in VS15. Everything worked fine.


minestarks:
@mjbvz Thanks!

PR #1413, opened 2016-11-07
Fix vsix vs15 build error

mjbvz:
**Bug**
Vsix signing fails when rezipping the vsixs on new VS2017 releases.

**Fix**
Pass additional required parameters to the `CreateZipPackage` task.


mjbvz:
The one test failure was on `NodejsTests.Debugger.Communication.DebuggerConnectionTests.RaiseConnectionClosedEvent` This seems to be a non deterministic failure unrelated to this change.

PR #1418, opened 2016-11-16
Remove profiler

RyanCavanaugh:
##### Bug

Expected: Profiler is not present
Actual: Is present

##### Fix

Remove profiler

##### Testing

Try to launch profiler. Verify it is not present

cc @billti @paulvanbrenk @mjbvz


Nodejs/Product/AssemblyVersion.cs
```diff
@@ -39,7 +39,7 @@ class AssemblyVersionInfo {
 
     // This version should never change from "4100.00"; BuildRelease.ps1
     // will replace it with a generated value.
-    public const string BuildNumber = "4100.00";
+    public const string BuildNumber = "41115.00";
```
RyanCavanaugh: Should I revert this? It changed on its own

paulvanbrenk:
LGTM

Nodejs/Product/AssemblyVersion.cs
paulvanbrenk: Yep, the comment says it should never change.

paulvanbrenk:
It's gone

PR #1422, opened 2016-11-17
Add localizability to Node tools

paulvanbrenk:
Issue: Need to localize node tools for VS

##### Bug
Node tools are not localized
##### Fix
Initialize localizable files

##### Testing

* Generated xlf files for localization
* Tool for generating xlif files, and localized vsct files


paulvanbrenk:
pinging @billti @RyanCavanaugh

PR #1425, opened 2016-11-22
Resource files for project properties and tools options

paulvanbrenk:
Adding resource files that can be localized for Project properties and Tools - Options


billti:
👍

PR #1431, opened 2016-12-08
Add localization to nodetools

paulvanbrenk:


Build/Common.Build.targets
```diff
@@ -145,6 +145,10 @@
     <Copy SourceFiles="@(DebugSymbolsProjectOutputGroupOutput->'%(FinalOutputPath)')"
           DestinationFolder="$(CopyOutputsToPath)Symbols" />
 
+    <Copy SourceFiles="@(SatelliteDllsProjectOutputGroupOutput)"
+          DestinationFiles="@(SatelliteDllsProjectOutputGroupOutput->'$(CopyOutputsToPath)UnsignedBinaries\%(TargetPath)')"
```
billti: Is TargetPath the right var here? From the docs: "The absolute path name of the primary output file for the build".

Nodejs/Product/InteractiveWindow/InteractiveWindow.csproj
```diff
@@ -164,6 +141,151 @@
...
+    <EmbeddedResource Include="Resources.it.resx" />
+    <EmbeddedResource Include="Resources.ja.resx" />
+    <EmbeddedResource Include="Resources.ko.resx" />
+    <EmbeddedResource Include="Resources.pl.resx" />
+    <EmbeddedResource Include="Resources.pt-BR.resx" />
+    <EmbeddedResource Include="Resources.ru.resx" />
+    <EmbeddedResource Include="Resources.tr.resx" />
+    <EmbeddedResource Include="Resources.zh-Hans.resx" />
+    <EmbeddedResource Include="Resources.zh-Hant.resx" />
+    <ConditionalEmbeddedResource Include="VSPackage.de.resx">
```
billti: Very minor nit... but any reason these are in alphabetical (or at least consistent) order? Makes checking they are all present and correct much easier.

Nodejs/Product/InteractiveWindow/InteractiveWindow.csproj
```diff
@@ -164,6 +141,151 @@
...
+    </EmbeddedResource>
+    <ConditionalEmbeddedResource Include="VSPackage.resx">
+      <MergeWithCTO>true</MergeWithCTO>
+      <ManifestResourceName>Microsoft.VSPackage</ManifestResourceName>
+      <SubType>Designer</SubType>
+    </ConditionalEmbeddedResource>
+  </ItemGroup>
+  <ItemGroup Condition="$(VSTarget) == '15.0'">
+    <VSCTCompile Include="InteractiveWindow.cs.vsct">
+      <ResourceName>Menus.ctmenu</ResourceName>
```
billti: Not super familiar with this stuff... but it's OK to add all language versions of the .vsct with the same "ResourceName"? Does the locale from the included item persist by some other means so the right one is selected at runtime?

Nodejs/Product/Npm/Npm.csproj
```diff
@@ -163,6 +163,26 @@
     </EmbeddedResource>
   </ItemGroup>
   <ItemGroup>
+    <EmbeddedResource Include="Resources.cs.resx" />
```
billti: Does this not need to be dependent on Dev15 like the other projects localized items?

Nodejs/Setup/BuildRelease.ps1
```diff
@@ -553,18 +565,35 @@ try {
...
                     Write-Output "Skipping symbol submission for $($i.VSName) because the build failed"
                     continue
                 }
                 submit_symbols "$project_name$spacename" "$buildnumber $($i.VSName) $config" "binaries" $i.signed_bindir $symbol_contacts
                 submit_symbols "$project_name$spacename" "$buildnumber $($i.VSName) $config" "symbols" $i.symboldir $symbol_contacts
 
+                Write-Output "End symbol submission for $($i.VSName)"                
+            }
+
+            foreach ($i in $target_info) {
```
billti: Why did you separate this out into another iteration over $target_info? The extra loop doesn't check failed logs and skip the setup build if the build failed. Is that correct?

Nodejs/Product/InteractiveWindow/InteractiveWindow.csproj
paulvanbrenk: In most cases when I hand authored them, I tried to keep 'en' as the first if it needed some specific logic, and the remaining locales in alphabetical order. When they were added by project system I didn't change the order.

Nodejs/Product/Npm/Npm.csproj
paulvanbrenk: Yes

PR #1447, opened 2016-12-19
Simplify VS2017 build and signing

paulvanbrenk:
Use VSSDK to build VSIX and JSON manifests
Ensure the signed sattelite assemblies are copied to VSIX
Remove no longer needed files


yuit:
Look good from what I understand

PR #1450, opened 2016-12-22
Include localized resources in the shared project.

paulvanbrenk:
Include missing resources, which caused an issue while debugging


Common/Product/SharedProject/SharedProject.proj
```diff
@@ -583,16 +583,6 @@
       <Link>SharedProject\WaitDialog.cs</Link>
       <Visible>true</Visible>
     </Compile>
-    <Compile Include="$(_ProjectSystemFilesLocation)\WebPiComponentPickerControl.cs">
```
billti: Why isn't this (and the file below) also in Dev15? Is it not needed?

paulvanbrenk: Yep, put that back

Issue #1452, opened 2017-01-03
General Properties page is hard to read in high dpi


anubmat:
![image](https://cloud.githubusercontent.com/assets/17298501/21620568/d626f82e-d1a9-11e6-8dd1-b3ff3a34802b.png)

PR #1453, opened 2017-01-03
Change dimensions and font size to improve appearance of general properties page in high dpi

anubmat:
Bug #1452 : General Properties page is hard to read in high dpi


paulvanbrenk:
Can you add a screenshot on how this looks after your changes?

anubmat:
This is how node project property page will look. In both cases text is big enough to read (looks small here). It will begin to look a little worse in Normal resolution, but will become readable in high-dpi. I adjusted locations and sizes a/c to high dpi. I think an improvement would be to have a table layout like TS, and make autoscalable columns.

Normal (1920X1080):
image

High dpi:
image

paulvanbrenk:
Thanks, that looks pretty good.

anubmat:
@RyanCavanaugh , can you help me check this in?

RyanCavanaugh:
Thanks!

PR #1455, opened 2017-01-06
Add UI culture so resources localize appropriately

RyanCavanaugh:
This causes the satellite assemblies to get built with the correct XAML resource manifest names so things load as expected. Verified on EN (which was also broken) and hand-loc'd Chinese since the Install NPM Packages dialog isn't loc'd yet

image

/cc @billti  @paulvanbrenk


Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -44,6 +44,7 @@
     <StartProgram>$(DevEnvDir)devenv.exe</StartProgram>
     <StartArguments>/rootSuffix Exp /Log</StartArguments>
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\..\</SolutionDir>
+    <UICulture>en</UICulture>
```
billti: So this fixes all cultures... for reals? 😱

Issue #403, opened 2015-08-31
Allow Mocha to be installed Globally


SchmuckerN:
Hello, 

I have a Visual Studio solution which contains multiple Node.js projects. I would like to run all tests of all projects using mocha. For that I have to install mocha for every project locally, so that the tests can be found by NTVS. So I'll have mocha installed multiple times on my harddrive. Isn't it possible to have it installed once only? 

 Thanks, 
 Natascha

andreesteve:
I needed this for one of my setups. I sent a pull request to improve the module look up: #1456.

@SchmuckerN do you want to give it a try and see if that works for you? You can just replace your existing mocha.js file under the path `"C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\Extensions\Microsoft\Node.js Tools for Visual Studio\1.1\TestFrameworks\Mocha\mocha.js"` with the file in the pull request (please keep a backup before you replace it).

PR #1456, opened 2017-01-14
Allow mocha to be installed globally or on any parent folder node_mod…

andreesteve:
#### Issue
#403  (Allow Mocha to be installed Globally) 

##### Fix
This PR changes how mocha module is found. Instead of only looking on the project folder node_modules, I transverse the file system tree looking at node_modules folder on parent directories. If the module is not found in any parent node_modules folder, then a global require is issued.

##### Testing
1. No globally installed mocha package, no mode_modules on project folder or parent folder -> Expects error message (module not found).
2. Mocha locally installed (previous behavior) -> Expects to loads local module as before.
3. Mocha installed on some parent node_modules folder -> Expects to load module from parent node_module folder.
4. Mocha not installed locally or on parent folder, but installed globally -> Expects to load global module. (**NOTE:**: on Windows one might need to set the NODE_PATH environment variable to point to NPM global install folder).

### DIFF
--- a/Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
+++ b/Nodejs/Product/Nodejs/TestFrameworks/mocha/mocha.js
@@ -4,7 +4,7 @@ var fs = require('fs');
 var path = require('path');
 
 // Choose 'tap' rather than 'min' or 'xunit'. The reason is that
-// 'min' produces undisplayable text to stdout and stderr under piped/redirect, 
+// 'min' produces undisplayable text to stdout and stderr under piped/redirect,
 // and 'xunit' does not print the stack trace from the test.
 var defaultMochaOptions = { ui: 'tdd', reporter: 'tap', timeout: 2000 };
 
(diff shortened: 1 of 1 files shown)


msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

paulvanbrenk:
@andreesteve  looks like this fell through the cracks. Do you mind singing the CLA so I can merge this in. 

One change I would like to see is to add the error message back in case the script can't find the mocha install.

andreesteve:
@paulvanbrenk , I've rebased the change on top of current master (the file was moved).

The error message, in case mocha is not found, is still there, on line 273. I think I removed the EOL not sure why - this was a while ago - so the formatting looks a bit different, but the error is still reported for such case.

It shows the CLA signed here, not sure how I sign it again. Do you have the link? I can try it again.

paulvanbrenk:
Thanks, I see it now.. looks good to me. May take me another week to actually get this in, but it will go in.

Thanks for the contribution!

marklam:
I've just hit this limitation, it would me nice if this PR was merged.

joj:
Overridden by more recent changes. Closing.

PR #1469, opened 2017-02-01
Fix build deadlock

aozgaa:
`SetVerbosity` needs the UI thread to observe the verbosity setting in the registry.

We move calls to `SetVerbosity` (renamed to `RefreshVerbosity`) to occur on the UI thread. We also simplify cache invalidation of the verbosity setting stored in the logger.

### DIFF
--- a/Common/Product/SharedProject/IDEBuildLogger.cs
+++ b/Common/Product/SharedProject/IDEBuildLogger.cs
@@ -457,8 +457,6 @@ namespace Microsoft.VisualStudioTools.Project {
             // If importance is too low for current settings, ignore the event
             bool logIt = false;
 
-            this.SetVerbosity();
-
             switch (this.Verbosity) {
                 case LoggerVerbosity.Quiet:
                     logIt = false;
@@ -507,11 +505,13 @@ namespace Microsoft.VisualStudioTools.Project {
         /// <summary>
         /// Sets the verbosity level.
         /// </summary>
-        private void SetVerbosity() {
+        public void SetVerbosity() {
             if (!this.haveCachedVerbosity) {
                 this.Verbosity = LoggerVerbosity.Normal;
 
                 try {
+                    serviceProvider.GetUIThread().MustBeCalledFromUIThread();
+
                     var settings = new ShellSettingsManager(serviceProvider);
                     var store = settings.GetReadOnlySettingsStore(SettingsScope.UserSettings);
                     if (store.CollectionExists(GeneralCollection) && store.PropertyExists(GeneralCollection, BuildVerbosityProperty)) {
--- a/Common/Product/SharedProject/IDEBuildLogger.cs
+++ b/Common/Product/SharedProject/IDEBuildLogger.cs
@@ -49,7 +49,6 @@ namespace Microsoft.VisualStudioTools.Project {
         private IVsHierarchy hierarchy;
         private IServiceProvider serviceProvider;
         private Dispatcher dispatcher;
-        private bool haveCachedVerbosity = false;
 
         // Queues to manage Tasks and Error output plus message logging
         private ConcurrentQueue<Func<ErrorTask>> taskQueue;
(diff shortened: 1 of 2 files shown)


Common/Product/SharedProject/ProjectNode.cs
```diff
@@ -2458,6 +2458,8 @@ internal enum EventTriggering {
             } else {
                 ((IDEBuildLogger)this.BuildLogger).OutputWindowPane = output;
```
vladima: 1. cast should be redundant since you've change the type of property to `IDEBuildLogger`
2. (nit) I doubt that we need this protected property since it seems that the only place where `BuildLogger` is access is this class - it can be done as an improvement in a separate PR.

vladima: the cast in line 2459 is still there

Common/Product/SharedProject/IDEBuildLogger.cs
```diff
@@ -507,34 +503,30 @@ class NavigableErrorTask : ErrorTask {
...
+        public void RefreshVerbosity()
+        {
+            this.Verbosity = LoggerVerbosity.Normal;
+
+            try
+            {
+                serviceProvider.GetUIThread().MustBeCalledFromUIThread();
+
+                var settings = new ShellSettingsManager(serviceProvider);
+                var store = settings.GetReadOnlySettingsStore(SettingsScope.UserSettings);
```
aozgaa: Must access to the SettingsStore occur on the UI thread?

AArnott: tip: The default answer to that must always be "yes" unless documentation proves otherwise.

Common/Product/SharedProject/IDEBuildLogger.cs
```diff
@@ -507,34 +503,30 @@ class NavigableErrorTask : ErrorTask {
...
-                        ex.ToString()
-                    );
-                    this.QueueOutputText(MessageImportance.High, message);
+        public void RefreshVerbosity()
+        {
+            this.Verbosity = LoggerVerbosity.Normal;
+
+            try
+            {
+                serviceProvider.GetUIThread().MustBeCalledFromUIThread();
```
AArnott: GetUIThread() makes a GetService call on the IServiceProvider. If this is a VS service provider then this will require an RPC call to complete, which can cause deadlocks. But the intent of this line appears to be to throw if not on the UI thread to avoid deadlocks. So I don't *think* you're reliably accomplishing what you mean to.
Can you simply call `ThreadHelper.ThrowIfNotOnUIThread()`?

Common/Product/SharedProject/ProjectNode.cs
```diff
@@ -2456,8 +2456,10 @@ internal enum EventTriggering {
                     oldLogger.Dispose();
                 }
             } else {
-                ((IDEBuildLogger)this.BuildLogger).OutputWindowPane = output;
+                this.BuildLogger.OutputWindowPane = output;
             }
+
+            this.buildLogger.RefreshVerbosity();
```
AArnott: This new line is about the only one that looks significant here. If we're making escrow changes, why make all the other changes to this file at this point?

Common/Product/SharedProject/IDEBuildLogger.cs
aozgaa: This method is only compiled in debug builds. We should broadly replace calls to this to `ThreadHelper.ThrowIfNotOnUIThread()` but I hoped to limit that in the scope of this PR.

AArnott: OK, sounds good. Including the part about replacing with ThreadHelper later so it's a Release build runtime check too later on.

AArnott: Although `MustBeCalledFromUIThread()` compiles out of Release builds, `GetUIThread()` *does not*, and that's the call I'm concerned about. So unless the IServiceProvider you're querying is free-threaded for this call, yous till have a potential RPC deadlock.

PR #1471, opened 2017-02-01
Hack vsman file

RyanCavanaugh:
Adds code to BuildRelease.ps1 to write correct size and sha256 info to the .vsman file

### DIFF
--- a/Nodejs/Setup/BuildRelease.ps1
+++ b/Nodejs/Setup/BuildRelease.ps1
@@ -243,6 +243,39 @@ function after-build($buildroot, $target) {
     }
 }
 
+# Fixes hashing information in the .vsman file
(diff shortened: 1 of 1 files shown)


Nodejs/Setup/BuildRelease.ps1
```diff
@@ -673,9 +706,12 @@ try {
                 } } | `
                 %{ Copy-Item $_.src $_.dest -Force -EA 0; $_ } | `
                 %{ "Copied $($_.src) -> $($_.dest)" }
+
+            $vsmanFile = "$($outdir)\NodejsTools.vsman"
+            fix-vs-manifest $vsmanFile $i.final_msidir
```
billti: Per the build errors, this should be conditional on VsTarget == "15.0", as the files won't exist for Dev14 builds.

PR #1474, opened 2017-02-03
Make sure whe have the right name for the Express 4 JS template.

paulvanbrenk:
The Id is pulled from the vspackage.resx file (and the matching localized files)


paulvanbrenk:
Replaced with #1519

PR #1483, opened 2017-02-24
Webkitdebugger for new node debugger protocol

paulvanbrenk:
Node will stop supporting the old debugger protocol.

This PR adds uses the webkit debugger, which already exists in VS, which supports the new V8 debugger protocol, which is expected to be the only supported in Node V8.

Note: this needs localization so the new option work in other languages than 'en', and needs some bugfixes in the WebKit debugger.


Nodejs/Product/Nodejs/Debugger/NodeDebugger.cs
```diff
@@ -76,82 +76,88 @@ sealed class NodeDebugger : IDisposable {
...
-            ushort? debuggerPort = null,
-            bool createNodeWindow = true)
+        public NodeDebugger(Uri debuggerEndpointUri, int id)
+            : this() {
+            _debuggerEndpointUri = debuggerEndpointUri;
+            _id = id;
+            _attached = true;
+        }
+
+        public NodeDebugger(string exe, string script, string dir, string env, string interpreterOptions, NodeDebugOptions debugOptions, ushort? debuggerPort = null, bool createNodeWindow = true)
```
billti: This line is nearly 200 chars wide now. Please leave it wrapped as it originally was so it is readable.

Nodejs/Product/Nodejs/Options/NodejsGeneralOptionsControl.cs
```diff
@@ -66,6 +66,11 @@ public partial class NodejsGeneralOptionsControl : UserControl {
             _waitOnNormalExit.Checked = page.WaitOnNormalExit;
             _editAndContinue.Checked = page.EditAndContinue;
             _checkForLongPaths.Checked = page.CheckForLongPaths;
+            _webkitDebugger.Checked = page.UseWebKitDebugger;
+
+#if !DEV15
+            _webkitDebugger.Enabled = false;
```
billti: So does this option appear at all on < Dev15? Or does it appear and is disabled?

Nodejs/Product/Nodejs/Options/NodejsGeneralOptionsControl.en.resx
```diff
@@ -118,125 +118,38 @@
...
-  <data name="&gt;&gt;_surveyNewsCheckLabel.ZOrder" xml:space="preserve">
-    <value>0</value>
-  </data>
-  <data name="_surveyNewsCheckCombo.Anchor" type="System.Windows.Forms.AnchorStyles, System.Windows.Forms">
-    <value>Left, Right</value>
-  </data>
-  <data name="_surveyNewsCheckCombo.Items" xml:space="preserve">
-    <value>Never</value>
+  <data name="_webkitDebugger.Text" xml:space="preserve">
+    <value>Use the new NodeJs debugger protocol (for NodeJs v7.5+)</value>
```
billti: This actually works from Node.js v6.8.0 onwards on Windows.

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -66,53 +67,147 @@ class NodejsProjectLauncher : IProjectLauncher {
...
+            var launchResults = new VsDebugTargetProcessInfo[1];
+            debugger.LaunchDebugTargets4(1, new[] { dbgInfo }, launchResults);
+        }
+
+        private static IntPtr MarshalDebugEngines(Guid[] debugEngines) {
+            if (debugEngines.Length == 0) {
+                return IntPtr.Zero;
+            }
+
+            var guiSize = Marshal.SizeOf(typeof(Guid));
```
billti: "guidSize"?

Nodejs/Product/Nodejs/Options/NodejsGeneralOptionsControl.cs
paulvanbrenk: it appears but is disabled. I can hide it, but then there is a weird gap between the normal options and the survey options.

Nodejs/Product/Nodejs/Debugger/NodeDebugger.cs
paulvanbrenk: done

Common/Product/SharedProject/ProjectNode.cs
```diff
@@ -5597,6 +5596,10 @@ internal enum EventTriggering {
         internal HierarchyNode FindNodeByFullPath(string name) {
             Site.GetUIThread().MustBeCalledFromUIThread();
 
+            if (name.StartsWith("mdha:", StringComparison.OrdinalIgnoreCase)) {
```
paulvanbrenk: found a better solution.

PR #1499, opened 2017-03-14
Fix node version in option string.

paulvanbrenk:
Testing shows that this only works reliably for Node v7.0 and up.


billti:
Why does this need to change? It's actually present and working on Windows since v6.8.

paulvanbrenk:
It doesn't reliably connect from Visual Studio, where 7.0+ does. Since the new protocol doesn't add any functionality, changing the string is cheaper than making it work with an old version of Node.

kant2002:
This change will require changes in the localization as well.

RyanCavanaugh:
Let's search/replace this in the localized copies rather than run a cycle with the loc team

paulvanbrenk:
Thanks, I'll update the localized files

PR #1512, opened 2017-03-27
Replace the dynamic TypeScript version check with a package prerequisite

minestarks:
##### Bug
NTVS uses typescriptServices.js to get the current TypeScript version, complains if the file is missing

##### Fix
Remove the dynamic check, properly declare a dependency on TypeScript using the vsixmanifest (Dev15 only)


billti:
Much of the code being removed is for Dev14 (Dev15 already includes at least TS 2.1 if you install the Node.js workload). As we still need to ship a final NTVS 1.3 release for Dev14 from this branch, this would break that functionality.

We do use `#ifdef`s to target the different versions, so maybe just exclude the removed code from Dev15 builds.

minestarks:
Ah I was wondering about that. I like having the package prerequisite there for Dev15, because this VSIX lets you install it standalone (not from setup). I'll add ```#ifdef```s.

PR #1517, opened 2017-04-04
Add null check

paulvanbrenk:
Issue # Null reference on response from debugger

### DIFF
--- a/Nodejs/Product/Nodejs/Debugger/Communication/DebuggerClient.cs
+++ b/Nodejs/Product/Nodejs/Debugger/Communication/DebuggerClient.cs
@@ -189,10 +189,10 @@ namespace Microsoft.NodejsTools.Debugger.Communication {
         /// </summary>
         /// <param name="message">Message.</param>
         private void HandleResponseMessage(JObject message) {
-            TaskCompletionSource<JObject> promise;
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/Debugger/Communication/DebuggerClient.cs
```diff
@@ -189,10 +189,10 @@ sealed class DebuggerClient : IDebuggerClient {
...
-            TaskCompletionSource<JObject> promise;
-            var messageId = (int)message["request_seq"];
+            var messageId = message["request_seq"];
 
-            if (_messages.TryGetValue(messageId, out promise)) {
+            TaskCompletionSource<JObject> promise;
+            if (messageId != null && _messages.TryGetValue((int)messageId, out promise)) {
                 promise.SetResult(message);
             } else {
                 Debug.Fail(string.Format(CultureInfo.CurrentCulture, "Invalid response identifier '{0}'", messageId));
```
minestarks: if ```messageId == null```, will this ```string.Format``` throw?

paulvanbrenk: thanks good catch

PR #1524, opened 2017-04-12
Test adapter improvements

billti:
The test-adapter-improvements branch was fairly out of date, as master has had a lot of work recently for VS 2017 build, string localization, Node.js debugger updates, etc... This is effectively that branch with master merged in, then the merge conflicts fixed, then I fetch the work in pull request https://github.com/Microsoft/nodejstools/pull/1430 from @ozyx, merged that and fixed the merge conflicts also.

This builds clean, but I haven't done much testing on it yet (or even code reviewed the existing changes in depth). Consider this the new baseline branch to get that work reviewed, tested, completed, and checked in.

PR #1527, opened 2017-04-13
Explicitly load the VS assemblies on Dev15 to fix unit test debugging

billti:
See issue description in the comment at https://github.com/Microsoft/nodejstools/pull/1524#issuecomment-293761532 .

This may not be the most elegant solution, or the right one long term, but it does work and is simple enough. If in the test executor process, then by loading the assemblies explicitly when it attempts to run tests, but before hitting any code that would cause .NET to try and locate the assemblies, they are already loaded when required.


paulvanbrenk:
We should figure out who the owner is for the exe and ask them to update the `app.config`.

billti:
Addressed feedback.

PR #1536, opened 2017-04-21
Scroll focused control into visible area

paulvanbrenk:
This fixes an issue where when tabbing through the controls in the propertypage the focused control is not automatically scrolled into focus.


RyanCavanaugh:
Why do we need all this Win32 garbage?

paulvanbrenk:
Great question, in VS this is handled by the property page baseclass... but we can't use that since `internal` only.. so copied from there.

PR #1538, opened 2017-04-24
Replace packages.config with project.json

paulvanbrenk:
This way we can use a wildcard for the version, which helps
prevent issues when referencing MicroBuild packages.

The latest version of the package requires the Framework version to be set to 4.6.1 so, doing that for all the SWIX/VSMAN projects. It doesn't have any affect on the build though.

### DIFF
--- a/EnvironmentSetup.ps1
+++ b/EnvironmentSetup.ps1
@@ -68,7 +68,7 @@ if ($microbuild -or ($vstarget -eq "15.0" -and -not $skipRestore)) {
     Write-Output ""
     Write-Output "Installing Nuget MicroBuild packages"
 
-    & "$rootdir\Nodejs\.nuget\nuget.exe" restore "$rootdir\Nodejs\Setup\swix\packages.config" -PackagesDirectory "$packagedir"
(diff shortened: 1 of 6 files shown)


EnvironmentSetup.ps1
```diff
@@ -68,7 +68,7 @@ if ($microbuild -or ($vstarget -eq "15.0" -and -not $skipRestore)) {
     Write-Output ""
     Write-Output "Installing Nuget MicroBuild packages"
 
-    & "$rootdir\Nodejs\.nuget\nuget.exe" restore "$rootdir\Nodejs\Setup\swix\packages.config" -PackagesDirectory "$packagedir"
+    & "$rootdir\Nodejs\.nuget\nuget.exe" restore "$rootdir\Nodejs\Setup\swix\project.json" -PackagesDirectory "$packagedir" -ConfigFile "$rootdir\Nodejs\.nuget\nuget.config"
```
billti: I recall this config file causing some issue before. Make sure this doesn't break the local or nightly build.

paulvanbrenk: This works with the nightly build, however for a local build you should set `-skipRestore` to `true`. This won't prevent you from building the product, but you won't be able to build the setup packages we need for the VS insertion.

PR #1539, opened 2017-04-25
Include an option to force the use of the chrome debugging protocol.

paulvanbrenk:
Checked with node 7.9, and no additional changes are needed.

We'll remove this when node 8 is final

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
@@ -16,6 +16,7 @@ using System.Web;
 using System.Windows.Forms;
 using Microsoft.NodejsTools.Debugger;
 using Microsoft.NodejsTools.Debugger.DebugEngine;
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -68,7 +69,7 @@ private int Start(string file, bool debug)
                 return VSConstants.S_OK;
             }
 
-            var chromeProtocolRequired = Nodejs.GetNodeVersion(nodePath) >= new Version(8, 0);
+            var chromeProtocolRequired = Nodejs.GetNodeVersion(nodePath) >= new Version(8, 0) || CheckDebugProtocolOption();
```
changsi-an: Do you need to fully qualify the name for a static method?

paulvanbrenk: Not if it's in the same class....

PR #1551, opened 2017-05-03
Changes to the parameters we pass to PineZorro, including the new ins…

msft-cwells:
…tall path for the Willow-installed VS instance of node.exe

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
@@ -359,8 +359,6 @@ namespace Microsoft.NodejsTools.Project
             var pathToNode2DebugAdapterRuntime = Environment.ExpandEnvironmentVariables(@"""%ALLUSERSPROFILE%\" +
                     $@"Microsoft\VisualStudio\NodeAdapter\{visualStudioInstallationInstanceID}\extension\out\src\nodeDebug.js""");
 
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -350,7 +350,11 @@ private void StartWithChromeV2Debugger(string program, string nodeRuntimeExecuta
 
             var setupConfiguration = new SetupConfiguration();
 
-            var visualStudioInstallationInstanceID = setupConfiguration.GetInstanceForCurrentProcess().GetInstanceId();
+            var setupInstance = setupConfiguration.GetInstanceForCurrentProcess();
+
+            var visualStudioInstallationInstanceID = setupInstance.GetInstanceId();
+
+            var pathToNodeExe = setupInstance.GetInstallationPath() + "\\JavaScript\\Node.JS\\v6.4.0_x86\\Node.exe";
```
paulvanbrenk: Can you switch this to `Path.Combine`, and add a comment why we need this specific version of Node? 

Is the following worth investigating: have the node installation package write a key to the registry, and have this component read that key... that way when the version of node is updated (or the path is changed), we don't have to remember to make changes here too.

msft-cwells: Changed to use Path.Combine, and added an explanatory comment.

In the future we can look into whether we can make the node versioning dynamic, but for now we are tied to this specific version, which is (indirectly) a willow dependency of NTVS workload

PR #1554, opened 2017-05-04
Dynamically generate the VSIX version

minestarks:
Obviously the 42.42.42.42 will need to be replaced with something dynamic.


paulvanbrenk:
Included in a PR which dynamically calculates the version number.

PR #1556, opened 2017-05-04
Start webbrowser when using the webkit debugger.

paulvanbrenk:
Ensure the debugger is launched when that option is selected.

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
@@ -86,7 +86,7 @@ namespace Microsoft.NodejsTools.Project
                 }
                 else
                 {
-                    StartAndAttachDebugger(file, nodePath);
+                    StartAndAttachDebugger(file, nodePath, startBrowser);
                 }
             }
             else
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -152,6 +158,18 @@ private void StartAndAttachDebugger(string file, string nodePath)
...
             AttachDebugger(dbgInfo);
+
+            if (startBrowser && uri != null)
+            {
+                OnPortOpenedHandler.CreateHandler(
+                    uri.Port,
+                    shortCircuitPredicate: () => process.HasExited,
+                    action: () =>
+                    {
+                        VsShellUtilities.OpenBrowser(url, (uint)__VSOSPFLAGS.OSP_LaunchNewBrowser);
```
billti: Does this take any notice of the browser selected in the "Start debugging" drop-down? I noticed it was always launching the default browser, no matter what was selected here (though that may be a different bug to address).

paulvanbrenk: Yeah, I think this is the VS API to launch the OS default browser.... but I think that's the same behavior as previous versions of Node. I'll investigate how to grab the selected browser, though that may require interaction with some ASP.NET code.

PR #1558, opened 2017-05-05
Minimal work to support the new NPM protocol for package search.

paulvanbrenk:
This is a minimal change to the interaction with the NPM registry, instead of using a local cache this queries the NPM registry directly.

Only removed the minimum amount of code to clean the UI up.

If the network is unavailable the UI will keep spinning on the search screen, but not crash VS. The user is still able to close the UI manually.

Note: the refresh button is removed.

image

### DIFF
--- a/Nodejs/Product/Npm/IPackageCatalog.cs
+++ b/Nodejs/Product/Npm/IPackageCatalog.cs
@@ -47,6 +47,10 @@ namespace Microsoft.NodejsTools.Npm {
         public long? ResultsCount => 0;
 
         public async Task<IEnumerable<IPackage>> GetCatalogPackagesAsync(string filterText) {
+
+            // All exceptions thrown here and in the called methods are handled by the
+            // NPM search dialog, so we don't have to do any exception handling here.
+
             var relativeUri = string.Format("/-/v1/search?text={0}", filterText);
             var searchUri = new Uri(defaultRegistryUri, relativeUri);
 
--- a/Nodejs/Product/Npm/IPackageCatalog.cs
+++ b/Nodejs/Product/Npm/IPackageCatalog.cs
@@ -16,16 +16,146 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net;
 using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.NodejsTools.Npm {
     public interface IPackageCatalog {
         DateTime LastRefreshed { get; }
 
-        Task<IEnumerable<IPackage>> GetCatalogPackagesAsync(string filterText, Uri registryUrl = null);
-
-        IPackage this[string name] { get; }
+        Task<IEnumerable<IPackage>> GetCatalogPackagesAsync(string filterText);
 
         long? ResultsCount { get; }
+    }
+
+    // This class is a wrapper catalog to directly query the NPM repo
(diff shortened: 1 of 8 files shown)


billti:
Looks like tests are failing the build. I know the tests aren't much use, but hopefully is a simple fix.

Nodejs/Product/Npm/IPackageCatalog.cs
```diff
@@ -16,16 +16,146 @@
...
+
+        public static readonly IPackageCatalog Instance = new EmptyPackageCatalog();
+
+        private EmptyPackageCatalog() { }
+
+        public DateTime LastRefreshed => DateTime.Now;
+
+        public long? ResultsCount => 0;
+
+        public async Task<IEnumerable<IPackage>> GetCatalogPackagesAsync(string filterText) {
```
billti: So this gets called repeatedly with new filterText as the user types?

Nodejs/Product/Npm/IPackageCatalog.cs
```diff
@@ -16,16 +16,146 @@
...
+                                return ReadPackagesFromArray(jsonReader);
+                            default:
+                                throw new InvalidOperationException("Unexpected json token.");
+                        }
+                    }
+                }
+            }
+
+            // should never get here
+            throw new InvalidOperationException("Unexpected json token.");
```
billti: What do these throws cause to happen (not just this, but in a number of places in this code if the data isn't as expected)? We shouldn't bring down VS if some remote API returns incorrect data. Hopefully they just get swallowed further up the chain. Can you verify and add a comment?

Nodejs/Product/Npm/IPackageCatalog.cs
paulvanbrenk: Yes, but there is a 3 second delay... so it shouldn't be too bad.

Nodejs/Product/Npm/IPackageCatalog.cs
paulvanbrenk: They do get swallowed I'll add a comment

josephdubeau:
Hello, 
I'm experiencing the same problem.  Is the a fix or work around?

paulvanbrenk:
@josephdubeau This is fixed in 1.3 for Visual Studio 2015, and Update 3 for Visual Studio 2017. Otherwise you can use npm from the command line to add packages to the `package.config` and/or install packages.

josephdubeau:
Thank you so much, I will try the update first.


________________________________

josephdubeau:
I able to do a command line update but not in VS 2017.  I have version 15.2 (26430.16)

Where can I download updates?

________________________________

paulvanbrenk:
Previews for Visual Studio 2017 are available here. [0]

[0] https://www.visualstudio.com/vs/preview/

Dredok:
Hi, I have visual studio 2017 enterprise and, as said by @josephdubeau it still happens in latest update (26430.16) (or its a regression IDK)

Is there something I can do to fix this? or should I rely to commandline npm?

paulvanbrenk:
@Dredok that version number points indicates you have update **2**, this is fixed in Update **3** which is in preview 8 currently.

image

Dredok:
@paulvanbrenk  oh I see! I thought it was a public update ... so I guess I will have to wait. Thanks

paulvanbrenk:
@Dredok to be clear it is public, just as a _preview_.

PR #1571, opened 2017-05-10
Build number generation v2

paulvanbrenk:
Autogenerate file, assembly, and vsix versions from the build number.


paulvanbrenk:
The CI build is failing because it doesn't support VS 2017 yet.

Build/Common.Build.settings
```diff
@@ -144,4 +144,38 @@
   <PropertyGroup>
     <CodeAnalysisRuleSet>$(BuildRoot)\Build\xTVS.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
+
+  <PropertyGroup>
+    <MajorVersion>1</MajorVersion>
+    <MinorVersion>4</MinorVersion>
+
+    <!-- This was reset for Dev15. It needs to eventually be reset to the current year for Dev16 -->
```
minestarks: This comment doesn't make as much sense here as it did in the original place you copied from. Here you can just set the year to 2017

paulvanbrenk: I still like to reset the year every so often to make sure the build numbers < int16.max

PR #1576, opened 2017-05-12
Make sure we send options for the install_typings in both cases.

paulvanbrenk:
This fixes a null reference on options in the case of `installTypingsForProject`

### DIFF
--- a/Nodejs/Product/Nodejs/TypingsAcquisitionTool/bin/install_typings
+++ b/Nodejs/Product/Nodejs/TypingsAcquisitionTool/bin/install_typings
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+,#!/usr/bin/env node
 /**
     Tool to aquire typings used for NTVS IntelliSense.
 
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/TypingsAcquisitionTool/bin/install_typings
```diff
@@ -18,19 +18,21 @@ var argv = minimist(process.argv.slice(2), {
...
+};
+
 var packagesToInstall = argv._;
 
 if (!packagesToInstall.length) {
     // top level package install
     typingsTool.installTypingsForProject(options)
 } else {
     typingsTool.runAll(packagesToInstall.map(function (name) {
-        var options = {
```
billti: This won't work as intended. By having options outside of this function body, you are passing the same options object in with every call to installTypingsForPackage, so unless it executes synchronously and doesn't refer to the object again - all later uses refer to the same object - so only the last settings take effect. Safer to just create a different 'options' object (maybe inline) if the other branch of the 'if' statement.

Issue #1545, opened 2017-04-26
Strange Icons in `Add Item` Menu


aozgaa:
TypeScript File, Json File, and LESS File are strange.

##### Expected Behavior 
Get Icons that say "JS" or "TS" on them.
##### Actual Behavior 
![image](https://cloud.githubusercontent.com/assets/3226984/25454979/12eb1998-2a83-11e7-9644-e7588750578f.png)

Microsoft Visual Studio Enterprise 2017 Int Preview
Version 15.2 (26424.2-Preview) d15rel Internal Preview - Dogfood
VisualStudio.15.IntPreview/15.2.0-Preview+26424.2.d15rel
Microsoft .NET Framework
Version 4.6.01586

JavaScript Language Service   2.0
JavaScript Language Service

JavaScript Project System   2.0
JavaScript Project System

Node.js Tools   1.3.50417.1
Adds support for developing and debugging Node.js apps in Visual Studio

TypeScript   2.2.0.0
TypeScript tools for Visual Studio

##### Steps to Reproduce
1. 
```
cd c:\
mkdir temp
cd temp
npm install express-generator -g
express --view=pug myapp
cd myapp
npm install
```
2. Start VS
3. `Ctrl+Shift+N` (Create new project)
4. Templates -> JavaScript -> From Existing Node.js code
5. Click `OK`
6. Pick folder containing Node.js code: `c:\temp\myapp`
7. `Alt-F` (Finish)
8. In solution explorer, Right Click On Project -> Add... -> New Item (`Ctrl+Shift+A`)
9. Observe icons per above.

paulvanbrenk:
Does the same happen in a console app?

PR #1577, opened 2017-05-13
Bunch of smaller fixes

paulvanbrenk:
* Remove the logconverter since it's no longer used
* Fix a null ref exception
* Fix the icons in add new item dialog (#1545)


aozgaa:
LGTM

PR #1580, opened 2017-05-15
Test adapter reg key

paulvanbrenk:
If the regkey is set to something, disable the NTVS test adapter.

### DIFF
--- a/Common/Product/TestAdapter/VsProjectExtensions.cs
+++ b/Common/Product/TestAdapter/VsProjectExtensions.cs
@@ -81,6 +81,9 @@ namespace Microsoft.VisualStudioTools.TestAdapter
         /// </summary>
         public static bool IsTestProject(this IVsProject project, Guid projectGuid)
         {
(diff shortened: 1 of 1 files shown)


paulvanbrenk:
@msft-cwells can you take a look

Common/Product/TestAdapter/VsProjectExtensions.cs
```diff
@@ -87,6 +94,21 @@ public static bool IsTestProject(this IVsProject project, Guid projectGuid)
             return (projectTypeGuids.IndexOf(projectGuid.ToString(), StringComparison.OrdinalIgnoreCase) >= 0);
         }
 
+        public static bool IsTestAdapaterEnabled()
```
msft-cwells: I just made some changes in my VS PR to cache this value, since "IsTestProject" gets called many times during any single TestAdapter operation.  We don't really need to hit the registry every time.   You don't need to make a similar change in NTVS, up to you.

PR #1584, opened 2017-05-18
Remove trailing whitespace

paulvanbrenk:
Remove empty whiteline at the end of most files.


amcasey:
While "modern" tuples are not a drop-in replacement for `Tuple` (since they use `ValueTuple`), this seems like a reasonable change.  Thanks for separating the commits (even though only one made it into the PR title 😛).

PR #1593, opened 2017-06-01
Vs2015u3pre3rollup

paulvanbrenk:
Fixes a bunch of issues around accessibility, and smaller UI fixes.


paulvanbrenk:
No this is still winforms.

I pulled all the localization from the page specific resx files, and added that to the general resx file. This makes it much easier to make changes in the future. (And will make the move to the new localization file much easier).

Issue #103, opened 2015-05-07
"press any key to continue" crash Win32Exception failed to start process during debugging


mousetraps:
No repro steps, but we can potentially just catch this exception.
An exception of type 'System.ComponentModel.Win32Exception' occurred in Microsoft.NodejsTools.dll but was not handled in user code

>   at Microsoft.NodejsTools.Debugger.NodeProcess.Start() in c:\myntvs\Nodejs\Product\Nodejs\Debugger\NodeProcess.cs:line 91
>    at Microsoft.NodejsTools.Debugger.NodeDebugger.Start(Boolean startListening) in c:\myntvs\Nodejs\Product\Nodejs\Debugger\NodeDebugger.cs:line 157
>    at Microsoft.NodejsTools.Debugger.DebugEngine.AD7Engine.Microsoft.VisualStudio.Debugger.Interop.IDebugEngineLaunch2.LaunchSuspended(String pszServer, IDebugPort2 port, String exe, String args, String dir, String env, String options, enum_LAUNCH_FLAGS launchFlags, UInt32 hStdInput, UInt32 hStdOutput, UInt32 hStdError, IDebugEventCallback2 ad7Callback, IDebugProcess2& process) in c:\myntvs\Nodejs\Product\Nodejs\Debugger\DebugEngine\AD7Engine.cs:line 563

mousetraps:
actually nevermind. this was caught somewhere down the line. the debugger tricked me. :smiley:

mousetraps:
![image](https://cloud.githubusercontent.com/assets/762848/7509568/22c75602-f447-11e4-84f2-4a30976d71fb.png)

mousetraps:
Okay I figured out when this issue occurs. It happens right after I switch Node.js versions using nvm, and that goes wrong for some reason (you'd also get the same message in the cmd prompt, sans the "press any key help stopped working" message. Typically another `nvm use` solves the issue.

PR #1602, opened 2017-06-09
Give usefull error when Nodejs exe is corrupt

paulvanbrenk:
Fixes issue #103

Catches the exception and gives a helpful errormessage

### DIFF
--- a/Nodejs/Product/PressAnyKey/PressAnyKey.cs
+++ b/Nodejs/Product/PressAnyKey/PressAnyKey.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
+using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


Nodejs/Product/PressAnyKey/PressAnyKey.cs
```diff
@@ -19,22 +20,35 @@ private static int Main(string[] args)
...
+            try
+            {
+                var proc = Process.Start(psi);
+                File.WriteAllText(args[1], proc.Id.ToString());
+                proc.WaitForExit();
+                exitCode = proc.ExitCode;
+            }
+            catch (Win32Exception exc)
+            {
+                Console.WriteLine($"Failed to start process: '{exc.Message}'.");
```
amcasey: Do these strings need to be localized?

Nodejs/Product/PressAnyKey/PressAnyKey.cs
```diff
@@ -19,22 +20,35 @@ private static int Main(string[] args)
...
+                Console.WriteLine("Probable cause is the Nodejs exe is corrupt, please re-install.");
```
amcasey: If we're displaying this to the user, we probably have to be more careful about names.  I think they use "Node.js".

Nodejs/Product/PressAnyKey/PressAnyKey.cs
```diff
@@ -19,22 +20,35 @@ private static int Main(string[] args)
...
+                exitCode = -1;
+            }
 
             if (args[0] == "both" ||
-                (proc.ExitCode == 0 && args[0] == "normal") ||
-                (proc.ExitCode != 0 && args[0] == "abnormal"))
+                (exitCode == 0 && args[0] == "normal") ||
```
amcasey: `args[0] == (exitCode == 0 ? "normal" : "abnormal")`?

Nodejs/Product/PressAnyKey/PressAnyKey.cs
```diff
@@ -19,22 +20,35 @@ private static int Main(string[] args)
...
 
-            var proc = Process.Start(psi);
-            File.WriteAllText(args[1], proc.Id.ToString());
-            proc.WaitForExit();
+            var exitCode = 0;
+            try
+            {
+                var proc = Process.Start(psi);
+                File.WriteAllText(args[1], proc.Id.ToString());
+                proc.WaitForExit();
```
amcasey: This can also throw `Win32Exception`.

Nodejs/Product/PressAnyKey/PressAnyKey.cs
amcasey: Apparently, nothing else in this assembly is.

Nodejs/Product/PressAnyKey/PressAnyKey.cs
amcasey: Error message to be updated.

Nodejs/Product/PressAnyKey/PressAnyKey.cs
amcasey: Error message to be updated.

Nodejs/Product/PressAnyKey/PressAnyKey.cs
paulvanbrenk: I think in that case your windows is broken...

Nodejs/Product/PressAnyKey/PressAnyKey.cs
paulvanbrenk: as discussed

Nodejs/Product/PressAnyKey/PressAnyKey.cs
paulvanbrenk: changed

Issue #1561, opened 2017-05-06
Mocha Test id depends on the file path


fforjan:
When running javascript unit test via Mocha on a TFS build, TFS is not able to detect that the test was failing in a previous build and tag as failing in the current build

##### Expected Behavior 
same test failing in consecutive build should be detected as failing in the first build.

##### Actual Behavior 
same test failing in consecutive build is not detected as failed in the first build

##### Steps to Reproduce
My assumption about this issue is about the test having some file path in its ID. the file path depends on the workspace location and is different between build agents:
Build A : 
![image](https://cloud.githubusercontent.com/assets/601810/25775121/7da1fe10-3252-11e7-8c5f-906af866697b.png)

Build A + 1
![image](https://cloud.githubusercontent.com/assets/601810/25775133/017c3f16-3253-11e7-9e01-c2b3b863d30b.png)

Which seems to be related to the fully qualified name : https://github.com/Microsoft/nodejstools/blob/master/Nodejs/Product/TestAdapter/TestFrameworks/NodejsTestInfo.cs#L42

On regular C# unit tests, you can see the ID seems to not only include the file name but the class name which should be the unique ID
![image](https://cloud.githubusercontent.com/assets/601810/25775160/dc9cf0ae-3253-11e7-84ae-d1e16b3bbd14.png)

fforjan:
Update to confirm when the test can be find in previous build :
![image](https://cloud.githubusercontent.com/assets/601810/25782762/79967bc6-3305-11e7-8b37-7907750b8c5e.png)
![image](https://cloud.githubusercontent.com/assets/601810/25782776/a726ea3a-3305-11e7-8627-ecf6382379dd.png)

paulvanbrenk:
Which version of VS and the Node Tools for VS are you using?

fforjan:
TFS 2015 build agent with VS 2015, NTVS tools is based on 1.3 plus private changes done by @ozyx  and pushed to you already.

paulvanbrenk:
@fforjan  Any progress on this work? i.e. should I push out a new 1.3 drop, or wait for this?

fforjan:
@ozyx was away and should be back tomorrow, I will discuss with him and give you a status

ozyx:
@paulvanbrenk Currently working on this now, will have a PR for you either today or tomorrow.

paulvanbrenk:
Awesome, going to take a look tomorrow. @billti feel free to swipe this from me

[...]

PR #1605, opened 2017-06-14
Use hashed file contents as unique test ID

ozyx:
Issue #1561

##### Bug
TFS unable to detect the same test failing in a previous build if the path of the `*.js` file differs.

Since a test's path is used in its `FullyQualifiedName`, the same test run on two different agents with different drive names (i.e: `D:/` vs. `C:/` drives) will not be recognized as the same test by TFS.

##### Fix
Hash the contents of the test file using SHA1 (so as not to slow down test discovery significantly) and use this as a `ModuleName` / unique ID.

Previous `FullyQualifiedId`: `C:\some\path\to\my\test.js::Suite 1 Test 1::Mocha`
New `FullyQualifiedId`: `test[60E439FF4023183BDC8C]::Suite 1 Test 1::Mocha`

**NOTE:** An assumption using this method is that two test files do not exist in the same project with the exact same name and contents.

##### Testing

1. Create sample test project
2. Copy project to a different path
3. Open both projects
4. The same test in both projects should have the same `FullyQualifiedId`

This will ensure intended TFS behavior of failed test detection in current vs. previous builds.

Please let me know if there are any improvements or changes I should make. Thanks!

### DIFF
--- a/Nodejs/Product/TestAdapter/TestExecutor.cs
+++ b/Nodejs/Product/TestAdapter/TestExecutor.cs
@@ -214,7 +214,7 @@ namespace Microsoft.NodejsTools.TestAdapter {
                 }
 
                 // All tests being run are for the same test file, so just use the first test listed to get the working dir
-                NodejsTestInfo testInfo = new NodejsTestInfo(tests.First().FullyQualifiedName);
+                NodejsTestInfo testInfo = new NodejsTestInfo(tests.First().FullyQualifiedName, tests.First().CodeFilePath);
                 var workingDir = Path.GetDirectoryName(CommonUtils.GetAbsoluteFilePath(settings.WorkingDir, testInfo.ModulePath));
 
                 foreach (var test in tests) {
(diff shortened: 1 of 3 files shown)


paulvanbrenk:
Overall this looks pretty good... I'm not sure cutting the hash in half, gives a random enough hash... and I'm sure we'll get flagged for using this algorithm, so we'll have to explain that we're not using it for encryption.

Would `GetHashCode` be good enough?

ozyx:
I'm not sure `GetHashCode` would suffice (I'm assuming you're referencing `string.GetHashCode`), as the hash needs to be unique to the file contents. Otherwise just using the file path would work fine. Is there an easier / cleaner way to hash the file contents?

ozyx:
Another way I could think to solve this problem would be:
- Use filepath relative to the solution or project location
    - **However, the executing binary doesn't know this info so it seems it would not be so straightforward...**

Let me know your thoughts on this. I'm not sure how big of an inconvenience it is getting flagged for crypto, but it seems hashing the file contents is the simplest way to get an ID unique to the test file itself.

paulvanbrenk:
Yeah, I'm thinking I should just take it as is for 1.3 and see if we can come up with something else for 1.4. I'll take another look this week and either merge it or explain any changes I want. 

Thanks!

ozyx:
Hi guys-- just wondering if you've had a chance to review this yet? We are using a private build internally currently and so far, so good. But would be nice to get an official release.
Thanks for your time!

paulvanbrenk:
I'll get a build out later today.

ozyx:
Great, thanks!

paulvanbrenk:
Here you go, please let me know if you encounter any issues.

https://github.com/Microsoft/nodejstools/releases/tag/v1.3.1

PR #1622, opened 2017-07-14
Remove no longer used menus

paulvanbrenk:
Issue #


Nodejs/Product/Nodejs/PkgCmdId.cs
```diff
@@ -2,22 +2,18 @@
 
 namespace Microsoft.NodejsTools
 {
-    internal class PkgCmdId
+    internal static class PkgCmdId
     {
         public const int cmdidReplWindow = 0x200;
         public const int cmdidOpenReplWindow = 0x201;
-        public const int cmdidOpenRemoteDebugProxyFolder = 0x202;
```
amcasey: I would have guessed that we'd keep this around as a comment to prevent the value from being reused (assuming that matters in this case?).

paulvanbrenk: They need to match values in the vcts file, so re-use is not an issue.

PR #1628, opened 2017-07-21
Added double quotes to npm search for allowing one character queries.

armanio123:
Work around for NPM registry issue where one character queries display 0 results.

Details on: https://github.com/npm/registry/issues/194

### DIFF
--- a/Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
+++ b/Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
@@ -106,7 +106,7 @@ namespace Microsoft.NodejsTools.NpmUI
 
             TelemetryHelper.LogSearchNpm();
 
-            var relativeUri = string.Format("/-/v1/search?text={0}", WebUtility.UrlEncode(filterText));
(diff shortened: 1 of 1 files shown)


billti:
Great job finding the root cause on NPM. Thanks!

bcoe:
@billti @armanio123 👋 from the ops team at npm. Unfortunately, in the short term, we'll likely need to shut down this loophole too. The reason we don't start searching until a minimum of 2 characters has been typed is to reduce CPU load on our search cluster; Queries that return a huge result set, e.g., `a`, `r`, translate into a fairly high-CPU load on our cluster, and slow response times. We will be doing a few things in the future to mitigate this, and better support your use-case:

* we'd like to build a specialized type-ahead search endpoint that is optimized specifically for this use-case.
* we will, over time, increase the size of our search cluster to better protect against slow queries.

Would love to know the use-case you're trying to solve, so that we can work with you as we write some of our next generation APIs 👍

paulvanbrenk:
@bcoe thanks for reaching out. Is the email address in your profile a good way to reach you?

bcoe:
@paulvanbrenk yes, `ben` at `npmjs.com` 👍

PR #1660, opened 2017-08-16
Don't crash when a path has invalid characters

paulvanbrenk:
Make sure we don't crash when a path has invalid characters.

### DIFF
--- a/Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
```diff
@@ -6511,15 +6511,24 @@ public virtual int ParseCanonicalName(string name, out uint itemId)
...
+                var child = FindNodeByFullPath(name);
+                if (child != null)
+                {
+                    itemId = child.HierarchyId;
+                    return VSConstants.S_OK;
+                }
+            }
+            catch (ArgumentException)
+            {
+                // This is expected when the path contains
```
amcasey: Do we want to try to return a narrower HRESULT in this case?

amcasey: If we're using Path utilities, it seems like IO exceptions might also be possible.

paulvanbrenk: Doc's say ArgumentException is the only one, since the method doesn't do any actual IO.

paulvanbrenk: The only path related return code was 'Path too long' which didn't seem appropriate.

PR #1665, opened 2017-08-17
Add output to export runner

armanio123:
Added a simple test passing message to the output.

### DIFF
--- a/Nodejs/Product/TestAdapter/TestFrameworks/ExportRunner/exportrunner.js
+++ b/Nodejs/Product/TestAdapter/TestFrameworks/ExportRunner/exportrunner.js
@@ -84,7 +84,8 @@ var run_tests = function (testCases, callback) {
             var testCase = require(test.testFile);
             result.title = test.testName;
             testCase[test.testName]();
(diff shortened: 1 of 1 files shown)


Nodejs/Product/TestAdapter/TestFrameworks/ExportRunner/exportrunner.js
```diff
@@ -85,6 +85,7 @@ var run_tests = function (testCases, callback) {
             result.title = test.testName;
             testCase[test.testName]();
             result.passed = true;
+            result.stdOut += "Test passed.\n";
```
billti: Is this what appears in the Test Explorer 'output' link? Shouldn't you do the same for when the test fails in the 'catch' clause then (maybe with the exception message in the output too)?

armanio123: Yes it will appear on the output link. The catch clause already contains a console.error that will also write to the output link. We could change it to instead append to the result.stdErr, but it will be no different.

PR #1667, opened 2017-08-22
Generate version displayed in Help/About based on Build Number

paulvanbrenk:
Generates the version shown in Help/About instead of hardcoding. (Showing 42.42.42.42 for dev build)

image

@minestarks we probably want something similar in TypeScript VS?


.gitignore
```diff
@@ -38,6 +38,7 @@ x64/
 *.vspscc
 *.vssscc
 .builds
+GeneratedHelpAboutVersion.cs
```
paulvanbrenk: Excluding the generated file, so we can reliably generate the file on the build machine.

Nodejs/Product/Nodejs/GeneratedHelpAboutVersion.tt
```diff
@@ -0,0 +1,23 @@
...
+//     the code is regenerated.
+// </auto-generated>
+//------------------------------------------------------------------------------
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.NodejsTools
+{
+    // This attribute is used to register the information needed to show this package
+    // in the Help/About dialog of Visual Studio.
+    [InstalledProductRegistration("#110", "#112", "<#= BuildVersion ?? "dev machine" #>", IconResourceID = 400)]
```
billti: Is this just a free-form string? Is it not better as "42.42.42.42" or whatever we use for the version on dev machines?

Nodejs/Product/Nodejs/GeneratedHelpAboutVersion.tt
```diff
@@ -0,0 +1,23 @@
+﻿<#@ template debug="false" hostspecific="true" language="C#" #>
```
billti: .tt files..... uuuuuuugh 😛

Nodejs/Product/Nodejs/GeneratedHelpAboutVersion.tt
paulvanbrenk: It should only show 'dev machine' when you edit the `.tt` file and save... on build, including F5 it will use the generated build version, which is `42.42.42.42` on the dev machine. So if this ever shows up, something has gone terribly wrong.

minestarks:
This is great, now do it for TypeScript please?

PR #1671, opened 2017-08-24
Remove TaskDialog code

paulvanbrenk:
We were unable to make them accessible, and they didn't add much value.
* For the Project Wizard, the assemlby is installed by default
* The NodeModules should never be added to the project
* The Max Path problem is solved with NPM 4+


Nodejs/Product/Nodejs/Project/NodejsFolderNode.cs
```diff
@@ -29,37 +33,18 @@ internal override int IncludeInProject(bool includeChildren)
...
+            if (CommonUtils.IsSameDirectory(nodeModulesPath, this.ItemNode.Url))
             {
+                ShouldIncludeNodeModulesFolderInProject();
                 return VSConstants.S_OK;
             }
             return base.IncludeInProject(includeChildren);
         }
 
-        private bool ShouldIncludeNodeModulesFolderInProject()
+        private void ShouldIncludeNodeModulesFolderInProject()
```
amcasey: This name no longer makes sense.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -586,45 +583,9 @@ internal struct LongPathInfo
...
-                            Process.Start("https://go.microsoft.com/fwlink/?LinkId=456511");
-                            break;
-                        default:
-                            System.Windows.Clipboard.SetText(e.Url);
-                            break;
-                    }
-                };
 
-                recheck:
+                Utilities.ShowMessageBox(
```
amcasey: I might be missing something, but it looks like we show the warning *before* doing the long path check.

amcasey: Also, this doesn't obviously contain the path causing the problem.

Nodejs/Product/Nodejs/SharedProject/ProjectNode.CopyPaste.cs
```diff
@@ -1113,7 +1113,6 @@ private Addition CanAddFileFromProjectReference(string projectRef, string target
                         {
                             throw;
                         }
-                        TaskDialog.ForException(this.Project.Site, e, string.Empty, this.Project.IssueTrackerUrl).ShowModal();
```
amcasey: No message box for this one?

paulvanbrenk: This only pointed the user to the a new issue page... which, without a stack or a repo project is pointless.

PR #1681, opened 2017-08-30
Replace FileSystemWatcher with IVsFileChangeEx

paulvanbrenk:
This uses less memory and plays nicer with VS

(this fixes internal bug 259728)

> Microsoft.VisualStudioTools.TestAdapter.TestFilesUpdateWatcher.AddWatch uses FileSystemWatcher to watch individual files, which is extremely inefficient, causing 25MB of virtual alloc in the solution I'm looking at. (as well as pinning tons of objects in GC heap, etc) 
> 
> If possible, please switch to using VS's file watcher system (IVsFileChangeEx). Contact Jeff Robison for more details on the VS file watcher.
> 
> Call stack:
> 392 25,640,960 System.dll!System.IO.FileSystemWatcher.Monitor
>  392 25,640,960 System.dll!System.IO.FileSystemWatcher.StartRaisingEvents
>   392 25,640,960 System.dll!System.IO.FileSystemWatcher.set_EnableRaisingEvents
>    391 25,575,424 Microsoft.NodejsTools.TestAdapter.dll!Microsoft.VisualStudioTools.TestAdapter.TestFilesUpdateWatcher.AddWatch
> 	391 25,575,424 Microsoft.NodejsTools.TestAdapter.dll!Microsoft.NodejsTools.TestAdapter.TestContainerDiscoverer.OnProjectLoaded
> 	 391 25,575,424 Microsoft.NodejsTools.TestAdapter.dll!Microsoft.VisualStudioTools.SolutionEventsListener.OnAfterOpenProject
> 	  391 25,575,424 f:\dd\ndp\clr\src\vm\comtoclrcall.cpp(874) : clr.dll!COMToCLRWorker + 1018 bytes
> 	   391 25,575,424 !0x01A5D060 SymErr = 0x7e
> 		354 23,150,592 f:\dd\src\env\msenv\core\vsslnpst.cpp(18737) : msenv.dll!CSolution::FireOnAfterOpenProject + 68 bytes
> 		37 2,424,832 f:\dd\src\env\msenv\core\vsslnpst.cpp(18091) : msenv.dll!CSolution::NotifySolutionEventsSinks<<lambda_f249d9ee35e9fc74cd77e99afb337efd> > + 95 bytes
> 
> Memspect Snapshot: \\visia-d15p4\share\openclosesln\3
>

### DIFF
--- a/Nodejs/Product/TestAdapter/TestContainerDiscoverer.cs
+++ b/Nodejs/Product/TestAdapter/TestContainerDiscoverer.cs
@@ -289,6 +289,7 @@ namespace Microsoft.NodejsTools.TestAdapter
             {
                 this.isDisposed = true;
                 this.testFilesAddRemoveListener.Dispose();
+                this.testFilesUpdateWatcher.Dispose();
                 this.solutionListener.Dispose();
             }
         }
@@ -460,7 +461,7 @@ namespace Microsoft.NodejsTools.TestAdapter
                     {
                         if (!string.IsNullOrEmpty(root) && CommonUtils.IsSubpathOf(root, p))
                         {
-                            this.testFilesUpdateWatcher.AddDirectoryWatch(root);
+                            this.testFilesUpdateWatcher.AddFolderWatch(root);
                             this.fileRootMap[p] = root;
                         }
                         else
@@ -540,7 +541,7 @@ namespace Microsoft.NodejsTools.TestAdapter
 
                             if (!string.IsNullOrEmpty(root) && CommonUtils.IsSubpathOf(root, e.File))
                             {
-                                this.testFilesUpdateWatcher.AddDirectoryWatch(root);
+                                this.testFilesUpdateWatcher.AddFolderWatch(root);
                                 this.fileRootMap[e.File] = root;
                             }
                             else
--- a/Nodejs/Product/TestAdapter/TestContainerDiscoverer.cs
+++ b/Nodejs/Product/TestAdapter/TestContainerDiscoverer.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
 using System.Collections.Generic;
@@ -34,7 +34,7 @@ namespace Microsoft.NodejsTools.TestAdapter
         private TestContainerDiscoverer([Import(typeof(SVsServiceProvider))]IServiceProvider serviceProvider, [Import(typeof(IOperationState))]IOperationState operationState)
             : this(serviceProvider,
                    new SolutionEventsListener(serviceProvider),
-                   new TestFilesUpdateWatcher(),
+                   new TestFilesUpdateWatcher(serviceProvider),
                    new TestFileAddRemoveListener(serviceProvider, Guids.NodejsBaseProjectFactory),
                    operationState)
         {
@@ -289,7 +289,6 @@ namespace Microsoft.NodejsTools.TestAdapter
             {
                 this.isDisposed = true;
                 this.testFilesAddRemoveListener.Dispose();
-                this.testFilesUpdateWatcher.Dispose();
                 this.solutionListener.Dispose();
             }
         }
--- a/Nodejs/Product/TestAdapter/TestContainerDiscoverer.cs
+++ b/Nodejs/Product/TestAdapter/TestContainerDiscoverer.cs
@@ -161,9 +161,9 @@ namespace Microsoft.NodejsTools.TestAdapter
                 return null;
             }
 
-            if (ErrorHandler.Succeeded(vsHierarchy.GetProperty(itemId, propid, out var o)))
+            if (ErrorHandler.Succeeded(vsHierarchy.GetProperty(itemId, propid, out var result)))
             {
-                return o;
+                return result;
             }
             return null;
         }
@@ -387,14 +387,10 @@ namespace Microsoft.NodejsTools.TestAdapter
                 {
                     continue;
                 }
+
                 var solution = (IVsSolution)this.serviceProvider.GetService(typeof(SVsSolution));
                 ErrorHandler.ThrowOnFailure(
-                    solution.SaveSolutionElement(
-                        0,
-                        (IVsHierarchy)project,
-                        0
-                    )
-                );
+                    solution.SaveSolutionElement((uint)__VSSLNSAVEOPTIONS.SLNSAVEOPT_SaveIfDirty, (IVsHierarchy)project, /* save entire project */ 0));
             }
         }
 
@@ -469,7 +465,7 @@ namespace Microsoft.NodejsTools.TestAdapter
                         }
                         else
                         {
-                            this.testFilesUpdateWatcher.AddWatch(p);
+                            this.testFilesUpdateWatcher.AddFileWatch(p);
                         }
                     }
                 }
@@ -506,13 +502,13 @@ namespace Microsoft.NodejsTools.TestAdapter
                     {
                         if (string.IsNullOrEmpty(root) || !CommonUtils.IsSubpathOf(root, p))
                         {
-                            this.testFilesUpdateWatcher.RemoveWatch(p);
+                            this.testFilesUpdateWatcher.RemoveFileWatch(p);
                         }
                         this.fileRootMap.Remove(p);
                     }
                     if (!string.IsNullOrEmpty(root))
                     {
-                        this.testFilesUpdateWatcher.RemoveWatch(root);
+                        this.testFilesUpdateWatcher.RemoveFolderWatch(root);
                     }
                 }
             }
@@ -549,7 +545,7 @@ namespace Microsoft.NodejsTools.TestAdapter
                             }
                             else
                             {
-                                this.testFilesUpdateWatcher.AddWatch(e.File);
+                                this.testFilesUpdateWatcher.AddFileWatch(e.File);
                             }
 
                             OnTestContainersChanged(e.Project);
@@ -563,12 +559,12 @@ namespace Microsoft.NodejsTools.TestAdapter
                             this.fileRootMap.Remove(e.File);
                             if (!this.fileRootMap.Values.Contains(root))
                             {
-                                this.testFilesUpdateWatcher.RemoveWatch(root);
+                                this.testFilesUpdateWatcher.RemoveFolderWatch(root);
                             }
                         }
                         else
                         {
-                            this.testFilesUpdateWatcher.RemoveWatch(e.File);
+                            this.testFilesUpdateWatcher.RemoveFileWatch(e.File);
                         }
 
                         // https://pytools.codeplex.com/workitem/1546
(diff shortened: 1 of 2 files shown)


Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
```diff
@@ -1,40 +1,33 @@
...
-            _fileWatchers = new Dictionary<string, FileSystemWatcher>(StringComparer.OrdinalIgnoreCase);
+            this.fileWatcher = (IVsFileChangeEx)serviceProvider.GetService<SVsFileChangeEx>();
         }
 
         public bool AddWatch(string path)
         {
             ValidateArg.NotNull(path, "path");
-
-            if (!string.IsNullOrEmpty(path))
+            if (!this.watchedFiles.ContainsKey(path) && ErrorHandler.Succeeded(this.fileWatcher.AdviseFileChange(path, (uint)_VSFILECHANGEFLAGS.VSFILECHG_Add, this, out uint cookie)))
```
billti: Are you not interested in _VSFILECHANGEFLAGS.VSFILECHG_Del events?

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
```diff
@@ -43,84 +36,47 @@ public bool AddDirectoryWatch(string path)
...
-
-        private void OnRenamed(object sender, RenamedEventArgs e)
+        int IVsFileChangeEvents.FilesChanged(uint cChanges, string[] rgpszFile, uint[] rggrfChange)
         {
-            var evt = FileChangedEvent;
-            if (evt != null)
+            for (var i = 0; i < cChanges; i++)
             {
-                evt(sender, new TestFileChangedEventArgs(null, e.FullPath, TestFileChangedReason.Renamed));
+                FileChangedEvent?.Invoke(this, new TestFileChangedEventArgs(null, rgpszFile[i], TestFileChangedReason.Changed));
```
billti: The old code had distinct events for Changed and Renamed. The new doesn't. Does this change in behavior cause any issues to subscribers?

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
```diff
@@ -43,84 +36,47 @@ public bool AddDirectoryWatch(string path)
...
             ValidateArg.NotNull(path, "path");
 
-            if (!string.IsNullOrEmpty(path))
+            if (this.watchedFiles.TryGetValue(path, out uint cookie))
             {
-                FileSystemWatcher watcher;
-
-                if (_fileWatchers.TryGetValue(path, out watcher))
+                this.watchedFiles.Remove(path);
+                if (Path.HasExtension(path))
```
billti: The presence or absence of an extension is a very error prone way to determine if something is a directory or not. Please make this code more determinisitc.

paulvanbrenk: Considering we're only dealing with `.js` or `.ts` files this works for this case. We don't want to hit the disk when we stop watching a file/folder.

I'll add a comment though.

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
paulvanbrenk: We care about all events.

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
paulvanbrenk: there was a subtle difference in behavior.

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
paulvanbrenk: Put the burden of getting this right on the user.

Nodejs/Product/TestAdapter/TestContainerDiscoverer.cs
```diff
@@ -289,7 +289,6 @@ void IDisposable.Dispose()
             {
                 this.isDisposed = true;
                 this.testFilesAddRemoveListener.Dispose();
-                this.testFilesUpdateWatcher.Dispose();
```
amcasey: What happened to the dispose call?

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
```diff
@@ -43,84 +39,71 @@ public bool AddDirectoryWatch(string path)
...
-                    watcher.Dispose();
-                }
+                this.watchedFiles.Remove(path);
+                return ErrorHandler.Succeeded(this.fileWatcher.UnadviseFileChange(cookie));
             }
+            return false;
         }
 
-        private void OnChanged(object sender, FileSystemEventArgs e)
+        public bool RemoveFolderWatch(string path)
```
amcasey: "RemoveDirectoryWatch" for symmetry with "AddDirectoryWatch"

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
```diff
@@ -1,40 +1,36 @@
...
+    internal class TestFilesUpdateWatcher : IVsFileChangeEvents
     {
-        private readonly IDictionary<string, FileSystemWatcher> _fileWatchers;
+        private readonly IDictionary<string, uint> watchedFiles = new Dictionary<string, uint>(StringComparer.OrdinalIgnoreCase);
         public event EventHandler<TestFileChangedEventArgs> FileChangedEvent;
 
-        public TestFilesUpdateWatcher()
+        private readonly IVsFileChangeEx fileWatcher;
+
+        public TestFilesUpdateWatcher(IServiceProvider serviceProvider)
```
amcasey: Personally, I think it would be cleaner for the caller to pass in the `IVsFileChangeEx`.

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
```diff
@@ -1,40 +1,36 @@
...
-                if (!_fileWatchers.ContainsKey(path) && Directory.Exists(directoryName))
-                {
-                    var watcher = new FileSystemWatcher(directoryName, filter);
-                    _fileWatchers[path] = watcher;
-
-                    watcher.Changed += OnChanged;
-                    watcher.EnableRaisingEvents = true;
-                    return true;
-                }
+                this.watchedFiles.Add(path, cookie);
```
amcasey: Presumably, there's some reason to believe calls will be sequentialized that we don't have to lock `watchedFiles`?

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
```diff
@@ -1,40 +1,36 @@
...
-            _fileWatchers = new Dictionary<string, FileSystemWatcher>(StringComparer.OrdinalIgnoreCase);
+            this.fileWatcher = serviceProvider.GetService<IVsFileChangeEx>(typeof(SVsFileChangeEx));
         }
 
-        public bool AddWatch(string path)
+        public bool AddFileWatch(string path)
         {
             ValidateArg.NotNull(path, "path");
 
-            if (!string.IsNullOrEmpty(path))
```
amcasey: Now we do watch empty paths?

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
```diff
@@ -43,84 +39,71 @@ public bool AddDirectoryWatch(string path)
...
-            GC.SuppressFinalize(this);
+            FileChangedEvent?.Invoke(this, new TestFileChangedEventArgs(null, pszDirectory, TestFileChangedReason.Changed));
+            return VSConstants.S_OK;
         }
 
-        protected virtual void Dispose(bool disposing)
+        private static TestFileChangedReason ConvertVSFILECHANGEFLAGS(uint flag)
         {
-            if (disposing && _fileWatchers != null)
+            if ((flag & (uint)(_VSFILECHANGEFLAGS.VSFILECHG_Size | _VSFILECHANGEFLAGS.VSFILECHG_Time | _VSFILECHANGEFLAGS.VSFILECHG_Attr)) != 0)
```
amcasey: Alternatively, eliminate this case and put the condition in the assert before `return TestFileChangedReason.Changed`.

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
```diff
@@ -1,40 +1,36 @@
...
-using System.IO;
+using System.Diagnostics;
+using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 
 namespace Microsoft.VisualStudioTools.TestAdapter
 {
-    internal class TestFilesUpdateWatcher : IDisposable
+    internal class TestFilesUpdateWatcher : IVsFileChangeEvents
```
amcasey: Should this type be disposable?  It seems like it ought to remove any remaining watches before it goes away.

amcasey:
I think @billti asked, but I didn't see a response: what happened to the Rename event?

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
```diff
@@ -43,84 +39,71 @@ public bool AddDirectoryWatch(string path)
...
 
-        private void OnRenamed(object sender, RenamedEventArgs e)
+        int IVsFileChangeEvents.FilesChanged(uint cChanges, string[] rgpszFile, uint[] rggrfChange)
         {
-            var evt = FileChangedEvent;
-            if (evt != null)
+            for (var i = 0; i < cChanges; i++)
             {
-                evt(sender, new TestFileChangedEventArgs(null, e.FullPath, TestFileChangedReason.Renamed));
+                FileChangedEvent?.Invoke(this, new TestFileChangedEventArgs(null, rgpszFile[i], ConvertVSFILECHANGEFLAGS(rggrfChange[i])));
```
amcasey: Skip the whole loop if `FileChangedEvent` is null?

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
paulvanbrenk: I'm going with `AddFolderWatch`

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
paulvanbrenk: Preferred keeping the existing pattern in the caller, who shouldn't care about how we implement filewatching

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
paulvanbrenk: fixed

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
paulvanbrenk: Yes, all of this happens on the UI thread

paulvanbrenk:
@amcasey can you take another look?

Nodejs/Product/TestAdapter/TestFilesUpdateWatcher.cs
```diff
@@ -9,49 +9,58 @@
 
 namespace Microsoft.VisualStudioTools.TestAdapter
 {
-    internal class TestFilesUpdateWatcher : IVsFileChangeEvents
+    internal class TestFilesUpdateWatcher : IVsFileChangeEvents, IDisposable
     {
         private readonly IDictionary<string, uint> watchedFiles = new Dictionary<string, uint>(StringComparer.OrdinalIgnoreCase);
+        private readonly IDictionary<string, uint> watchedFolders = new Dictionary<string, uint>(StringComparer.OrdinalIgnoreCase);
+
+        private bool disposed;
```
amcasey: Just use a null check on `fileWatcher`?

amcasey:
Still seems fine

PR #1686, opened 2017-08-30
Removed unused parameters

digeff:
Removed unused parameters from Chrome V2 debugger


Nodejs/Product/Nodejs/Debugger/NodeDebugProvider.cs
```diff
@@ -113,31 +113,6 @@ private void SetupDebugTargetInfoForWebkitV2Protocol(ref VsDebugTargetInfo vsDeb
         {
             // todo: refactor the debugging and process starting so we can re-use
 
-            var setupConfiguration = new SetupConfiguration();
```
paulvanbrenk: Does this mean we can remove a reference?

digeff: Done

PR #1687, opened 2017-08-30
Update projects to use new 'packagereference'

paulvanbrenk:


paulvanbrenk:
Having trouble making this build on the build server.

PR #1698, opened 2017-09-12
Add task to run npm install if types packages are needed

billti:
Pull request #1666 from @paulvanbrenk modified a bunch of templates to add a package.json dependency where none was previously (to install the `@types` packages instead of using the bundled `typings` definitions). However these templates did not have the task configured to run `npm install` after project creation, thus none of the `@types` packages were being installed, and the templates were in error by default once created.

This pull request adds the needed post-creation task to the templates to install the required type definitions. (See https://github.com/Microsoft/nodejstools/blob/master/Nodejs/Product/ProjectWizard/NpmWizardExtension.cs#L26 )

CC @amcasey


billti:
Fixed up all the templates to use a tsconfig.json and contain the Node.js core type definitions also.

Issue #1341, opened 2016-10-13
Cannot Installed Scoped Modules using NPM UI


mjbvz:
##### Expected Behavior

Uses should be able to install scoped modules, such as `@types/node`, using the NTVS npm UI.
##### Actual Behavior

It seems that scoped modules are not displayed in the UI:

![image](https://cloud.githubusercontent.com/assets/12821956/19371126/7ab335f4-9165-11e6-854b-4b611e2575e1.png)

I tried searching for `@types/node`, `types/node`, and just `node` but no searches seemed to return the target module.

You can install scoped modules just fine using the interactive window or the command line.

Sc0tTyXL:
Yup, same for company packages :(

paulvanbrenk:
This is fixed in Visual Studio 2017 Update 4

Issue #1340, opened 2016-10-13
Not Installed Scoped Module in Package.json Incorrectly Displayed


mjbvz:
##### Steps
1. Create a new, blank console Node JavaScript app.
2. Edit the `package.json` to include:
   
   ``` json
   "dependencies": {
     "@types/node": "6.0.x"
   }
   ```
3. Save the `package.json`.
##### Expected

`@types/node` should be listed in the npm solution explorer node, with a `"Not Installed"` message next to it.
##### Actual

`@types/node` incorrectly listed as just `node` and marked as `not listed in package.json`

![image](https://cloud.githubusercontent.com/assets/12821956/19371019/a669c574-9164-11e6-8c1f-d3305fa21357.png)

paulvanbrenk:
This is fixed in Visual Studio 2017 Update 4

Issue #1339, opened 2016-10-13
Installed Scoped Modules not In Package.json Not Displayed in NPM Node


mjbvz:
##### Steps
1. Create new blank NTVS console JS app.
2. In interactive window, run `.npm install @types/nodes`
##### Expected

`@types/node` is listed in the npm solution explorer tree. It should be listed as not included in the `package.json`.
##### Actual

Package is installed but not listed

![image](https://cloud.githubusercontent.com/assets/12821956/19370919/e89c3022-9163-11e6-957b-28dbfa7e2e53.png)

paulvanbrenk:
This is fixed in Visual Studio 2017 Update 4

Issue #635, opened 2015-12-10
Support scoped registries in .npmrc


benquarmby:
The NTVS package dialog does a good job of supporting a custom registry in `.npmrc`:

```
registry = 'http://npm.mycompany.com/'
```

However, it is often preferable to use a scope for private NPM packages and leave the public registry unchanged:

```
@mycompany:registry = 'http://npm.mycompany.com/'
```

This significantly reduces the burden on a private NPM registry, since it only has to serve requests for scoped packages.

PR #1707, opened 2017-09-26
Scoped packages

paulvanbrenk:
* No longer Quote Search queries, this fixes the issue where scoped packages couldn't be installed using the UI
* Use the /<packagename>/latest endpoint for single letter queries
* Correctly infer the name for packages not yet installed, this only affects scoped packages

This fixes at least: 

#1341, #1340, #1339, #635

### DIFF
--- a/Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
+++ b/Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
@@ -106,135 +106,86 @@ namespace Microsoft.NodejsTools.NpmUI
 
             TelemetryHelper.LogSearchNpm();
 
-            string relativeUri;
             if (filterText.Length == 1)
             {
-                // Special case since the search API won't retur results for 
-                // single chararcter queries.
-                relativeUri = $"/{WebUtility.UrlEncode(filterText)}/latest";
-
-                using (var response = await SearchNpmRegistry(relativeUri))
-                {
-                    /* We expect the following response
-                      {
-                        "name": "express",
-                        "scope": "unscoped",
-                        "version": "4.15.2",
-                        "description": "Fast, unopinionated, minimalist web framework",
-                        "keywords": [ "express", "framework", "sinatra", "web", "rest", "restful", "router", "app", "api" ],
-                        "date": "2017-03-06T13:42:44.853Z",
-                        "links": {
-                          "npm": "https://www.npmjs.com/package/express",
-                          "homepage": "http://expressjs.com/",
-                          "repository": "https://github.com/expressjs/express",
-                          "bugs": "https://github.com/expressjs/express/issues"
-                        },
-                        "author": {
-                          "name": "TJ Holowaychuk",
-                          "email": "tj@vision-media.ca"
-                        },
-                        "publisher": {
-                          "username": "dougwilson",
-                          "email": "doug@somethingdoug.com"
-                        },
-                        "maintainers": [
-                          {
-                            "username": "dougwilson",
-                            "email": "doug@somethingdoug.com"
-                          }
-                        ]
-                      }*/
-                    var reader = new StreamReader(response.GetResponseStream());
-                    using (var jsonReader = new JsonTextReader(reader))
-                    {
-                        while (jsonReader.Read())
-                        {
-                            switch (jsonReader.TokenType)
-                            {
-                                case JsonToken.StartObject:
-                                    var token = JToken.ReadFrom(jsonReader);
-                                    var package = ReadPackage(token, new NodeModuleBuilder());
-                                    if (package != null)
-                                    {
-                                        return new[] { package };
-                                    }
-                                    break;
-                                default:
-                                    throw new InvalidOperationException("Unexpected json token.");
-                            }
-                        }
-                    }
-                }
+                return await QueryNpmForSingleCharAsync(filterText);
             }
             else
             {
-                relativeUri = $"/-/v1/search?text={WebUtility.UrlEncode(filterText)}";
+                return await QueryNpmAsync(filterText);
+            }
+        }
 
-                using (var response = await SearchNpmRegistry(relativeUri))
-                {
-                    /* We expect the following response:
-                     {
-                          "objects": [
-                            {
-                              "package": {
-                                "name": "express",
-                                "scope": "unscoped",
-                                "version": "4.15.2",
-                                "description": "Fast, unopinionated, minimalist web framework",
-                                "keywords": [ "express", "framework", "sinatra", "web", "rest", "restful", "router", "app", "api" ],
-                                "date": "2017-03-06T13:42:44.853Z",
-                                "links": {
-                                  "npm": "https://www.npmjs.com/package/express",
-                                  "homepage": "http://expressjs.com/",
-                                  "repository": "https://github.com/expressjs/express",
-                                  "bugs": "https://github.com/expressjs/express/issues"
-                                },
-                                "author": {
-                                  "name": "TJ Holowaychuk",
-                                  "email": "tj@vision-media.ca"
-                                },
-                                "publisher": {
-                                  "username": "dougwilson",
-                                  "email": "doug@somethingdoug.com"
-                                },
-                                "maintainers": [
-                                  {
-                                    "username": "dougwilson",
-                                    "email": "doug@somethingdoug.com"
-                                  }
-                                ]
-                              },
-                              "score": {
-                                "final": 0.9549640105248649,
-                                "detail": {
-                                  "quality": 0.9427473299991661,
-                                  "popularity": 0.9496544159654299,
-                                  "maintenance": 0.9707450455348992
-                                }
-                              },
-                              "searchScore": 100000.95
-                            }
-                          ],
-                          "total": 10991,
-                          "time": "Tue May 09 2017 22:41:07 GMT+0000 (UTC)"
-                        }
-                    */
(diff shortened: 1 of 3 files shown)


Nodejs/Product/Npm/SPI/RootPackage.cs
```diff
@@ -11,11 +12,11 @@ namespace Microsoft.NodejsTools.Npm.SPI
     internal class RootPackage : IRootPackage
     {
         public RootPackage(
-            string fullPathToRootDirectory,
-            bool showMissingDevOptionalSubPackages,
-            Dictionary<string, ModuleInfo> allModules = null,
-            int depth = 0,
-            int maxDepth = 1)
+             string fullPathToRootDirectory,
```
amcasey: Extra whitespace?

Nodejs/Product/Npm/SPI/RootPackage.cs
```diff
@@ -47,6 +48,20 @@ internal class RootPackage : IRootPackage
...
+
+            if (this.PackageJson != null)
+            {
+                this.Name = this.PackageJson.Name;
+            }
+            else
+            {
+                // this is the root package so the full folder name after node_nodules is the name
+                // of the package
+                var index = this.Path.IndexOf(NodejsConstants.NodeModulesFolder, StringComparison.OrdinalIgnoreCase);
```
amcasey: Does it need to be surrounded by path separators?  What if there's a "node_modules2" folder?

Nodejs/Product/Npm/SPI/RootPackage.cs
```diff
@@ -47,6 +48,20 @@ internal class RootPackage : IRootPackage
...
+                var name = this.Path.Substring(index + NodejsConstants.NodeModulesFolder.Length);
+
+                this.Name = name.TrimStart('\\', '/');
```
amcasey: Why not just add 1 to the start index?

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -183,6 +244,14 @@ private void Run()
             throw new InvalidOperationException("Unexpected json token.");
         }
 
+        private static Task<WebResponse> SearchNpmRegistry(string relativeUri)
```
amcasey: SearchNpmRegistryAsync?

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -106,74 +106,135 @@ private void Run()
...
-                          "searchScore": 100000.95
                         }
-                      ],
-                      "total": 10991,
-                      "time": "Tue May 09 2017 22:41:07 GMT+0000 (UTC)"
                     }
-                */
+                }
+            }
+            else
```
amcasey: The code shared by these cases appears to be insignificant.  Would it be cleaner to split it into two separate methods?

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -106,74 +106,135 @@ private void Run()
 
             TelemetryHelper.LogSearchNpm();
```
amcasey: We don't care which version of search we used?

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -106,74 +106,135 @@ private void Run()
...
-
-            var request = WebRequest.Create(searchUri);
-            using (var response = await request.GetResponseAsync())
+            string relativeUri;
+            if (filterText.Length == 1)
             {
-                /* We expect the following response:
-                 {
-                      "objects": [
+                // Special case since the search API won't retur results for
```
amcasey: "return"

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -106,74 +106,135 @@ private void Run()
...
-            var request = WebRequest.Create(searchUri);
-            using (var response = await request.GetResponseAsync())
+            string relativeUri;
+            if (filterText.Length == 1)
             {
-                /* We expect the following response:
-                 {
-                      "objects": [
+                // Special case since the search API won't retur results for 
+                // single chararcter queries.
```
amcasey: "character"

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -106,74 +106,135 @@ private void Run()
...
+                          "email": "doug@somethingdoug.com"
+                        },
+                        "maintainers": [
+                          {
+                            "username": "dougwilson",
+                            "email": "doug@somethingdoug.com"
+                          }
+                        ]
+                      }*/
+                    var reader = new StreamReader(response.GetResponseStream());
```
amcasey: `StreamReader` appears to be disposable.

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -106,74 +106,135 @@ private void Run()
...
-                              }
-                            ]
-                          },
-                          "score": {
-                            "final": 0.9549640105248649,
-                            "detail": {
-                              "quality": 0.9427473299991661,
-                              "popularity": 0.9496544159654299,
-                              "maintenance": 0.9707450455348992
+                            switch (jsonReader.TokenType)
```
amcasey: Consider using an `if`, rather than a single-case `switch`.

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -106,74 +106,135 @@ private void Run()
...
+                            {
+                                case JsonToken.StartObject:
```
amcasey: Do we want to throw if this happens a second time?

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -106,74 +106,135 @@ private void Run()
...
+                                    var token = JToken.ReadFrom(jsonReader);
+                                    var package = ReadPackage(token, new NodeModuleBuilder());
+                                    if (package != null)
+                                    {
+                                        return new[] { package };
+                                    }
+                                    break;
+                                default:
+                                    throw new InvalidOperationException("Unexpected json token.");
```
amcasey: Would it be interesting to know something about the bad token?

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
amcasey: Never mind - I see that most of the "code" is actually sample responses.

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -106,74 +106,135 @@ private void Run()
 
             TelemetryHelper.LogSearchNpm();
 
-            var relativeUri = string.Format("/-/v1/search?text=\"{0}\"", WebUtility.UrlEncode(filterText));
-            var searchUri = new Uri(defaultRegistryUri, relativeUri);
-
-            var request = WebRequest.Create(searchUri);
-            using (var response = await request.GetResponseAsync())
+            string relativeUri;
```
amcasey: There's no reason to share this across cases.

Nodejs/Product/Npm/SPI/RootPackage.cs
paulvanbrenk: I believe we always construct the path to contain 'node_modules' but wrapping in path separators should cover that.

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
paulvanbrenk: Nope

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
paulvanbrenk: No, don't really want to build a state machine for an assert

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
paulvanbrenk: yes

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
paulvanbrenk: yep

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
amcasey: Could you not just assert that the stream is done after the object is read?

paulvanbrenk: I don't really care if there's more stuff send by the NPM api after the package I care about... And considering they can (and have) changed the response without warning, i prefer to be as lenient as possible.

amcasey: We appear to throw if we see anything other than an object. 😛

paulvanbrenk: True, since that's the thing we care about... so silently failing would be worse.

PR #1719, opened 2017-10-04
Prop page update

paulvanbrenk:
Prepare property pages to make theming easier for a future Visual Studio update.

Because we now derive from a baseclass I could delete a lot of code, which we previously copied from that baseclass.

### DIFF
--- a/Nodejs/Product/Nodejs/Nodejs.csproj
+++ b/Nodejs/Product/Nodejs/Nodejs.csproj
@@ -118,7 +118,7 @@
     <Reference Include="Microsoft.VisualStudio.Debugger.Interop.11.0, Version=11.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
     <Reference Include="Microsoft.VisualStudio.Debugger.InteropA, Version=9.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
     <Reference Include="Microsoft.VisualStudio.Designer.Interfaces, Version=1.0.5000.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
-      <EmbedInteropTypes>True</EmbedInteropTypes>
+      <EmbedInteropTypes>False</EmbedInteropTypes>
     </Reference>
     <Reference Include="Microsoft.VisualStudio.Editor, Version=$(VSTarget).0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
     <Reference Include="Microsoft.VisualStudio.Language.Intellisense, Version=$(VSTarget).0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
@@ -1185,6 +1185,9 @@
   <ItemGroup>
     <PackageReference Include="MicroBuild.Core">
       <Version>0.2.0</Version>
+    </PackageReference>
+    <PackageReference Include="Microsoft.VisualStudio.AppDesigner">
+      <Version>15.3.0-rc-6162104</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.VisualStudio.Telemetry">
       <Version>15.3.799-masterDDDBA9E4</Version>
--- a/Nodejs/Product/Nodejs/Project/NodejsGeneralPropertyPage.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsGeneralPropertyPage.cs
@@ -1,7 +1,8 @@
-// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
 using System.Runtime.InteropServices;
+using System.Windows.Forms;
 using Microsoft.VisualStudioTools;
 using Microsoft.VisualStudioTools.Project;
 
@@ -10,14 +11,21 @@ namespace Microsoft.NodejsTools.Project
     [Guid("62E8E091-6914-498E-A47B-6F198DC1873D")]
     internal class NodejsGeneralPropertyPage : CommonPropertyPage
     {
-        private readonly NodejsGeneralPropertyPageControl _control;
+        private readonly NodejsGeneralPropertyPageControl control;
 
         public NodejsGeneralPropertyPage()
         {
-            this._control = new NodejsGeneralPropertyPageControl(this);
+            this.control = new NodejsGeneralPropertyPageControl(this);
         }
 
-        public override System.Windows.Forms.Control Control => this._control;
+        protected override Control CreateControl()
+        {
+            return this.control;
+        }
+
+        public override Control Control => this.control;
+
+        protected override Type ControlType => typeof(NodejsGeneralPropertyPageControl);
 
         internal override CommonProjectNode Project
         {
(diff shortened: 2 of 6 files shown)


Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -118,7 +118,7 @@
     <Reference Include="Microsoft.VisualStudio.Debugger.Interop.11.0, Version=11.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
     <Reference Include="Microsoft.VisualStudio.Debugger.InteropA, Version=9.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
     <Reference Include="Microsoft.VisualStudio.Designer.Interfaces, Version=1.0.5000.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
-      <EmbedInteropTypes>True</EmbedInteropTypes>
+      <EmbedInteropTypes>False</EmbedInteropTypes>
```
paulvanbrenk: build complained about this.

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -1186,6 +1186,9 @@
     <PackageReference Include="MicroBuild.Core">
       <Version>0.2.0</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.VisualStudio.AppDesigner">
```
paulvanbrenk: New package that contains the baseclass for the property pages.

Nodejs/Product/Nodejs/Nodejs.csproj
amcasey: `EmbedInteropTypes` scares me - consider finding someone knowledgeable to confirm.

Nodejs/Product/Nodejs/Project/NodejsGeneralPropertyPageControl.cs
```diff
@@ -49,7 +50,7 @@ private void LocalizeLabels()
             this._startBrowser.Text = Resources.PropertiesStartBrowser;
 
             this._browsePath.AccessibleName = Resources.PropertiesBrowsePathAccessibleName;
-            this._browsePath.AccessibleName = Resources.PropertiesBrowseDirectoryAccessibleName;
```
amcasey: Oops. 😄

Nodejs/Product/Nodejs/Project/NodejsGeneralPropertyPageControl.cs
```diff
@@ -6,11 +6,12 @@
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Windows.Forms;
+using Microsoft.VisualStudio.Editors.PropertyPages;
 using Microsoft.VisualStudioTools.Project;
 
 namespace Microsoft.NodejsTools.Project
 {
-    internal partial class NodejsGeneralPropertyPageControl : UserControl
```
amcasey: Consider sealing this class.

Nodejs/Product/Nodejs/SharedProject/CommonPropertyPage.cs
```diff
@@ -1,342 +1,27 @@
...
     /// <summary>
     /// Base class for property pages based on a WinForm control.
     /// </summary>
-    public abstract class CommonPropertyPage : IPropertyPage
+    public abstract class CommonPropertyPage : PropPageBase
     {
-        private IPropertyPageSite site;
-        private CommonProjectNode _project;
-        private bool dirty;
+        protected override Size DefaultSize { get; set; } = new Size(800, 600);
```
amcasey: It seems very strange that this is writeable.

Nodejs/Product/Nodejs/SharedProject/CommonPropertyPage.cs
```diff
@@ -1,342 +1,27 @@
...
-            Utilities.ArgumentNotNull("pRect", pRect);
-
-            var r = pRect[0];
-
-            this.Control.Location = new Point(r.left, r.top);
-            this.Control.Size = new Size(r.right - r.left, r.bottom - r.top);
-        }
-
-        void IPropertyPage.SetObjects(uint count, object[] punk)
+        protected override void SetObjects(uint count, object[] punk)
```
amcasey: Can we do better than "punk"?

Nodejs/Product/Nodejs/SharedProject/CommonPropertyPage.cs
```diff
@@ -347,9 +32,9 @@ void IPropertyPage.SetObjects(uint count, object[] punk)
             {
                 if (punk[0] is ProjectConfig)
                 {
-                    if (this._project == null)
+                    if (this.Project == null)
                     {
-                        this._project = (CommonProjectNode)((CommonProjectConfig)punk.First()).ProjectMgr;
+                        this.Project = (CommonProjectNode)((CommonProjectConfig)punk.First()).ProjectMgr;
```
amcasey: Why did we test for `ProjectConfig` but cast to `CommonProjectConfig`?

Nodejs/Product/Nodejs/SharedProject/CommonPropertyPage.cs
```diff
@@ -347,9 +32,9 @@ void IPropertyPage.SetObjects(uint count, object[] punk)
             {
                 if (punk[0] is ProjectConfig)
```
amcasey: Don't forget that you can now name the result of an `is` check.

Nodejs/Product/Nodejs/SharedProject/CommonPropertyPage.cs
```diff
@@ -1,342 +1,27 @@
...
-            var r = pRect[0];
-
-            this.Control.Location = new Point(r.left, r.top);
-            this.Control.Size = new Size(r.right - r.left, r.bottom - r.top);
-        }
-
-        void IPropertyPage.SetObjects(uint count, object[] punk)
+        protected override void SetObjects(uint count, object[] punk)
         {
             if (punk == null)
```
amcasey: `|| count == 0`?

amcasey: Possibly have to clean `this.Project`?

Nodejs/Product/Nodejs/SharedProject/CommonPropertyPage.cs
```diff
@@ -347,9 +32,9 @@ void IPropertyPage.SetObjects(uint count, object[] punk)
...
                 if (punk[0] is ProjectConfig)
                 {
-                    if (this._project == null)
+                    if (this.Project == null)
                     {
-                        this._project = (CommonProjectNode)((CommonProjectConfig)punk.First()).ProjectMgr;
+                        this.Project = (CommonProjectNode)((CommonProjectConfig)punk.First()).ProjectMgr;
                     }
 
                     var configs = new List<CommonProjectConfig>();
```
amcasey: Why are we building this list?

amcasey: If there's no need to do so, it seems like the null check can be lifted out a level.

Nodejs/Product/Nodejs/Project/NodejsGeneralPropertyPage.cs
```diff
@@ -38,49 +46,42 @@ internal override CommonProjectNode Project
...
-                {
-                    startWebBrowser = true;
-                }
-                this._control.StartWebBrowser = startWebBrowser;
-            }
-            finally
+            // Attempt to parse the boolean.  If we fail, assume it is true.
+            if (!bool.TryParse(this.Project.GetUnevaluatedProperty(NodeProjectProperty.StartWebBrowser), out var startWebBrowser))
             {
-                this.Loading = false;
```
amcasey: The new impl doesn't need this?

Nodejs/Product/Nodejs/Project/NodejsGeneralPropertyPage.cs
```diff
@@ -38,49 +46,42 @@ internal override CommonProjectNode Project
...
+                startWebBrowser = true;
```
billti: Should probably fix the other bug I discussed with you before making this change, where we spin up a thread that waits forever for the Node.js web port to be listening before it opens the browser.

Nodejs/Product/Nodejs/SharedProject/CommonPropertyPage.cs
```diff
@@ -1,342 +1,27 @@
...
-        {
-            Utilities.ArgumentNotNull("pRect", pRect);
-
-            var r = pRect[0];
-
-            this.Control.Location = new Point(r.left, r.top);
-            this.Control.Size = new Size(r.right - r.left, r.bottom - r.top);
-        }
-
-        void IPropertyPage.SetObjects(uint count, object[] punk)
```
billti: That's quite a lot of removed code. We really get all that for free now? There was no special handling in there and the new base class does all the right things? Definitely want to get some good testing on the various scenarios here.

Nodejs/Product/Nodejs/SharedProject/CommonPropertyPage.cs
paulvanbrenk: they're either configs or properties... so not really.

Nodejs/Product/Nodejs/SharedProject/CommonPropertyPage.cs
paulvanbrenk: yep

Nodejs/Product/Nodejs/SharedProject/CommonPropertyPage.cs
paulvanbrenk: bad copy/paste

Nodejs/Product/Nodejs/SharedProject/CommonPropertyPage.cs
paulvanbrenk: yep

Nodejs/Product/Nodejs/SharedProject/CommonPropertyPage.cs
paulvanbrenk: Unsure.. I've not been able to recreate the case where objects == null and count != 0... so i'm thinking the check for 0 is superfluous, but i'm not confident enough to remove.

PR #1729, opened 2017-10-09
Interactive window

paulvanbrenk:
* Make sure we don't hit a null arg exception when doing telemetry
* add info command to interactive window to help debugging issues

### DIFF
--- a/Nodejs/Product/Nodejs/Repl/InfoReplCommand.cs
+++ b/Nodejs/Product/Nodejs/Repl/InfoReplCommand.cs
@@ -1,0 +1,45 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel.Composition;
+using System.Diagnostics;
+using System.Threading.Tasks;
+
+namespace Microsoft.NodejsTools.Repl
+{
+    [Export(typeof(IReplCommand))]
+    internal class InfoReplCommand : IReplCommand
+    {
+        public Task<ExecutionResult> Execute(IReplWindow window, string arguments)
+        {
+            if (window.Evaluator is NodejsReplEvaluator nodeEval)
+            {
+                try
+                {
+                    var nodeExePath = nodeEval.NodeExePath;
+                    if (string.IsNullOrEmpty(nodeExePath))
+                    {
+                        nodeExePath = nodeEval.GetNodeExePath();
+                    }
+                    var nodeVersion = FileVersionInfo.GetVersionInfo(nodeExePath);
+
+                    window.WriteLine($"Using Node.js exe from: '{nodeExePath}'");
+                    window.WriteLine($"Node.js Version: {nodeVersion.ProductVersion}");
+                }
+                catch(Exception e)
+                {
+                    window.WriteLine("Failed to retrieve Nodejs.exe information.");
+                    window.WriteError(e);
+                }
+            }
+
+            return ExecutionResult.Succeeded;
+        }
+
+        public string Description => Resources.ReplInfoDescription;
+
+        public string Command => "info";
+
+        public object ButtonContent => null;
+    }
+}
--- a/Nodejs/Product/Nodejs/Repl/InfoReplCommand.cs
+++ b/Nodejs/Product/Nodejs/Repl/InfoReplCommand.cs
@@ -9,7 +9,7 @@ using System.Threading.Tasks;
 namespace Microsoft.NodejsTools.Repl
 {
     [Export(typeof(IReplCommand))]
-    internal class InfoReplCommand : IReplCommand
+    internal sealed class InfoReplCommand : IReplCommand
     {
         public Task<ExecutionResult> Execute(IReplWindow window, string arguments)
         {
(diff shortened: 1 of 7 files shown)


Nodejs/Product/Nodejs/Repl/InfoReplCommand.cs
```diff
@@ -0,0 +1,45 @@
...
+    [Export(typeof(IReplCommand))]
+    internal class InfoReplCommand : IReplCommand
+    {
+        public Task<ExecutionResult> Execute(IReplWindow window, string arguments)
+        {
+            if (window.Evaluator is NodejsReplEvaluator nodeEval)
+            {
+                try
+                {
+                    var nodeExePath = nodeEval.NodeExePath;
```
armanio123: Prefer conditional operator
var nodeExePath = string.IsNullOrEmpty(nodeEval.NodeExePath) ? nodeEval.GetNodeExePath() : nodeEval.NodeExePath;

Nodejs/Product/Nodejs/Repl/InfoReplCommand.cs
```diff
@@ -0,0 +1,45 @@
...
+                    if (string.IsNullOrEmpty(nodeExePath))
```
armanio123: IsNullOrWhiteSpace preferred or necessary?

Nodejs/Product/Nodejs/Repl/InfoReplCommand.cs
paulvanbrenk: no

Nodejs/Product/Nodejs/Repl/InfoReplCommand.cs
paulvanbrenk: no, we probably want to fail in that case, and give the user a clear error message

Nodejs/Product/Nodejs/Repl/InfoReplCommand.cs
```diff
@@ -0,0 +1,46 @@
...
+                    {
+                        nodeExePath = nodeEval.GetNodeExePath();
+                    }
+                    var nodeVersion = FileVersionInfo.GetVersionInfo(nodeExePath);
+
+                    window.WriteLine(string.Format(CultureInfo.CurrentUICulture, Resources.ReplNodeInfo, nodeExePath));
+                    window.WriteLine(string.Format(CultureInfo.CurrentUICulture, Resources.ReplNodeVersion, nodeVersion.ProductVersion));
+                }
+                catch(Exception e)
```
amcasey: Is this boilerplate or do we actually expect an exception?  If the main failure mode is a bad path, it seems like detecting it and printing a message to that effect would be more helpful than a stack trace.

Nodejs/Product/Nodejs/Repl/InfoReplCommand.cs
```diff
@@ -0,0 +1,46 @@
...
+using System;
+using System.ComponentModel.Composition;
+using System.Diagnostics;
+using System.Globalization;
+using System.Threading.Tasks;
+
+namespace Microsoft.NodejsTools.Repl
+{
+    [Export(typeof(IReplCommand))]
+    internal class InfoReplCommand : IReplCommand
```
amcasey: Nit: `sealed`

Nodejs/Product/Nodejs/Repl/NodejsReplEvaluator.cs
```diff
@@ -39,6 +39,8 @@ public NodejsReplEvaluator(INodejsReplSite site)
             this._site = site;
         }
 
+        public string NodeExePath { get; private set; }
```
amcasey: Shouldn't this compute and then cache `GetNodeExePath`?  Why would that be the caller's responsibility?

Nodejs/Product/TestAdapter/TestExecutor.cs
```diff
@@ -195,7 +195,7 @@ private void LogTelemetry(int testCount, Version nodeVersion, bool isDebugging)
             //todo: when we have support for the Node 8 debugger log which version of the debugger people are actually using
 
             var defaultSession = TelemetryService.DefaultSession;
```
amcasey: This could probably be inlined. :)

Nodejs/Product/Nodejs/Repl/InfoReplCommand.cs
paulvanbrenk: We already check when setting the path in a bunch of places. This is a backstop to not crash the interactive window when trying to get diagnostics.

Nodejs/Product/Nodejs/Repl/NodejsReplEvaluator.cs
```diff
@@ -39,6 +39,20 @@ public NodejsReplEvaluator(INodejsReplSite site)
             this._site = site;
         }
 
+        private string nodeExePath;
+
+        public string NodeExePath
+        {
+            get
+            {
+                if (string.IsNullOrEmpty(this.nodeExePath))
```
amcasey: Should this be a straight null check?  Wouldn't empty string indicate that it had already run (and failed)?

Nodejs/Product/Nodejs/Repl/NodejsReplEvaluator.cs
```diff
@@ -39,6 +39,20 @@ public NodejsReplEvaluator(INodejsReplSite site)
...
+                {
+                    this.nodeExePath = this.GetNodeExePath();
```
amcasey: Is any thread safety required here?

Nodejs/Product/Nodejs/Repl/NodejsReplEvaluator.cs
```diff
@@ -143,6 +152,30 @@ private void EnsureConnected()
...
         }
 
+        /// <summary>
+        /// Checks if Node.js Exe is installed correctly.
+        /// Writes an error message if it's not.
+        /// </summary>
+        /// <returns></returns>
+        public bool EnsureNodeInstalled()
+        {
+            if (string.IsNullOrWhiteSpace(this.NodeExePath))
```
amcasey: Null is unexpected, correct?

Nodejs/Product/Nodejs/Repl/NodejsReplEvaluator.cs
paulvanbrenk: Yes, but when a new project is loaded we want to try again, maybe they configured the Node location there correct.

Nodejs/Product/Nodejs/Repl/NodejsReplEvaluator.cs
paulvanbrenk: no, initialization is all on the UI thread.

Issue #1735, opened 2017-10-20
Test adapter always fails if codepage is UTF-8 using vstest.console.exe


mbp:
Background:
We are using VSTS to run unit tests. We use the "Visual Studio Test" Build Task for `*.njsproj` projects. This invokes `vstest.console.exe`

A few days ago, the build stopped working without us making any changes. The reason: A new version of of the VSTS Build Task was released with this commit: https://github.com/Microsoft/vsts-tasks/commit/88417fb0ee15e0af69d2f439dff4e69b82fc0166

For some reason `vstest.console.exe` is now running in UTF-8 codepage. And this causes vstest.console.exe combined with Node.Js tools to fail.

##### Expected Behavior 

Tests can be run using vstest.console.exe and pass, even when code page is UTF-8

##### Actual Behavior 

All tests fail for no apparent reason. Output:

> Microsoft (R) Test Execution Command Line Tool Version 15.0.26720.2
> Copyright (c) Microsoft Corporation.  All rights reserved.
>
> Starting test execution, please wait...
> Information: Processing: C:\Users\mbp\Desktop\TestProject\app.tests.js
>
> Information:   Creating TestCase:C:\Users\mbp\Desktop\TestProject\app.tests.js::unit test example::Mocha
>
> Information: Processing finished for framework of Mocha
>
> Failed   unit test example
> Results File: C:\Users\mbp\TestResults\mbp_EC2AMAZ-DAD7KPJ 2017-10-20 08_59_18.trx
>
> Total tests: 1. Passed: 0. Failed: 1. Skipped: 0.
> Test Run Failed.
> Test execution time: 0.6170 Seconds

- NTVS Version: 1.4.10918.1
- Visual Studio Version: 2017 15.4.1
- Node.js Version: 6.10

##### Steps to Reproduce
1. Start command prompt
2. Call `c:\Windows\system32\chcp.com 65001`
3. Call `vstest.console.exe Projectfile.njsproj /TestAdapterPath:C:\PathToNodeJsTestAdapter`

Below you can see the concrete files I used:

package.json
```
{
  "name": "test-project",
  "version": "0.0.0",
  "description": "test-project",
  "main": "app.js",
  "author": {
    "name": "test-project"
  },
  "devDependencies": {
    "mocha": "^3.2.0"
  }
}
```

app.tests.js
```
describe('unit test', function () {
    it('example', function () {
    });
});
```

TestProject.njsproj
```
[...]

PR #1746, opened 2017-10-24
Unit test fixes

paulvanbrenk:
Issue #1735

This PR fixes 3 issues. 
* The issue described in 1735, due to a change in VSTS the unit tests now run with a UTF-8 console, this causes a BOM to be added to the json, which crashes the JSON.parse call... 
* Due to a recent change in `vstest.console.exe`, the Node Tools Test Adapter is no longer found when run from the command line. Renaming the projects to match the new expected name (i.e. test adapter assemblies end with `*.testadapter.dll`, fixes that.
* And we add a fallback option to figure out the root of the VS installation, so we can find all the required targets in the project file when parsing.


Nodejs/Product/TestAdapterImpl/TestFrameworks/run_tests.js
```diff
@@ -9,6 +9,12 @@ var rl = readline.createInterface({
 
 rl.on('line', (line) => {
     rl.close();
+
+    // strip the BOM in case of UTF-8
+    if (line.charCodeAt(0) === 0xFEFF) {
```
amcasey: Strictly speaking, this should only happen on the first line.

Nodejs/Product/TestAdapter/Properties/AssemblyInfo.cs
```diff
@@ -1,15 +1,11 @@
...
-[assembly: AssemblyTitle("Node.js Tools for Visual Studio - Test Explorer Integration")]
+[assembly: AssemblyTitle("Node.js Tools for Visual Studio - Test Explorer Integration Shims")]
 [assembly: AssemblyDescription("")]
 [assembly: AssemblyConfiguration("")]
 
 [assembly: ComVisible(false)]
-[assembly: ProvideCodeBase(AssemblyName = "Microsoft.NodejsTools.TestAdapter", CodeBase = "Microsoft.NodejsTools.TestAdapter.dll", Version = "1.0.0.0")]
-
-[assembly: InternalsVisibleTo("TestAdapterTests, PublicKey=002400000480000094000000060200000024000052534131000400000100010007d1fa57c4aed9f0a32e84aa0faefd0de9e8fd6aec8f87fb03766c834c99921eb23be79ad9d5dcc1dd9ad236132102900b723cf980957fc4e177108fc607774f29e8320e92ea05ece4e821c0a5efe8f1645c4c0c93c1ab99285d622caa652c1dfad63d745d6f2de5f17e5eaf0fc4963d261c8a12436518206dc093344d5ad293")]
+//[assembly: ProvideCodeBase(AssemblyName = "Microsoft.NodejsTools.TestAdapter", CodeBase = "Microsoft.NodejsTools.TestAdapter.dll", Version = "1.0.0.0")]
```
amcasey: Delete?

Nodejs/Product/TestAdapterImpl/TestDiscoverer.cs
```diff
@@ -35,6 +35,11 @@ public void DiscoverTests(IEnumerable<string> sources, IDiscoveryContext discove
 
             var env = new Dictionary<string, string>();
             var root = Environment.GetEnvironmentVariable(NodejsConstants.NodeToolsVsInstallRootEnvironmentVariable);
+            if (string.IsNullOrEmpty(root))
+            {
+                root = Environment.GetEnvironmentVariable("VSINSTALLDIR");
+            }
```
amcasey: Consider extracting a helper method.

amcasey:
Are the project renames just for consistency?  It seems like only the assembly names actually matter.

PR #1754, opened 2017-11-03
Re-order the project property page

paulvanbrenk:
This is in preparation for some other changes.

image

Validated tab-order, narrator, and zoom behavior


Nodejs/Product/Nodejs/Project/NodejsGeneralPropertyPageControl.cs
```diff
@@ -273,15 +255,21 @@ private void PortChanged(object sender, EventArgs e)
...
+            this._nodeExeErrorProvider.SetError(this._workingDir, errorMessage);
+
+            Changed(sender, e);
+
+            bool ValidateWorkingDir(string workingDir)
             {
-                this._nodeExeErrorProvider.SetError(this._workingDir, string.Empty);
+                if (workingDir.Contains("$("))
+                {
+                    return true;
```
amcasey: This is because it's impossible to validate, rather than because it's valid, correct?

paulvanbrenk: Yes

Issue #1360, opened 2016-10-19
Exception In Interactive Window when Holding Alt and Pressing Up Arrow Key 


mjbvz:
##### Steps to Reproduce
1. Open NTVS interactive window
2. Enter a line of text and press enter to evaluate it.
3. On the next input, hold alt and press the up arrow key.
##### Expected Behavior

Noop
##### Actual Behavior

![image](https://cloud.githubusercontent.com/assets/12821956/19532914/86f1121e-95f2-11e6-9b7d-9d747311aa01.png)

This exception does not seem to come from our code, but from VS. I also do not see any details in the activitylog file.

Alt plus arrow normally moves text in a VS text editor. We may need to disable this behavior in the interactive window.

paulvanbrenk:
Fixed moving to shared VS Interactive Window

Issue #1130, opened 2016-07-06
.clear Command in Interactive Window is Broken


mjbvz:
Discovered during our mob programming session writing tests for the interactive window (#1063)
##### Expected Behavior

In the interactive window:

```
> .clear
> 2
```

Should print `2`
##### Actual Behavior

Output is `TypeError: sandbox argument must have been converted to a context`

mjbvz:
One workaround is resetting the interactive window if you get into this state.

paulvanbrenk:
This works as expected with the move to the VS Interactive Window

PR #1778, opened 2017-11-14
Move to shared InteractiveWindow base

paulvanbrenk:
Instead of depending on a fork of the shared interactive window move to consume the VS interactive window as a nuget package. 

This fixes a number of issues #1360, and #1130. And makes it easier to add improvements in the future, e.g. IntelliSense, and use a debug console.

Note: for readability this PR doesn't actually delete the no longer needed code.


Nodejs/Product/Nodejs/NodejsPackage.cs
```diff
@@ -180,31 +181,26 @@ private void MakeDebuggerContextAvailable()
...
+            if (toolWindowType == Guids.NodejsInteractiveWindow)
             {
-                window = (IReplWindow2)provider.CreateReplWindow(
-                    this.ReplContentType,
-                    Resources.InteractiveWindowTitle,
-                    Guids.TypeScriptLanguageInfo,
-                    NodejsReplEvaluatorProvider.NodeReplId
-                );
+                var model = (IComponentModel)GetService(typeof(SComponentModel));
+                var replProvider = model.GetService<InteractiveWindowProvider>();
```
amcasey: Consider extracting a method that returns the repl provider.

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -1196,6 +1191,9 @@
     <PackageReference Include="Microsoft.VisualStudio.Threading">
       <Version>15.5.13-beta</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.VisualStudio.VsInteractiveWindow">
+      <Version>2.3.0</Version>
+    </PackageReference>
```
amcasey: This is roughly equivalent to a packages.json entry?

paulvanbrenk: Yes, the 'new' way to add nuget packages...

Nodejs/Product/Nodejs/NodejsPackage.cs
paulvanbrenk: yes

Issue #1728, opened 2017-10-09
Need a way to restore npm packages in open-folder solutions


amcasey:
When you Open Folder, there's no `npm` node in Solution Explorer, so there's no way to restore npm packages (without resorting to the command line).  It would be helpful to have some way to trigger this through the Solution Explorer.

jgwinner:
Just open and save package.json. That'll trigger it, depending on your project settings - EVEN IF you do not have a project open!

I've found that in my case it runs npm automatically, and with apparently slightly different versions. Mucked up my node files all over the place.

There IS a setting in the project-settings to tell it not to run npm, but I thought if I was just opening package.json as a text file (Run VS2017-> File Open), it wouldn't 'execute' project options. I was wrong!

== John ==

amcasey:
@jgwinner, I opened this on your behalf because I understood you were asking for a menu item to explicitly trigger npm install.  Please let me know if that's not the case.

jgwinner:
Doh! Sorry had too many bugs open. That is correct. I was trying to help you out, LOL.

It's actually sort of the opposite symptom - when you open package.json, no matter what, it messes with your npm packages. 

As you mention, a good fix would be an optional way to trigger the package rebuild. Thank you! I would normally assume this would be part of a "build" stage as with everything else. Then, the build settings can stay with the current defaults.

== John ==

paulvanbrenk:
We def. need to make this easier for users to control... Install on Save is not the most discoverable.

jgwinner:
Exactly!

Thanks, you guys are very responsive, much appreciated.

== John ==

ihuchukwu:
> Just open and save package.json. That'll trigger it, depending on your project settings - EVEN IF you do not have a project open!
> 
> I've found that in my case it runs npm automatically, and with apparently slightly different versions. Mucked up my node files all over the place.
> 
> There IS a setting in the project-settings to tell it not to run npm, but I thought if I was just opening package.json as a text file (Run VS2017-> File Open), it wouldn't 'execute' project options. I was wrong!
> 
> == John ==

Thanks John. that was perfect!

PR #1781, opened 2017-11-15
Add support for npm commands in folder view

paulvanbrenk:
Adds a right-click menu item to package.json with the following commands:
    * Install Missing Packages
    * Update Packages
    * Install New Package (opens UI)

Fixes: #1728


Nodejs/Product/Nodejs/Workspace/NpmActionProviderFactory.cs
```diff
@@ -0,0 +1,138 @@
...
+                var projectHome = Path.GetDirectoryName(this.PackageJsonPath);
+
+                var npmController = NpmControllerFactory.Create(
+                      projectHome,
+                      NodejsConstants.NpmCachePath);
+
+                using (var npmWorker = new NpmWorker(npmController))
+                using (var manager = new NpmPackageInstallWindow(npmController, npmWorker))
+                {
+                    manager.ShowModal();
```
amcasey: >manager.ShowModal(); [](start = 20, length = 20)

This is non-blocking?

Nodejs/Product/Nodejs/Workspace/NpmActionProviderFactory.cs
```diff
@@ -0,0 +1,138 @@
...
+            this.CommandId = commandId;
+            this.PackageJsonPath = packageJsonPath;
+        }
+
+        public FileContext Source { get; }
+
+        public string PackageJsonPath { get; }
+
+        // Unused, since we also implement IVsCommandItem the Display Name used comes from the vsct file
+        public string DisplayName => "";
```
amcasey: >"" [](start = 37, length = 2)

Throw not supported?

billti:
As discussed, it would be great to add a "Build" command if package.json contains a 'build' script entry.

paulvanbrenk:
@billti going to make a separate PR for that.

paulvanbrenk:
@amcasey can you take another look? I added an output window wrapper so the output from the NPM commands is shown.

amcasey:
@paulvanbrenk Doesn't ntvs already have code for writing to the output window?  Should some of it be shared?

paulvanbrenk:
@amcasey the code we have doesn't work with the threading requirements in the Open Folder case... so this is another wrapper.

billti:
Those last couple comments sound sub-optimal. How much of this code is largely a duplication of code that already exists?

paulvanbrenk:
@billti I can maybe save 20 lines in OutputWindowWrapper...

Nodejs/Product/Nodejs/Workspace/NpmActionProviderFactory.cs
```diff
@@ -0,0 +1,168 @@
...
+            public FileContextActionProvider(IWorkspace workspaceContext, OutputPaneWrapper outputPane)
+            {
+                this.workspaceContext = workspaceContext;
+                this.outputPane = outputPane;
+            }
+
+            public async Task<IReadOnlyList<IFileContextAction>> GetActionsAsync(string filePath, FileContext fileContext, CancellationToken cancellationToken)
+            {
+                await this.workspaceContext.JTF.SwitchToMainThreadAsync();
+                this.outputPane.ShowWindow();
```
amcasey: Why is this called here?

Nodejs/Product/Nodejs/Workspace/NpmActionProviderFactory.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        private abstract class NpmPackagesAction : IFileContextAction, IVsCommandItem
+        {
+            public NpmPackagesAction(string packageJsonPath, FileContext source, uint commandId, FileContextActionProvider provider)
+            {
+                this.Source = source;
+                this.CommandId = commandId;
+                this.PackageJsonPath = packageJsonPath;
+                this.WriteLine = provider.outputPane.WriteLine;
+                this.JTF = provider.workspaceContext.JTF;
```
amcasey: Where is this used?

Nodejs/Product/Nodejs/Workspace/NpmActionProviderFactory.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            private void WriteNpmOutput(object sender, NpmLogEventArgs e)
+            {
+                if (!string.IsNullOrWhiteSpace(e.LogText))
+                {
+                    this.WriteLine(e.LogText.TrimEnd('\r', '\n'));
+                }
+            }
+
+            protected readonly Action<string> WriteLine;
```
amcasey: Personally, I would expose a method to subclasses.

Nodejs/Product/Nodejs/Workspace/NpmActionProviderFactory.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            protected readonly JoinableTaskFactory JTF;
+
+            public FileContext Source { get; }
+
+            public string PackageJsonPath { get; }
+
+            // Unused, since we also implement IVsCommandItem the Display Name used comes from the vsct file
+            public string DisplayName => null;
```
amcasey: Throw NotSupportedException?

Nodejs/Product/Nodejs/Workspace/NpmActionProviderFactory.cs
paulvanbrenk: Since they're all private classes.. I'm ok with this.

Nodejs/Product/Nodejs/Workspace/NpmActionProviderFactory.cs
paulvanbrenk: breaks anycode

Nodejs/Product/Nodejs/Workspace/NpmActionProviderFactory.cs
paulvanbrenk: to ensure the output pane is initialized... then we no longer need to switch to the UI thread when invoking any commands.

Nodejs/Product/Nodejs/Workspace/NpmActionProviderFactory.cs
amcasey: You can probably just pass the wrapper.

Nodejs/Product/Nodejs/Workspace/NpmActionProviderFactory.cs
amcasey: Comment to that effect?

PR #1795, opened 2017-11-27
Simplify the API to read package.json files

paulvanbrenk:
Remove unnecessary complex API.

Replaces interface and 4 derived classes with 2 new methods on a factory class. Didn't change any of the logic in the implementation. That would be a next step which would include perf, improvements to project load.


Nodejs/Product/Npm/PackageJsonFactory.cs
```diff
@@ -1,14 +1,109 @@
...
 namespace Microsoft.NodejsTools.Npm
 {
-    public class PackageJsonFactory
+    public static class PackageJsonFactory
     {
-        public static IPackageJson Create(IPackageJsonSource source)
+        public static IPackageJson Create(string fullPathToFile)
         {
-            return source.Package == null ? null : new PackageJson(source.Package);
+            if (string.IsNullOrEmpty(fullPathToFile) && !fullPathToFile.EndsWith("package.json"))
```
amcasey: Should this check be case-insensitive?

amcasey: Also, I would find a null check clearer than `IsNullOrEmpty`.

Nodejs/Product/Npm/PackageJsonFactory.cs
```diff
@@ -1,14 +1,109 @@
...
+                            return Create(reader);
+                        }
+                    }
+                    catch (PackageJsonException pje)
+                    {
+                        WrapExceptionAndRethrow(pje);
+                    }
+                    catch (Exception exc) when (exc is IOException || exc is UnauthorizedAccessException)
+                    {
+                        if (attempts <= 0) { throw; }
```
amcasey: Would it make sense to check `attempts` as part of the exception filter?  That would have the advantage of preventing the stack from being unwound.

Nodejs/Product/Npm/PackageJsonFactory.cs
paulvanbrenk: Yes, that should be case-insensitive

Nodejs/Product/Npm/SPI/RootPackage.cs
```diff
@@ -25,7 +25,7 @@ internal class RootPackage : IRootPackage
             {
                 if (packageJsonFile.Length < 260)
                 {
-                    this.PackageJson = PackageJsonFactory.Create(new DirectoryPackageJsonSource(fullPathToRootDirectory));
```
amcasey: Does the code handle `PackageJson` being null?

Nodejs/Product/Npm/PackageJsonFactory.cs
```diff
@@ -1,14 +1,109 @@
...
+
+The following error occurred:
+
+{0}", ae.Message),
+                    ae);
+            }
+
+            return source == null ? null : new PackageJson(source);
+
+            void WrapExceptionAndRethrow(Exception ex)
```
amcasey: Why pull this out but not the one for ArgumentException?  I'd probably inline it.

Nodejs/Product/Npm/PackageJsonFactory.cs
```diff
@@ -1,14 +1,109 @@
...
+                    retryInterval *= 2; // exponential backoff
+                }
+            }
+
+            return null;
+
+            void WrapExceptionAndRethrow(Exception ex)
+            {
+                throw new PackageJsonException(
+                    string.Format(CultureInfo.CurrentCulture, @"Error reading package.json at '{0}': {1}", fullPathToFile, ex.Message),
```
amcasey: Probably redundant to say "package.json" when it's going to appear at the end of the path anyway.

Nodejs/Product/Npm/PackageJsonFactory.cs
```diff
@@ -1,14 +1,109 @@
...
+                    ex);
+            }
+        }
+
+        public static IPackageJson Create(TextReader reader)
+        {
+            object source = null;
+            try
+            {
+                var text = reader.ReadToEnd();
```
amcasey: I'm pretty sure Newtonsoft can deserialize from a reader without the need for this temporary string.

Nodejs/Product/Npm/SPI/RootPackage.cs
paulvanbrenk: yes

Nodejs/Product/Npm/PackageJsonFactory.cs
paulvanbrenk: nope

Nodejs/Product/Npm/PackageJsonFactory.cs
```diff
@@ -1,14 +1,71 @@
...
+                throw new ArgumentException("Expected full path to 'package.json' file.", nameof(fullPathToFile));
+            }
+
+            if (File.Exists(fullPathToFile))
+            {
+                var retryInterval = 500;
+                var attempts = 5;
+
+                // populate _source with retries for recoverable errors.
+                while (--attempts >= 0)
```
amcasey: Out of curiosity, why isn't this a `for` loop?

Nodejs/Product/Npm/PackageJsonFactory.cs
```diff
@@ -1,14 +1,71 @@
...
+                {
+                    try
+                    {
+                        using (var fin = new FileStream(fullPathToFile, FileMode.Open, FileAccess.Read, FileShare.Read))
+                        using (var reader = new StreamReader(fin))
+                        {
+                            object source = null;
+
+                            var text = reader.ReadToEnd();
```
amcasey: If we're just going to read the whole file, would it be easier to call `File.ReadAllText`?

paulvanbrenk: `File.ReadAllText` doesn't seem to set `FileMode`, `FileAccess`, or `FileShare`.

amcasey: 1) I'm not that concerned, since it lets go when it's done.
2) I'm pretty sure it does.

PR #1806, opened 2017-12-14
Npm scripts

paulvanbrenk:
Add support for arbitrary npm scripts in AnyCode

image

There is some moving functionality around, since you can't combine a `INodeExtender` and a `IFileContextActionProvider`, but we need to use the `INodeExtender` to allow a dynamic number of scripts in the npm context menu


amcasey:
We can't synthesize an "NPM" node under the folder in Solution Explorer?

paulvanbrenk:
@amcasey maybe... but not in scope for now.

amcasey:
Which part of the screenshot am I supposed to be looking at?

paulvanbrenk:
@amcasey the NPM flyout

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+                            return VSConstants.S_OK;
+
+                        case PkgCmdId.cmdidWorkSpaceNpmUpdate:
+                            ExecNpmUpdate(node);
+                            return VSConstants.S_OK;
+                    }
+
+                    if (nCmdID >= PkgCmdId.cmdidWorkSpaceNpmDynamicScript && nCmdID < PkgCmdId.cmdidWorkSpaceNpmDynamicScriptMax)
+                    {
+                        ExecDynammic(node, nCmdID);
```
billti: Typo. Only one m in dynamic.

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+                    }
+                }
+
+                return false;
+            }
+
+            private bool QueryDynamic(uint nCmdID, string filePath, ref string customTitle)
+            {
+                var index = nCmdID - PkgCmdId.cmdidWorkSpaceNpmDynamicScript;
+                var packageJson = PackageJsonFactory.Create(filePath);
```
billti: Does this end up re-reading the package.json file for every script?

billti: Also, does it re-read every time you right click? Or is there a change it could show stale scripts when you right-click if the package.json file got modified?

Nodejs/Product/Nodejs/NodejsTools.vsct
```diff
@@ -225,32 +247,42 @@
...
+      
+      <!-- workspace -->
+      <Button guid="guidNodeToolsWseContextMenuCmdSet" id="cmdidWorkSpaceNpmDynamicScript" priority="0x0130" type="Button">
+        <CommandFlag>DynamicItemStart</CommandFlag>
+        <CommandFlag>NoShowOnMenuController</CommandFlag>
+        <CommandFlag>DefaultInvisible</CommandFlag>
+        <CommandFlag>DynamicVisibility</CommandFlag>
+        <Strings>
+          <ButtonText>Start Script</ButtonText>
+          <CommandName>StartScript</CommandName>
```
billti: What is this "StartScript" about?

billti:
I assume this has no impact on behavior when loading an .njsproj project?

Nodejs/Product/Nodejs/PkgCmdId.cs
```diff
@@ -24,8 +24,12 @@ internal static class PkgCmdId
         public const int cmdidNpmOpenModuleHomepage = 0x306;
         public const int menuIdNpm = 0x3000;
 
-        public const int cmdidWorkSpaceNpmInstallMissing = 0x4001;
-        public const int cmdidWorkSpaceNpmInstallNew = 0x4002;
-        public const int cmdidWorkSpaceNpmUpdate = 0x4003;
+        public const int cmdidWorkSpaceNpmInstallMissing = 0x0200;
```
amcasey: >0x0200 [](start = 59, length = 6)

This won't collide with `cmdidReplWindow` because it's in a submenu?

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -377,6 +379,67 @@ public void Dispose()
...
+                quoteArgs: false,
+                outputEncoding: redirector == null ? null : Encoding.UTF8
+                ))
+            {
+                var whnd = process.WaitHandle;
+                if (whnd == null)
+                {
+                    // Process failed to start, and any exception message has
+                    // already been sent through the redirector
+                    if (redirector != null)
```
amcasey: >if (redirector != null) [](start = 20, length = 23)

Alternatively, use `?.`

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -377,6 +379,67 @@ public void Dispose()
...
+                        redirector.WriteErrorLine("Error - cannot start npm");
+                    }
+                }
+                else
+                {
+                    var i = await Task.Run(() => WaitHandle.WaitAny(new[] { whnd }));
+                    if (i == 0)
+                    {
+                        Debug.Assert(process.ExitCode.HasValue, "npm process has not really exited");
+                        process.Wait();
```
amcasey: >process.Wait(); [](start = 24, length = 15)

What does this wait for?  The assert suggests the process has already exited.

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -377,6 +379,67 @@ public void Dispose()
...
+                    // Process failed to start, and any exception message has
+                    // already been sent through the redirector
+                    if (redirector != null)
+                    {
+                        redirector.WriteErrorLine("Error - cannot start npm");
+                    }
+                }
+                else
+                {
+                    var i = await Task.Run(() => WaitHandle.WaitAny(new[] { whnd }));
```
amcasey: >WaitHandle.WaitAny(new[] { whnd  [](start = 49, length = 32)

Why not use `whnd.WaitOne()`?

Nodejs/Product/Nodejs/PkgCmdId.cs
```diff
@@ -24,8 +24,12 @@ internal static class PkgCmdId
...
+        public const int cmdidWorkSpaceNpmInstallNew = 0x0201;
+        public const int cmdidWorkSpaceNpmUpdate = 0x0202;
+        public const int cmdidWorkSpaceNpmDynamicScript = 0x1000;
+
+        // only allow 100 scripts
+        public const int cmdidWorkSpaceNpmDynamicScriptMax = cmdidWorkSpaceNpmDynamicScript + 100;
```
amcasey: >100 [](start = 94, length = 3)

It might be nice to synthesize a bunch of dummy scripts and confirm that something sane happens just under/over the limit.

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -377,6 +379,67 @@ public void Dispose()
...
+                    if (i == 0)
+                    {
+                        Debug.Assert(process.ExitCode.HasValue, "npm process has not really exited");
+                        process.Wait();
+                        if (process.StandardOutputLines != null)
```
amcasey: >StandardOutputLines [](start = 36, length = 19)

I've had problems with this method of redirecting output.  In particular, if the output buffer fills up, the process may pause until something consumes it (which won't happen until it completes, which won't happen until something consumes the output, ...).  I have a sample implementation of a different pattern, if you're interested.

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
```diff
@@ -377,6 +379,67 @@ public void Dispose()
...
+                        {
+                            standardOutputLines = process.StandardOutputLines.ToList();
```
amcasey: >ToList [](start = 78, length = 6)

Why `ToList`?  To force evaluation?  To make a copy?

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            // Note: all the Exec commands are async, this allows us to call them in a fire and forget
+            // pattern, without blocking the UI or losing any logging
+
+            private async void ExecNpmInstallMissing(WorkspaceVisualNodeBase node)
+            {
+                using (var npmController = this.CreateController(node.Workspace))
+                using (var commander = npmController.CreateNpmCommander())
+                {
+                    await commander.Install();
```
amcasey: Does this update the status bar or something?  How does the user know it's working?

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+                // doesn't support running arbitrary scripts. And changing that is outside
+                // the scope of these changes.
+
+                var index = nCmdID - PkgCmdId.cmdidWorkSpaceNpmDynamicScript;
+                var packageJson = PackageJsonFactory.Create(((IFileNode)node).FullPath);
+
+                Debug.Assert(packageJson != null, "Failed to create package.json");
+
+                var scripts = packageJson.Scripts;
+                if (index < scripts.Length)
```
amcasey: >if (index < scripts.Length) [](start = 16, length = 27)

Can this fail?

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+                {
+                    var npmPath = NpmHelpers.GetPathToNpm();
+
+                    await NpmWorker.ExecuteNpmCommandAsync(
+                        npmPath,
+                        executionDirectory: node.Workspace.Location,
+                        arguments: new[] { "run-script", scripts[index].CommandName },
+                        visible: true); // show the CMD window
```
amcasey: >// show the CMD window [](start = 40, length = 22)

If it shows a window, should this method return `Task`?

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+                }
+            }
+
+            public bool QueryStatus(List<WorkspaceVisualNodeBase> selection, Guid pguidCmdGroup, uint nCmdID, ref uint cmdf, ref string customTitle)
+            {
+                if (selection.Count != 1 || !EnsurePackageJson(selection.First()))
```
amcasey: >First [](start = 73, length = 5)

`Single` would be more explicit.

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+                }
+
+                if (pguidCmdGroup == Guids.NodeToolsWorkspaceCmdSet)
+                {
+                    switch (nCmdID)
+                    {
+                        // we can always install missing, new, and update npm packages
+                        case PkgCmdId.cmdidWorkSpaceNpmInstallMissing:
+                        case PkgCmdId.cmdidWorkSpaceNpmInstallNew:
+                        case PkgCmdId.cmdidWorkSpaceNpmUpdate:
```
amcasey: >cmdidWorkSpaceNpmUpdate [](start = 38, length = 23)

Does update make sense if nothing is installed?

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+                            cmdf = (uint)(OLECMDF.OLECMDF_SUPPORTED | OLECMDF.OLECMDF_ENABLED);
+                            return true;
+                    }
+
+                    // Each id we return true for, this method is called with id + 1
+                    // this way we can add each script to the context menu
+                    if (nCmdID >= PkgCmdId.cmdidWorkSpaceNpmDynamicScript && nCmdID < PkgCmdId.cmdidWorkSpaceNpmDynamicScriptMax)
```
amcasey: >cmdidWorkSpaceNpmDynamicScriptMax [](start = 95, length = 33)

Generally, I would expect "Max" to be included in the range, so this might be a misnomer.

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+
+                        if (QueryDynamic(nCmdID, node.FullPath, ref customTitle))
+                        {
+                            cmdf = (uint)(OLECMDF.OLECMDF_SUPPORTED | OLECMDF.OLECMDF_ENABLED);
+                            return true;
+                        }
+                    }
+                }
+
+                return false;
```
amcasey: >return false; [](start = 16, length = 13)

Does `cmdf` need to be populated?

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+            }
+
+            private bool QueryDynamic(uint nCmdID, string filePath, ref string customTitle)
+            {
+                var index = nCmdID - PkgCmdId.cmdidWorkSpaceNpmDynamicScript;
```
amcasey: >nCmdID - PkgCmdId.cmdidWorkSpaceNpmDynamicScript [](start = 28, length = 48)

Consider extracting a helper that maps filePath and nCmdID to script.

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+                var scripts = packageJson.Scripts;
+                if (index < scripts.Length)
+                {
+                    customTitle = $"npm run-script {scripts[index].CommandName}";
+                    return true;
+                }
+                return false;
+            }
+
+            internal static bool EnsurePackageJson(WorkspaceVisualNodeBase node)
```
amcasey: >internal [](start = 12, length = 8)

Why is this internal?

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+            {
+                return (node is IFileNode fileNode && StringComparer.OrdinalIgnoreCase.Equals(fileNode.FileName, "package.json"));
+            }
+
+            private INpmController CreateController(IWorkspace workspace)
```
amcasey: >private [](start = 12, length = 7)

`static`?

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
paulvanbrenk: Yes, we do... looks like there isn't a better way, without running the risk of stale commands. (And considering a package.json is not huge, this is probably fine).

Nodejs/Product/Nodejs/NodejsTools.vsct
paulvanbrenk: It's a placeholder string, which'll be replaced in the Query calls with the actual name of the scripts.

Nodejs/Product/Nodejs/PkgCmdId.cs
paulvanbrenk: No, they only to be unique per group.

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
paulvanbrenk: private now

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
paulvanbrenk: done

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
paulvanbrenk: no

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
paulvanbrenk: changed

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
paulvanbrenk: yeah, works the same way as in the project.. also we don't have a reliable way of checking if packages are installed and the newest version.

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
paulvanbrenk: sure

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
paulvanbrenk: Generally no, unless an external process changes the `package.json` while the menu is open.... e..g git pull

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
paulvanbrenk: Yes, and write to the output window

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
amcasey: Um, shouldn't we ignore *any* index if that happens?  What if someone deletes script 0?  The user probably doesn't want the script after the one they clicked to run.

Nodejs/Product/Nodejs/NpmUI/NpmWorker.cs
paulvanbrenk: wrapper does the right thing

Nodejs/Product/Nodejs/Workspace/ContextMenuProvider.cs
paulvanbrenk: yeah, but considering how unlikely this scenario is...

PR #1809, opened 2017-12-15
Put the progressbar back in the Npm UI

paulvanbrenk:
Add the progress bar back when installing npm packages from the npm UI.

This fixes:

https://developercommunity.visualstudio.com/content/problem/154181/no-visual-feedback-when-installing-npm-package.html

### DIFF
--- a/Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
+++ b/Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
@@ -233,11 +233,9 @@ namespace Microsoft.NodejsTools.NpmUI
 
             if (filtered.Any())
             {
-                var rootPackage = this.npmController?.RootPackage;
-
                 newItems.AddRange(filtered.Select(package => new ReadOnlyPackageCatalogEntryViewModel(
                     package,
-                    rootPackage?.Modules[package.Name])));
+                    this.npmController?.RootPackage?.Modules[package.Name])));
             }
 
             await this.dispatcher.BeginInvoke((Action)(() =>
@@ -375,10 +373,13 @@ namespace Microsoft.NodejsTools.NpmUI
             }
             set
             {
-                this.isExecutingCommand = value;
+                if (this.isExecutingCommand != value)
+                {
+                    this.isExecutingCommand = value;
 
-                OnPropertyChanged();
-                OnPropertyChanged("ExecutionProgressVisibility");
+                    OnPropertyChanged();
+                    OnPropertyChanged(nameof(ExecutionProgressVisibility));
+                }
             }
         }
 
@@ -403,18 +404,18 @@ namespace Microsoft.NodejsTools.NpmUI
                 if (disposing)
                 {
                     this.filterTimer?.Dispose();
-                }
 
-                var controller = this.npmController;
-                if (controller != null)
-                {
-                    this.npmController.FinishedRefresh -= this.NpmController_FinishedRefresh;
-                }
-                var worker = this.npmWorker;
-                if (worker != null)
-                {
-                    this.npmWorker.CommandStarted -= this.NpmWorker_CommandStarted;
-                    this.npmWorker.CommandCompleted -= this.NpmWorker_CommandCompleted;
+                    var controller = this.npmController;
+                    if (controller != null)
+                    {
+                        controller.FinishedRefresh -= this.NpmController_FinishedRefresh;
+                    }
+                    var worker = this.npmWorker;
+                    if (worker != null)
+                    {
+                        worker.CommandStarted -= this.NpmWorker_CommandStarted;
+                        worker.CommandCompleted -= this.NpmWorker_CommandCompleted;
+                    }
                 }
                 disposed = true;
             }
--- a/Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
+++ b/Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
@@ -15,7 +15,7 @@ using Microsoft.VisualStudio.Shell;
 
 namespace Microsoft.NodejsTools.NpmUI
 {
-    internal class NpmPackageInstallViewModel : INotifyPropertyChanged
+    internal class NpmPackageInstallViewModel : INotifyPropertyChanged, IDisposable
     {
         internal enum Indices
         {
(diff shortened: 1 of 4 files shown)


Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
```diff
@@ -218,16 +233,11 @@ private void StartFilter()
 
             if (filtered.Any())
             {
-                IRootPackage rootPackage = null;
-                var controller = this.npmController;
-                if (controller != null)
-                {
-                    rootPackage = controller.RootPackage;
-                }
+                var rootPackage = this.npmController?.RootPackage;
```
amcasey: >rootPackage [](start = 20, length = 11)

Now that it's only consumed once, you could probably just inline it.

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
```diff
@@ -357,17 +367,64 @@ public PackageCatalogEntryViewModel SelectedPackage
...
+            get
+            {
+                return this.isExecutingCommand;
+            }
+            set
+            {
+                this.isExecutingCommand = value;
+
+                OnPropertyChanged();
+                OnPropertyChanged("ExecutionProgressVisibility");
```
amcasey: >ExecutionProgressVisibility [](start = 35, length = 27)

`nameof`

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
```diff
@@ -357,17 +367,64 @@ public PackageCatalogEntryViewModel SelectedPackage
...
-                handler(this, new PropertyChangedEventArgs(propertyName));
+                if (disposing)
+                {
+                    this.filterTimer?.Dispose();
+                }
+
+                var controller = this.npmController;
+                if (controller != null)
+                {
+                    this.npmController.FinishedRefresh -= this.NpmController_FinishedRefresh;
```
amcasey: >this.npmController [](start = 20, length = 18)

`controller`.  Was it extracted for fear of a race?

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
```diff
@@ -357,17 +367,64 @@ public PackageCatalogEntryViewModel SelectedPackage
...
-            if (handler != null)
+            this.PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!disposed)
             {
-                handler(this, new PropertyChangedEventArgs(propertyName));
+                if (disposing)
```
amcasey: >if (disposing) [](start = 16, length = 14)

Why doesn't this apply to the event changes as well?

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
paulvanbrenk: no the compiler complained

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
```diff
@@ -357,17 +367,64 @@ public PackageCatalogEntryViewModel SelectedPackage
...
+                }
+                disposed = true;
             }
         }
+
+        // This code added to correctly implement the disposable pattern.
+        public void Dispose()
+        {
+            // Do not change this code. Put cleanup code in Dispose(bool disposing) above.
+            Dispose(true);
```
amcasey: `GC.SuppressFinalize(this);`?

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
```diff
@@ -357,17 +367,64 @@ public PackageCatalogEntryViewModel SelectedPackage
...
+
+                var controller = this.npmController;
+                if (controller != null)
+                {
+                    this.npmController.FinishedRefresh -= this.NpmController_FinishedRefresh;
+                }
+                var worker = this.npmWorker;
+                if (worker != null)
+                {
+                    this.npmWorker.CommandStarted -= this.NpmWorker_CommandStarted;
```
amcasey: >this.npmWorker [](start = 20, length = 14)

`worker`

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
```diff
@@ -357,17 +367,64 @@ public PackageCatalogEntryViewModel SelectedPackage
...
+        {
+            if (!disposed)
             {
-                handler(this, new PropertyChangedEventArgs(propertyName));
+                if (disposing)
+                {
+                    this.filterTimer?.Dispose();
+                }
+
+                var controller = this.npmController;
```
amcasey: >var controller = this.npmController [](start = 16, length = 35)

The reason not to just set the corresponding property to null is that we don't want to fire a property changed event?

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
amcasey: Strictly speaking, this should only happen if the value actually changed.

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
paulvanbrenk: mostly worried about the VS shutdown/solution close scenario

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
paulvanbrenk: only required when you override the finalizer

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
paulvanbrenk: yep

Issue #1606, opened 2017-06-16
node_modules folder causes several minutes of lag when exploring files in VS2017


briman0094:
I'm not entirely sure if this is the right place for this question as it pertains more to VS2017 than Node.js tools, but as it is triggered by large `node_modules` folders, I figured I would ask here.

I am working on a project with a large amount of Node dependencies in the main `package.json`, and of course each dependency installs several of its own to the root of the `node_modules` folder (tree is flattened in the latest versions of NPM). I am also working on part of the project in VS Code as it has far better support for Pug and SCSS files than VS2017 does, and sometimes handles TypeScript better than ReSharper.

Because I am working in both IDEs, if I create a file in VS Code, I must add it manually to the project in VS2017 by clicking Show All Files and then "Include In Project" for each of the files I need to add. The problem with this is that when I click "Show All Files", it tries to enumerate the directory structure of the entire `node_modules` folder which freezes the whole IDE for about 5 minutes. It also freezes whenever I install or remove any packages, even if the `node_modules` folder is hidden from the Solution Explorer.

Is there any way to have the IDE completely ignore the `node_modules` folder even in "Show All Files" mode? If not, would this be possible as a feature in NTVS? I can't work effectively if the IDE freezes this often.

paulvanbrenk:
This has been called out before, and this is something the project system would have to support... as a work around, you can use the "add existing item" menu option, so you don't have to use "show all files". I do realize this is annoying when adding folders.

briman0094:
I actually figured something out since writing this...I decided to try setting the `node_modules` directory and all children as "Hidden", and it seems to have made Visual Studio ignore the folder even with "Show All Files" enabled. Of course, this isn't a perfect solution because any time I need to reinstall packages, or if I clone the project on a new machine, I have to remember to go mark `node_modules` as Hidden or I'll end up freezing my IDE for 5 minutes.

paulvanbrenk:
Can you share your `package.json` so we can try to reproduce.

uniqueiniquity:
https://developercommunity.visualstudio.com/content/problem/148195/nodejs-projects-lead-to-high-load-of-visual-studio.html seems to a related issue.

paulvanbrenk:
This should be greatly improved by  #1835

Issue #1506, opened 2017-03-16
Reverse directory scan freezes VS2017, rendering njsproj unusable for my project


lupestro:
After I build my project, VS2017 freezes for a _very_ long time.  Using procmon, I can see that this coincides with a reverse scan through the entire directory tree of my project, including the 24,362 files and 3999 folders in node_modules, resulting in a freeze that lasts for 23 1/2 minutes. As soon as this scan is over, VS2017 frees right up again, as if nothing had happened. 

I don't encounter this on non-JavaScript projects, but then I seldom have any directory trees quite as large as node_modules tends to be. So this could be a general VS2017 issue, but it will seldom be seen on projects that aren't using Node or something else with large directory trees.

Possible relevant details:

It doesn't matter whether I am kicking off my build inside or outside of the environment - the scan and the freeze seems happen about when the build is done. 

As this is an Ember build, I am already marking my tmp area hidden in the file system and, indeed, no files in tmp are being scanned, so this takes the questions of links off the table. (The end of an ember build leaves the tmp area empty anyway. An ember serve, not so much.)

The build is actually placing its results outside the project tree, into the www directory in the accompanying ASP.NET core web app project that provides my Web APIs. This other project lies outside my solution, as well, so it isn't visible in VS2017 at the time all the problems occur. Given this, I'm not sure what change triggers it to rescan my directory tree.

I have a .tfignore excluding node_modules, bower_components, tmp, and dist subtrees from TFS, and that seems to be effective, so I'm reluctant to believe Visual Studio's TFS support is involved in this.

##### Expected Behavior 

File changes as a result of builds cause the environment to notice some things in the background but it doesn't affect further editing.

##### Actual Behavior 

The IDE becomes completely unresponsive for many, many minutes.

Node.js Tools   1.3.50201.08
NPM Task Runner   1.4.81
Microsoft Visual Studio Professional 2017
Version 15.0.26228.9 D15RTWSVC
- Node.js Version:  v6.10.0 (npm 4.2.0) 

##### Steps to Reproduce
1.  Load any ember-cli project, even default
2.  Follow instructions for loading modules into npm and bower.
3.  Load NodeJS project around it, making all JS into content - this is strictly client-side.
3.  Kick off build either from a command line for from NPM Task Explorer. (Admin or not doesn't matter)
4.  Watch the fur fly.

fulls1z3:
[...]

Issue #1128, opened 2016-07-06
Poor VS performance with a large number of NPM packages in node_modules


kant2002:
##### Expected Behavior

Would be good if memory usage never goes above 500Mb.
##### Actual Behavior

Latest 10 minutes right now private increase to 1.2Gb and jumping up to 1.8Gb. VS instance become unresponive up to 10 minutes. 

- NTVS Version: 1.2.40627.01
- Visual Studio Version: VS2013 Update 3
##### Steps to Reproduce
1. Not sure how to reproduce but will take a look, if other freezes happens, will try to repro
##### Additional informaiton

Large memory objects (>10Mb)

```

02f23e98   731565     11705048 Microsoft.VisualStudio.Shell.IVsHierarchyItem[]
71039f14   731759     11708144 System.Runtime.CompilerServices.AsyncMethodBuilderCore+MoveNextRunner
27953f2c   292873     11714920 System.Collections.Generic.HashSet`1[[Microsoft.VisualStudio.PlatformUI.HierarchyItem, Microsoft.VisualStudio.Shell.TreeNavigation.HierarchyProvider]]
61072aa0   755557     12088912 Microsoft.Internal.VisualStudio.PlatformUI.HierarchyItemPair
2175a0e0   730420     14608400 Microsoft.VisualStudio.PlatformUI.HierarchyManager+<>c__DisplayClass14_0
7103a278   731676     14633520 System.Threading.Tasks.StandardTaskContinuation
71090764   389032     14673104 System.Int32[]
276ee990   753007     15060140 Microsoft.VisualStudio.PlatformUI.HierarchyItemIdentity
710474d4   731636     17559264 System.Threading.Tasks.SynchronizationContextAwaitTaskContinuation
70c62430   736272     17670528 System.Collections.Generic.List`1[[System.Object, mscorlib]]
34d19560   292978     19128000 System.Collections.Generic.HashSet`1+Slot[[Microsoft.VisualStudio.PlatformUI.HierarchyItem, Microsoft.VisualStudio.Shell.TreeNavigation.HierarchyProvider]][]
7108ed0c   766285     22365608 System.Object[]
53dbcea0   731635     23412320 System.Func`2[[System.Threading.Tasks.Task`1[[System.Int32, mscorlib]], mscorlib],[System.Threading.Tasks.Task`1[[System.Int32, mscorlib]], mscorlib]]
710848e4   733805     23481760 System.Action
[...]

PR #1827, opened 2018-01-02
Perf issues on project load and external file changes

paulvanbrenk:
These changes make loading projects significantly faster, when the user has opted out of 'Show All Files' (the default).

And no longer listening for filechanges reduces the number of full project cone scans, to sync the Solution Explorer with the Disk. (because we will hit the FileSystemWatcher error case a lot less often)

This improves #1128, #1506, and #1606

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
@@ -326,7 +326,12 @@ namespace Microsoft.NodejsTools.Project
 
                 base.Reload();
 
-                SyncFileSystem();
+                if (this.IsShowingAllFiles)
+                {
+                    // We only need to sync the files on load if we're actually showing them
+                    // otherwise we'll sync on idle.
+                    this.SyncFileSystem();
+                }
 
                 this.ModulesNode.ReloadHierarchySafe();
             }
--- a/Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
@@ -437,7 +437,8 @@ namespace Microsoft.VisualStudioTools.Project
             {
                 InternalBufferSize = 1024 * 4,  // 4k is minimum buffer size
                 IncludeSubdirectories = true
-                /* don't need to set NotifyFilters, since default LastWrite, FileName, and DirectoryName works for us */
+                // don't need to set NotifyFilters, 
+                // since the default (LastWrite, FileName, and DirectoryName) works for us
             };
 
             // Set Event Handlers
--- a/Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
@@ -44,7 +44,6 @@ namespace Microsoft.VisualStudioTools.Project
         private int suppressFileWatcherCount;
         private bool isShowingAllFiles;
         private object automationObject;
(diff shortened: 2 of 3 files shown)


Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
```diff
@@ -810,16 +795,6 @@ private bool NoPendingFileSystemRescan()
...
 
-        internal void RegisterFileChangeNotification(FileNode node, FileSystemEventHandler handler)
-        {
-            this.fileChangedHandlers[node.Url] = handler;
-        }
-
-        internal void UnregisterFileChangeNotification(FileNode node)
-        {
-            this.fileChangedHandlers.Remove(node.Url);
-        }
```
amcasey: This was (unrelated) dead code?

paulvanbrenk: Related, because it's dead we no longer need to care about the filechanges.

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
```diff
@@ -438,13 +437,12 @@ private FileSystemWatcher CreateFileSystemWatcher(string dir)
             {
                 InternalBufferSize = 1024 * 4,  // 4k is minimum buffer size
                 IncludeSubdirectories = true
+                /* don't need to set NotifyFilters, since default LastWrite, FileName, and DirectoryName works for us */
```
amcasey: >/* don't need to set NotifyFilters, since default LastWrite, FileName, and DirectoryName works for us */ [](start = 16, length = 104)

I found this comment confusing.  Is not setting `NotifyFilter` so unconventional that it requires as explanation?  If so, I'd just say "the default value" rather than listing the exact filters.

paulvanbrenk: It took me a while to figure out what the actual default was... so being explicit seemed helpful. I can reword

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
amcasey: Unrelated to the first commit, related to the second.

---
In reply to: [159298890](https://github.com/Microsoft/nodejstools/pull/1827#discussion_r159298890) [](ancestors = 159298890)

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
amcasey: In this case, it's documented on MSDN.

---
In reply to: [159307551](https://github.com/Microsoft/nodejstools/pull/1827#discussion_r159307551) [](ancestors = 159307551)

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -326,7 +326,12 @@ protected override void Reload()
 
                 base.Reload();
 
-                SyncFileSystem();
+                if (this.IsShowingAllFiles)
+                {
+                    // We only need to sync the files on load if we're actually showing them
+                    // otherwise we'll sync on idle.
+                    this.SyncFileSystem();
```
billti: Does this code get re-executed if you toggle show all files on/off? I do that frequently working in NTVS projects.

paulvanbrenk: As discussed, no. Show all files enumerates the files we know about, which is a collection that is constructed on idle. I'll ensure this collection is correct when toggling show all files. In the next PR, when we no longer do the disk enumeration on the UI thread.

PR #1832, opened 2018-01-04
Move the enumeration of the files on disk to the BG thread

paulvanbrenk:
This change moves the enumeration of files to a BG thread. This way
we no longer block the UI thread when there's a large number of changes
when the user invokes a build/test outside of VS.

### DIFF
--- a/Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
@@ -1,7 +1,10 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
+using System.Threading;
 using Microsoft.VisualStudio.Shell.Interop;
 
 namespace Microsoft.VisualStudioTools.Project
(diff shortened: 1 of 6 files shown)


Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
```diff
@@ -12,128 +15,226 @@ internal abstract partial class CommonProjectNode
...
         {
-            private readonly Stack<DirState> remainingDirs = new Stack<DirState>();
+            private readonly BlockingCollection<DirState> dirs = new BlockingCollection<DirState>(new ConcurrentStack<DirState>());
+
             private readonly CommonProjectNode project;
 
-            public DiskMerger(CommonProjectNode project, HierarchyNode parent, string dir)
+            private readonly Thread worker;
+
+            private readonly ManualResetEventSlim processingDirsEvent = new ManualResetEventSlim(initialState: true);
```
billti: Why use the *Slim rather than just ManualResetEvent? Slim uses spinning and is for when wait times are expected to be very short, which is probably unlikely if processing all directories.

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
```diff
@@ -12,128 +15,226 @@ internal abstract partial class CommonProjectNode
...
+            {
+                if (!this.isDisposed)
+                {
+                    this.processingDirsEvent.Wait();
+                }
+            }
+
+            private void Merge()
+            {
+                while (!this.isDisposed && !this.dirs.IsCompleted)
```
billti: Here, and in other places (e.g. "AddDir" above), you have no lock around checking `isDisposed`, and then accessing `this.dirs`. Thus is could be disposed in-between the check and the access in multi-threaded scenarios (which this is), causing occasional race-conditions/crashes.

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
```diff
@@ -12,128 +15,226 @@ internal abstract partial class CommonProjectNode
...
+                {
+                    if (this.dirs.TryTake(out var dir, Timeout.Infinite))
```
billti: There are a number of places outside this file where Clear() is called, which empties the queue. Isn't this just going to hang forever if that occurs then this line is called?

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
```diff
@@ -12,128 +15,226 @@ internal abstract partial class CommonProjectNode
...
+                            foreach (var file in Directory.EnumerateFiles(dir.Path))
+                            {
+                                if (this.project.IsFileHidden(file))
+                                {
+                                    continue;
+                                }
+                                var fileNode = this.project.FindNodeByFullPath(file);
+                                if (fileNode == null)
+                                {
+                                    fileNode = this.InvokeOnUIThread(() => this.project.AddAllFilesFile(dir.Parent, file));
```
billti: Is it worth trying to batch these? Seems like if a large folder (lots of files) is added, this is going to thrash back and forth between the UI & background thread to add them all individually.

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
```diff
@@ -12,128 +15,226 @@ internal abstract partial class CommonProjectNode
...
+            private readonly ManualResetEventSlim processingDirsEvent = new ManualResetEventSlim(initialState: true);
+
+            private bool isDisposed = false;
+
+            public DiskMerger(CommonProjectNode project)
             {
                 this.project = project;
-                this.remainingDirs.Push(new DirState(dir, parent));
+
+                this.worker = new Thread(this.Merge)
```
billti: We've generally tried to avoid adding more threads to devenv.exe. That was one of the reasons I did so much work to add the TaskScheduler in the language service (and move to running Tasks instead of starting threads). Any reason we can't use Tasks here?

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
billti: Actually this thread will just be sitting here... then if/when dispose is called, you'll either have an orphaned thread doing nothing waiting here, or an unhandled exception will occur (as either TryTake with throw when the underlying BlockingCollection is disposed, or if it ever continues calling processingDirsEvent.Reset with throw as that was disposed).

paulvanbrenk: Yes, in the case of an empty collection TryTake (with a timeout) will block until `AddingCompleted` is called. TryTake without a timeout does not block when the collection is empty.

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
```diff
@@ -399,27 +403,30 @@ protected override void Reload()
...
-            if (this.currentMerger == null)
-            {
-                this.currentMerger = new DiskMerger(this, this, this.ProjectHome);
-            }
-            while (this.currentMerger.ContinueMerge(this.ParentHierarchy != null))
-            {
-            }
-            this.currentMerger = null;
+            // ensure the merge is started, and wait for completion
+            // this blocks the UI thread
```
billti: Doesn't this just block whatever thread it is called on (which may or may not be the UI thread).

Also, does that risk deadlocking? If this is blocking the UI thread, but there are many places in `Merge` where you call `InvokeOnUIThread`.

paulvanbrenk:
Going to try this again with `Tasks`

PR #1835, opened 2018-01-06
Make DiskMerger class Async, so we don't block the UI thread

paulvanbrenk:
Switches to using the VS FileWatcher service, and uses async/await

### DIFF
--- a/Nodejs/Product/Nodejs/SharedProject/CommonFolderNode.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/CommonFolderNode.cs
@@ -226,7 +226,6 @@ namespace Microsoft.VisualStudioTools.Project
 
         public override void RenameFolder(string newName)
         {
-            var oldName = this.Url;
             base.RenameFolder(newName);
 
             if (this.ProjectMgr.TryDeactivateSymLinkWatcher(this))
--- a/Nodejs/Product/Nodejs/SharedProject/CommonFolderNode.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/CommonFolderNode.cs
@@ -227,15 +227,7 @@ namespace Microsoft.VisualStudioTools.Project
         public override void RenameFolder(string newName)
         {
             var oldName = this.Url;
-//            this._project.SuppressFileChangeNotifications();
-            try
-            {
-                base.RenameFolder(newName);
-            }
-            finally
-            {
- //               this._project.RestoreFileChangeNotifications();
-            }
+            base.RenameFolder(newName);
 
             if (this.ProjectMgr.TryDeactivateSymLinkWatcher(this))
             {
--- a/Nodejs/Product/Nodejs/SharedProject/CommonFolderNode.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/CommonFolderNode.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
 using System.Diagnostics;
@@ -227,14 +227,14 @@ namespace Microsoft.VisualStudioTools.Project
         public override void RenameFolder(string newName)
         {
             var oldName = this.Url;
-            this._project.SuppressFileChangeNotifications();
+//            this._project.SuppressFileChangeNotifications();
             try
             {
                 base.RenameFolder(newName);
             }
             finally
             {
-                this._project.RestoreFileChangeNotifications();
+ //               this._project.RestoreFileChangeNotifications();
             }
 
             if (this.ProjectMgr.TryDeactivateSymLinkWatcher(this))
--- a/Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
@@ -53,11 +53,28 @@ namespace Microsoft.VisualStudioTools.Project
                     return true;
                 }
 
+                return await this.InvokeOnUIThread(() => this.ContinueMergeAsyncWorker(dir, hierarchyCreated));
+            }
+
+            private async Task<bool> ContinueMergeAsyncWorker((string Name, HierarchyNode Parent) dir, bool hierarchyCreated)
+            {
                 var wasExpanded = hierarchyCreated ? dir.Parent.GetIsExpanded() : false;
                 var missingChildren = new HashSet<HierarchyNode>(dir.Parent.AllChildren);
+
+                var thread = this.project.Site.GetUIThread();
+
                 try
                 {
-                    foreach (var curDir in Directory.EnumerateDirectories(dir.Name))
+                    // enumerate disk on worker thread 
+                    var folders = await Task.Run(() =>
+                    {
+                        thread.MustNotBeCalledFromUIThread();
+                        return Directory.GetDirectories(dir.Name, "*", SearchOption.TopDirectoryOnly);
+                    }).ConfigureAwait(true);
+
+                    thread.MustBeCalledFromUIThread();
+
+                    foreach (var curDir in folders)
                     {
                         if (this.project.IsFileHidden(curDir))
                         {
@@ -78,7 +95,7 @@ namespace Microsoft.VisualStudioTools.Project
                         var existing = this.project.FindNodeByFullPath(curDir);
                         if (existing == null)
                         {
-                            existing = await this.InvokeOnUIThread(() => this.project.AddAllFilesFolder(dir.Parent, curDir + Path.DirectorySeparatorChar, hierarchyCreated));
+                            existing = this.project.AddAllFilesFolder(dir.Parent, curDir + Path.DirectorySeparatorChar, hierarchyCreated);
                         }
                         else
                         {
@@ -96,7 +113,16 @@ namespace Microsoft.VisualStudioTools.Project
                 try
                 {
                     var newFiles = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
-                    foreach (var file in Directory.EnumerateFiles(dir.Name))
+
+                    var files = await Task.Run(() =>
+                    {
+                        thread.MustNotBeCalledFromUIThread();
+                        return Directory.GetFiles(dir.Name, "*", SearchOption.TopDirectoryOnly);
+                    }).ConfigureAwait(true);
+
+                    thread.MustBeCalledFromUIThread();
+
+                    foreach (var file in files)
                     {
                         if (this.project.IsFileHidden(file))
                         {
@@ -114,7 +140,7 @@ namespace Microsoft.VisualStudioTools.Project
                         }
                     }
 
-                    await this.InvokeOnUIThread(() => AddNewFiles(dir.Parent, newFiles));
+                    AddNewFiles(dir.Parent, newFiles);
                 }
                 catch
                 {
@@ -124,17 +150,17 @@ namespace Microsoft.VisualStudioTools.Project
                     // state.  Set it back to what it was
                     if (hierarchyCreated)
                     {
-                        await this.InvokeOnUIThread(() => dir.Parent.ExpandItem(wasExpanded ? EXPANDFLAGS.EXPF_ExpandFolder : EXPANDFLAGS.EXPF_CollapseFolder));
+                        dir.Parent.ExpandItem(wasExpanded ? EXPANDFLAGS.EXPF_ExpandFolder : EXPANDFLAGS.EXPF_CollapseFolder);
                     }
                     return true;
                 }
 
                 // remove the excluded children which are no longer there
-                await this.InvokeOnUIThread(() => this.RemoveMissingChildren(missingChildren));
+                this.RemoveMissingChildren(missingChildren);
 
                 if (hierarchyCreated)
                 {
-                    await this.InvokeOnUIThread(() => dir.Parent.ExpandItem(wasExpanded ? EXPANDFLAGS.EXPF_ExpandFolder : EXPANDFLAGS.EXPF_CollapseFolder));
+                    dir.Parent.ExpandItem(wasExpanded ? EXPANDFLAGS.EXPF_ExpandFolder : EXPANDFLAGS.EXPF_CollapseFolder);
                 }
 
                 return true;
@@ -168,9 +194,9 @@ namespace Microsoft.VisualStudioTools.Project
                 return this.project.Site.GetUIThread().InvokeAsync(action);
             }
 
-            private Task<T> InvokeOnUIThread<T>(Func<T> func)
+            private Task<T> InvokeOnUIThread<T>(Func<Task<T>> func)
             {
-                return this.project.Site.GetUIThread().InvokeAsync(func);
+                return this.project.Site.GetUIThread().InvokeTask(func);
             }
         }
     }
--- a/Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
@@ -1,7 +1,10 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
+using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
+using System.Threading.Tasks;
 using Microsoft.VisualStudio.Shell.Interop;
 
 namespace Microsoft.VisualStudioTools.Project
@@ -22,13 +25,13 @@ namespace Microsoft.VisualStudioTools.Project
         /// </summary>
         private sealed class DiskMerger
         {
-            private readonly Stack<DirState> remainingDirs = new Stack<DirState>();
+            private readonly ConcurrentStack<(string Name, HierarchyNode Parent)> remainingDirs = new ConcurrentStack<(string, HierarchyNode)>();
             private readonly CommonProjectNode project;
 
             public DiskMerger(CommonProjectNode project, HierarchyNode parent, string dir)
             {
                 this.project = project;
-                this.remainingDirs.Push(new DirState(dir, parent));
+                this.remainingDirs.Push((dir, parent));
             }
 
             /// <summary>
@@ -37,17 +40,16 @@ namespace Microsoft.VisualStudioTools.Project
             /// 
             /// Returns true if the merge needs to continue, or false if the merge has completed.
             /// </summary>
-            public bool ContinueMerge(bool hierarchyCreated)
+            public async Task<bool> ContinueMergeAsync(bool hierarchyCreated)
             {
                 if (this.remainingDirs.Count == 0)
                 {
                     // all done
-                    this.project.BoldStartupItem();
+                    await this.InvokeOnUIThread(this.project.BoldStartupItem);
                     return false;
                 }
 
-                var dir = this.remainingDirs.Pop();
-                if (!Directory.Exists(dir.Name))
+                if (!this.remainingDirs.TryPop(out var dir) || !Directory.Exists(dir.Name))
                 {
                     return true;
                 }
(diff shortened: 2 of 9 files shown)


Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
```diff
@@ -801,20 +747,17 @@ protected override ReferenceContainerNode CreateReferenceContainerNode()
...
-                new FileSystemChange(this, WatcherChangeTypes.Deleted, e.OldFullPath, isRename: true),
-                new FileSystemChange(this, WatcherChangeTypes.Created, e.FullPath, isRename: true));
-        }
-
-        private void FileExistanceChanged(object sender, FileSystemEventArgs e)
+        private void FolderChangedOnDisk(object sender, FolderChangedEventArgs e)
         {
-            this.QueueFileSystemChanges(new FileSystemChange(this, e.ChangeType, e.FullPath));
+            var file = e.FileName;
+            if (File.Exists(file) || Directory.Exists(file))
```
amcasey: >if (File.Exists(file) || Directory.Exists(file)) [](start = 12, length = 48)

Why do we have to hit the disk here?  Is this the event you mentioned was inadequate?

paulvanbrenk: Yep, they currently don't report what changed to the file/folder. Hopefully in the v2 of the API. Note this is all on BG thread, so it's not as bad as it seems.

Nodejs/Product/Nodejs/SharedProject/CommonFolderNode.cs
```diff
@@ -227,14 +227,14 @@ internal override int IncludeInProjectWithProgress(bool includeChildren)
         public override void RenameFolder(string newName)
         {
             var oldName = this.Url;
-            this._project.SuppressFileChangeNotifications();
+//            this._project.SuppressFileChangeNotifications();
```
amcasey: Delete?

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
```diff
@@ -431,59 +412,18 @@ private void BoldStartupItem()
...
-                // don't need to set NotifyFilters, 
-                // since the default (LastWrite, FileName, and DirectoryName) works for us
-            };
-
-            // Set Event Handlers
-            watcher.Created += this.FileExistanceChanged;
-            watcher.Deleted += this.FileExistanceChanged;
-            watcher.Renamed += this.FileNameChanged;
-            watcher.Error += this.WatcherError;
+            this.StopWatchingFiles();
```
amcasey: >StopWatchingFiles [](start = 17, length = 17)

Inline?

Nodejs/Product/Nodejs/SharedProject/CommonFolderNode.cs
amcasey: There seems to be a bunch of commented-out code.

---
In reply to: [160538343](https://github.com/Microsoft/nodejstools/pull/1835#discussion_r160538343) [](ancestors = 160538343)

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
```diff
@@ -431,59 +412,18 @@ private void BoldStartupItem()
             }
         }
 
-        private FileSystemWatcher CreateFileSystemWatcher(string dir)
+        private void CreateFileSystemWatcher()
```
amcasey: >CreateFileSystemWatcher [](start = 21, length = 23)

Rename?

amcasey: Also, what does this do?

---
In reply to: [160539363](https://github.com/Microsoft/nodejstools/pull/1835#discussion_r160539363) [](ancestors = 160539363)

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.FileSystemChange.cs
```diff
@@ -181,7 +180,7 @@ private void ChildCreated(HierarchyNode child)
                         var folderNodeWasExpanded = folderNode.GetIsExpanded();
 
                         // then add the folder nodes
-                        this.project.MergeDiskNodes(folderNode, this.path);
+                        await this.project.MergeDiskNodesAsync(folderNode, this.path).ConfigureAwait(true);
```
amcasey: >.ConfigureAwait(true) [](start = 85, length = 21)

Why is this one special?

Nodejs/Product/Nodejs/SharedProject/FileChangeManager.cs
```diff
@@ -1,92 +1,140 @@
...
                 }
 
-                set
+                if (pszFile == null)
                 {
-                    this.fileChangeCookie = value;
+                    throw new ArgumentNullException(nameof(pszFile));
                 }
+
+                this.fileChangeManager.FolderChangedOnDisk?.Invoke(this, new FolderChangedEventArgs(pszDirectory, pszFile, (_VSFILECHANGEFLAGS)0 /* default for now, untill VS implements API that returns actual change */));
```
amcasey: >untill [](start = 165, length = 6)

Typo

Nodejs/Product/Nodejs/SharedProject/FileChangeManager.cs
```diff
@@ -116,90 +171,83 @@ public void Dispose()
...
 
             // Unsubscribe from the observed source files.
-            foreach (var info in this.observedItems.Values)
+            foreach (var (ItemID, FileChangeCookie) in this.observedFiles.Values)
             {
-                ErrorHandler.ThrowOnFailure(this.fileChangeService.UnadviseFileChange(info.FileChangeCookie));
+                var hr = this.fileChangeService.UnadviseFileChange(FileChangeCookie);
+                if (ErrorHandler.Failed(hr)) { break; }
+                // don't want to crash VS during cleanup
+                Debug.Assert(ErrorHandler.Succeeded(hr), "UnadviseFileChange failed");
```
amcasey: >Debug.Assert(ErrorHandler.Succeeded(hr), "UnadviseFileChange failed"); [](start = 16, length = 70)

Aren't succeeded and failed mutually exclusive?

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
```diff
@@ -431,59 +411,17 @@ private void BoldStartupItem()
...
-            {
-                IncludeSubdirectories = true,
-                NotifyFilter = NotifyFilters.Attributes,
-                InternalBufferSize = 1024 * 4, // 4k is minimum buffer size
-            };
-
-            // Set Event Handlers
-            watcher.Changed += this.FileAttributesChanged;
-            watcher.Error += this.WatcherError;
+            this.StopWatchingFiles();
```
amcasey: >StopWatchingFiles [](start = 17, length = 17)

Why stop first?  In case more files have come under observation in the meantime?

paulvanbrenk: This is called when the project is reloaded for various reasons, and I want to make sure we're not watching any old files.

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
```diff
@@ -1401,7 +1310,7 @@ public int GetSpecificEditorProperty(string mkDocument, int propid, out object r
                 throw new ArgumentException("Was null or empty", nameof(mkDocument));
             }
 
-            // Make sure that the document moniker passed to us is part of this project
+            // Make sure that the document moniker passed to us is part of this project
```
amcasey: >// Make sure that the document moniker passed to us is part of this project     [](start = 12, length = 79)

Whitespace

Nodejs/Product/Nodejs/SharedProject/FileChangeManager.cs
paulvanbrenk: yeah, the debug should be before the `break`

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
```diff
@@ -858,26 +790,20 @@ private void TriggerIdle()
...
-            if (!CommonUtils.HasEndSeparator(curDir))
-            {
-                curDir = curDir + Path.DirectorySeparatorChar;
-            }
-            this.symlinkWatchers[curDir] = CreateFileSystemWatcher(curDir);
+            curDir = CommonUtils.EnsureEndSeparator(curDir);
+            this.fileWatcher.ObserveFolder(curDir);
         }
 
         internal bool TryDeactivateSymLinkWatcher(HierarchyNode child)
```
amcasey: >HierarchyNode child [](start = 50, length = 19)

It's strange that this isn't symmetric with `CreateSymLinkWatcher`, but I see that that's not new.

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
```diff
@@ -22,41 +25,56 @@ internal abstract partial class CommonProjectNode
         /// </summary>
         private sealed class DiskMerger
         {
-            private readonly Stack<DirState> remainingDirs = new Stack<DirState>();
+            private readonly ConcurrentStack<(string Name, HierarchyNode Parent)> remainingDirs = new ConcurrentStack<(string, HierarchyNode)>();
```
amcasey: >ConcurrentStack [](start = 29, length = 15)

Does this still need to be a `ConcurrentStack`?

paulvanbrenk: yeah, continueMerge could be on any thread, where the worker is on the UI thread.

Nodejs/Product/Nodejs/SharedProject/CommonFolderNode.cs
```diff
@@ -227,15 +227,7 @@ internal override int IncludeInProjectWithProgress(bool includeChildren)
         public override void RenameFolder(string newName)
         {
             var oldName = this.Url;
-            this._project.SuppressFileChangeNotifications();
-            try
-            {
-                base.RenameFolder(newName);
```
amcasey: >base.RenameFolder(newName); [](start = 16, length = 27)

Does the new watcher no longer report renames?  Why is suppression no longer required?

Nodejs/Product/Nodejs/SharedProject/CommonFolderNode.cs
```diff
@@ -227,15 +227,7 @@ internal override int IncludeInProjectWithProgress(bool includeChildren)
         public override void RenameFolder(string newName)
         {
             var oldName = this.Url;
```
amcasey: >var oldName = this.Url; [](start = 12, length = 23)

Unused?

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
```diff
@@ -528,6 +466,12 @@ protected override void Dispose(bool disposing)
                     this.idleManager.OnIdle -= this.OnIdle;
                     this.idleManager.Dispose();
                 }
+
+                if (this.fileWatcher != null)
```
amcasey: >if (this.fileWatcher != null) [](start = 16, length = 29)

Can `this.fileWatcher` be null?

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
```diff
@@ -528,6 +466,12 @@ protected override void Dispose(bool disposing)
...
+                {
+                    this.fileWatcher.FolderChangedOnDisk -= this.FolderChangedOnDisk;
```
amcasey: >this.fileWatcher.FolderChangedOnDisk -= this.FolderChangedOnDisk; [](start = 20, length = 65)

The other dispose call doesn't appear to unsubscribe first.  Is the difference important?

Nodejs/Product/Nodejs/SharedProject/CommonFolderNode.cs
paulvanbrenk: We only get one notificiation in the new watcher for the older, vs. a notification for each file in the old watcher. Further the new watcher doesn't allow suppressing notifications on subfolders specifically, so we would need to stop notifications on all folders, and I don't want to miss notifications on those other folders while renaming.

Nodejs/Product/Nodejs/SharedProject/CommonFolderNode.cs
paulvanbrenk: Looks like it

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
```diff
@@ -904,7 +830,7 @@ private void OnIdle(object sender, ComponentManagerEventArgs e)
                     {
                         // we have more file merges to process, do this
                         // before reflecting any other pending updates...
-                        if (!merger.ContinueMerge(this.ParentHierarchy != null))
+                        if (!(await merger.ContinueMergeAsync(this.ParentHierarchy != null)))
```
amcasey: >if [](start = 24, length = 2)

The caller above used a while loop.  Is the difference important?

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
```diff
@@ -955,7 +881,7 @@ protected override NodeProperties CreatePropertiesObject()
 
         public override void Close()
```
amcasey: >Close [](start = 29, length = 5)

What's the relationship between `Close` and `Dispose`?  Is there a danger of leaving the object alive with a disposed `fileWatcher`?

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
```diff
@@ -22,41 +25,56 @@ internal abstract partial class CommonProjectNode
...
-                if (!Directory.Exists(dir.Name))
+                if (!this.remainingDirs.TryPop(out var dir) || !Directory.Exists(dir.Name))
                 {
                     return true;
                 }
 
+                return await this.InvokeOnUIThread(() => this.ContinueMergeAsyncWorker(dir, hierarchyCreated));
+            }
+
+            private async Task<bool> ContinueMergeAsyncWorker((string Name, HierarchyNode Parent) dir, bool hierarchyCreated)
```
amcasey: Why pass `dir` as a tuple?  It looks like it's only used for its members.

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.FileSystemChange.cs
```diff
@@ -13,7 +14,7 @@ internal abstract partial class CommonProjectNode
         /// Represents an individual change to the file system.  We process these in bulk on the
         /// UI thread.
         /// </summary>
-        [DebuggerDisplay("{nameof(FileSystemChange)}: {isRename} {Type} {path}")]
+        [DebuggerDisplay("FileSystemChange: {isRename} {Type} {path}")]
```
amcasey: >FileSystemChange [](start = 26, length = 16)

Why not `nameof`?

Nodejs/Product/Nodejs/SharedProject/FileChangeManager.cs
```diff
@@ -1,92 +1,145 @@
...
+                if (filesChanged == null)
+                {
+                    throw new ArgumentNullException(nameof(filesChanged));
+                }
+
+                if (flags == null)
                 {
-                    return this.itemID;
+                    throw new ArgumentNullException(nameof(flags));
                 }
```
amcasey: These checks seem only to detect programmer errors in VS.  Would it make more sense to assert?  If there's actually a danger of bad input, we should also check that `numberOfFilesChanged` isn't out of bounds.

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.FileSystemChange.cs
paulvanbrenk: doesn't appear to work in string literals.

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
paulvanbrenk: Looks like `Close` also calls `Dispose` so removing there

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
paulvanbrenk: Nope

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
paulvanbrenk: Yes, here we want to check if VS is still idle between each folder we process.

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
paulvanbrenk: `Close` calls `Dispose`

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.DiskMerger.cs
paulvanbrenk: convenience when passing as argument

Nodejs/Product/Nodejs/SharedProject/FileChangeManager.cs
paulvanbrenk: kept the pattern used in the `FilesChanged` method

Nodejs/Product/Nodejs/SharedProject/FileChangeManager.cs
```diff
@@ -227,27 +277,54 @@ internal void IgnoreItemChanges(string fileName, bool ignore)
...
                 // Get the cookie that was used for this.observedItems to this file.
-                var itemInfo = this.observedItems[fullFileName];
+                var (ItemID, FileChangeCookie) = value;
 
-                // Remove the file from our observed list. It's important that this is done before the call to 
-                // UnadviseFileChange, because for some reason, the call to UnadviseFileChange can trigger a 
-                // FilesChanged event, and we want to be able to filter that event away.
-                this.observedItems.Remove(fullFileName);
+                // Stop observing the file
+                ErrorHandler.ThrowOnFailure(this.fileChangeService.UnadviseFileChange(FileChangeCookie));
```
amcasey: >ThrowOnFailure [](start = 29, length = 14)

It looks like `StopObservingFolder` just returns false.  Why the difference?

Nodejs/Product/Nodejs/SharedProject/FileChangeManager.cs
```diff
@@ -227,27 +277,54 @@ internal void IgnoreItemChanges(string fileName, bool ignore)
...
+            if (this.observedFolders.TryRemove(fullFolderName, out var cookie))
+            {
                 // Stop observing the file
-                ErrorHandler.ThrowOnFailure(this.fileChangeService.UnadviseFileChange(itemInfo.FileChangeCookie));
+                return ErrorHandler.Succeeded(this.fileChangeService.UnadviseDirChange(cookie));
+            }
+            return false;
+        }
+
+        private void CheckDisposed()
```
amcasey: >CheckDisposed [](start = 21, length = 13)

Does this make sense in a class accessed from multiple threads?  It looks like it could be disposed after performing the check but before executing the actual logic.

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.cs
amcasey: I still think these could be inlined.

Nodejs/Product/Nodejs/SharedProject/CommonProjectNode.FileSystemChange.cs
amcasey: It's probably `DebuggerDisplayAttribute` that doesn't support it.

Nodejs/Product/Nodejs/SharedProject/FileChangeManager.cs
paulvanbrenk: Existing callers required that for `StopObservingFolder`, changed for `*File` since I do prefer that.

andrzejmartyna:
Hi, the layman question.
Could you not only move logic to background but also forget about "node_modules" folder completely. Just ignore, no watch, no touch, no scan, no publish, etc.

The reasons for the question are:
1) I've got some native code with NSJRPOJ project that I would prefer to not change / no dig into / no search for other IDE
2) The cost of "node_modules" for operations is very high: long publishing to Azure, consuming place, freezing VS, also I've tried to delete the folder manually from target location through FTP but it takes 30-40 minutes! (sure, I should search for better method but still it so strange to have this useless folder in the first place!)

BTW. In the doc for creating project there is an option "Exclude node_modules" which I do not see now.
So it was some issues here in the past. All ideas I've found for excluding node_modules simply do not work for NSPROJs
https://github.com/Microsoft/nodejstools/wiki/Projects

PR #1836, opened 2018-01-19
Added jsconfig template and common exclude directories to tsconfig.

armanio123:
Add jsconfig.json file to the template list and also adds the most common exclude directories to tsconfig.


Nodejs/Product/Nodejs/Templates/Files/TypeScriptTsConfig/jsconfig.vstemplate
```diff
@@ -0,0 +1,15 @@
+<VSTemplate Version="3.0.0" Type="Item" xmlns="http://schemas.microsoft.com/developer/vstemplate/2005">
+  <TemplateData>
+    <Name Package="{2ffe45c4-5c73-493c-b187-f2e955ff875e}" ID="1044" />
```
armanio123: The ID 1044 and 1045 used for the name and description respectively have not been added yet to the Visual Studio resources. Instead, VS will display the default name for both fields (jsconfig in this case) until the resource name exists.

Nodejs/Product/Nodejs/Templates/Files/TypeScriptTsConfig/tsconfig.json
```diff
@@ -8,6 +8,12 @@
     "target": "es5"
   },
   "exclude": [
-    "node_modules"
+    "**/bin",
```
armanio123: I'm not sure if all the directories are relevant for this kind of project, but I think is still a good idea to include them as a general list of excluded. Let me know if you think this should be removed.

billti: I don't think we should change the tsconfig.json template. People don't keep TypeScript source files under those folders.

Nodejs/Product/Nodejs/Templates/Files/TypeScriptTsConfig/jsconfig.vstemplate
billti: Why don't you add them? The file is in this repo: https://github.com/Microsoft/nodejstools/blob/master/Nodejs/Product/Nodejs/VSPackage.resx#L171

Nodejs/Product/Nodejs/Templates/Files/TypeScriptTsConfig/jsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
```
billti: Why are you putting these files in the same folder with the tsconfig.json template files, when every other file gets its own folder? Separate into a Templates/Files/JsConfig folder.

armanio123: I thought of not cluttering the with a lot of folders as both of them are similar files. But I don't have a preference in one or another. Changed.

Nodejs/Product/Nodejs/Templates/Files/TypeScriptTsConfig/jsconfig.vstemplate
armanio123: My plan was to unify the text in one place for all the jsconfig templates. But since we have decided that for now only to do NodeJS, I have added the changes this repo. We can modify it later if necessary.

Nodejs/Product/Nodejs/Templates/Files/TypeScriptTsConfig/tsconfig.json
armanio123: Ok, rollback.

Nodejs/Product/Nodejs/Templates/Files/TypeScriptJsConfig/jsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "exclude": [
+    "**/bin",
+    "**/bower_components",
+    "**/jspm_packages",
+    "**/node_modules",
+    "**/obj",
+    "**/platforms",
+    "**/wwwroot/lib"
```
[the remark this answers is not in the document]
RyanCavanaugh: Correct; this section controls the initial list of input files and we will never consider them for any other behavior.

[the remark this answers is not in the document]
armanio123: Based on Ryan's input, any huge libraries should be added to the safelist instead of excluding the whole folder. I will remove this from the PR and analyze the libraries for the safelist.

PR #1851, opened 2018-02-05
Fix null reference in project properties

paulvanbrenk:
Make sure we don't error on non-existing user project file.

### DIFF
--- a/Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
@@ -1881,7 +1881,7 @@ namespace Microsoft.VisualStudioTools.Project
         {
             this.Site.GetUIThread().MustBeCalledFromUIThread();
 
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
```diff
@@ -1881,7 +1881,7 @@ public virtual void MovePropertyToProjectFile(string propertyName)
         {
             this.Site.GetUIThread().MustBeCalledFromUIThread();
 
-            var prop = this.userBuildProject.GetProperty(propertyName);
+            var prop = this.userBuildProject?.GetProperty(propertyName);
```
amcasey: Is the access on line 1900 safe?

paulvanbrenk: Yes, because of this call ` this.EnsureUserProjectFile();`

PR #1854, opened 2018-02-12
Harden the search for the installed TypeScript SDK

paulvanbrenk:
There's a bug in the setup engine, which doesn't always enumerate the
installed version of the TypeScript SDK.


minestarks:
OK but please don't merge this until the root cause bug is confirmed and there is a plan for fixing it. There might end up being a better workaround.

PR #1868, opened 2018-02-26
Fix null reference in build

paulvanbrenk:


Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
```diff
@@ -1001,11 +986,11 @@ protected override void Dispose(bool disposing)
                     SetBuildProject(null);
                 }
 
-                this.BuildLogger = null;
```
amcasey: >this.BuildLogger = null; [](start = 16, length = 24)

That's amazing

PR #1892, opened 2018-03-13
Fix test case discovery in Jasmine

paulvanbrenk:
Instead of depending on specFilter, which is broken in 3.0,
we now enumerate the topSuite property, and it's children
to discover the test cases in Jasmine unit tests.

### DIFF
--- a/Nodejs/Product/TestAdapterImpl/TestFrameworks/Jasmine/jasmine.js
+++ b/Nodejs/Product/TestAdapterImpl/TestFrameworks/Jasmine/jasmine.js
@@ -1,3 +1,4 @@
+// @ts-check
 "use strict";
 
 var EOL = require("os").EOL;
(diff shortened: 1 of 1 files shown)


Nodejs/Product/TestAdapterImpl/TestFrameworks/Jasmine/jasmine.js
```diff
@@ -80,43 +81,59 @@ function initializeJasmine(Jasmine, projectFolder) {
...
-    var suiteName = spec.getSpecName(spec);
-    if (suiteName && suiteName.length > 0) {
-        return suiteName.substr(0, suiteName.length - spec.description.length - 1);
-    }
-    return suiteName;
+/**
+ * @param {jasmine.Suite} suite
+ * @param {object[]} testList
+ * @param {string} testFile
+ */
```
amcasey: Would it be useful to doc the return type?

Nodejs/Product/TestAdapterImpl/TestFrameworks/Jasmine/jasmine.js
```diff
@@ -80,43 +81,59 @@ function initializeJasmine(Jasmine, projectFolder) {
...
+function enumerateSpecs(suite, testList, testFile) {
+    suite.children.forEach((child) => {
+        if (child instanceof jasmine.Suite) {
+            enumerateSpecs(child, testList, testFile);
+        } else {
+            testList.push({
+                test: child.getSpecName(child),
```
amcasey: Is there a difference between `suite.getSpecName(child)` and `child.getSpecName(child)`?

Nodejs/Product/TestAdapterImpl/TestFrameworks/Jasmine/jasmine.js
```diff
@@ -80,43 +81,59 @@ function initializeJasmine(Jasmine, projectFolder) {
...
+                suite: suite.description,
+                file: testFile,
+                line: 0,
```
amcasey: The line and column don't matter?

Nodejs/Product/TestAdapterImpl/TestFrameworks/Jasmine/jasmine.js
```diff
@@ -80,43 +81,59 @@ function initializeJasmine(Jasmine, projectFolder) {
...
+ * @param {string} discoverResultFile
+ * @param {string} projectFolder
+ */
 function find_tests(testFileList, discoverResultFile, projectFolder) {
     var Jasmine = detectJasmine(projectFolder);
     if (!Jasmine) {
         return;
     }
+    var jasmineInstance = initializeJasmine(Jasmine, projectFolder);
+    jasmineInstance.env.specFilter = (spec) => {
```
amcasey: Can this be simplified to `_ => false`?

Nodejs/Product/TestAdapterImpl/TestFrameworks/Jasmine/jasmine.js
paulvanbrenk: they do, but there is no way to get them from the Jasmine API.

Nodejs/Product/TestAdapterImpl/TestFrameworks/Jasmine/jasmine.js
paulvanbrenk: The compiler was smart enough to figure out that was `void`, so it didn't need help.

Nodejs/Product/TestAdapterImpl/TestFrameworks/Jasmine/jasmine.js
paulvanbrenk: No

PR #1915, opened 2018-04-10
Add inspect brk parameter

digeff:
Issue #
Debugging Node doesn't work on the latest VS version


paulvanbrenk:
replaced with #1917

PR #1933, opened 2018-04-17
Remove node crash

paulvanbrenk:
This should make the HierarchyId map more reliable

### DIFF
--- a/Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
@@ -16,7 +16,7 @@ namespace Microsoft.VisualStudioTools.Project
         /// </summary>
         public uint Add(HierarchyNode node)
         {
-            VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread();
+            Debug.Assert(VisualStudio.Shell.ThreadHelper.CheckAccess());
 
 #if DEBUG
             foreach (var reference in this.nodes.Values)
@@ -35,7 +35,9 @@ namespace Microsoft.VisualStudioTools.Project
             {
                 idx = this.nodes.Count;
             }
-            this.nodes[idx] = new WeakReference<HierarchyNode>(node);
+
+            var addSuccess = this.nodes.TryAdd(idx, new WeakReference<HierarchyNode>(node));
+            Debug.Assert(addSuccess, "Failed to add a new item");
 
             return (uint)idx + 1;
         }
@@ -45,22 +47,18 @@ namespace Microsoft.VisualStudioTools.Project
         /// </summary>
         public void Remove(HierarchyNode node)
         {
-            VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread();
+            Debug.Assert(VisualStudio.Shell.ThreadHelper.CheckAccess());
 
             Debug.Assert(node != null, "Called with null node");
 
             var idx = (int)node.ID - 1;
 
-            var removeCheck = this.nodes.TryGetValue(idx, out var weakRef);
+            var removeCheck = this.nodes.TryRemove(idx, out var weakRef);
 
             Debug.Assert(removeCheck, "How did we get an id, which we haven't seen before");
             Debug.Assert(weakRef != null, "Double delete is not expected.");
+            Debug.Assert(weakRef.TryGetTarget(out var found) && object.ReferenceEquals(node, found), "The node has the wrong id, or was GC-ed before.");
 
-            var checkReference = weakRef.TryGetTarget(out var found) && object.ReferenceEquals(node, found);
-
-            Debug.Assert(checkReference, "The node has the wrong id.");
-
-            this.nodes[idx] = null;
             this.freedIds.Push(idx);
         }
 
@@ -71,18 +69,19 @@ namespace Microsoft.VisualStudioTools.Project
         {
             get
             {
-                VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread();
+                Debug.Assert(VisualStudio.Shell.ThreadHelper.CheckAccess());
 
-                var i = (int)itemId - 1;
-                if (0 <= i && i < this.nodes.Count)
+                var idx = (int)itemId - 1;
+                if (0 <= idx && idx < this.nodes.Count)
                 {
-                    var reference = this.nodes[i];
-                    if (reference != null && reference.TryGetTarget(out var node))
+                    if (this.nodes.TryGetValue(idx, out var reference) && reference != null && reference.TryGetTarget(out var node))
                     {
                         Debug.Assert(node != null);
                         return node;
                     }
                 }
+
+                // This is a valid return value, this gets called by VS after we deleted the item
                 return null;
             }
         }
--- a/Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
@@ -20,8 +20,8 @@ namespace Microsoft.VisualStudioTools.Project
         /// </summary>
         public uint Add(HierarchyNode node)
         {
-            Debug.Assert(VisualStudio.Shell.ThreadHelper.CheckAccess());
-
+            VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread();
+            Debug.Assert(node != null, "The node added here should never be null.");
 #if DEBUG
             foreach (var reference in this.nodes.Values)
             {
@@ -33,7 +33,6 @@ namespace Microsoft.VisualStudioTools.Project
                     }
                 }
             }
-
 #endif
             if (!this.freedIds.TryPop(out var idx))
             {
@@ -48,16 +47,16 @@ namespace Microsoft.VisualStudioTools.Project
 
         private uint NextIndex()
         {
+            // +1 since 0 is not a valid HierarchyId
             return (uint)this.nodes.Count + 1;
         }
-
 
         /// <summary>
         /// Must be called from the UI thread
         /// </summary>
         public void Remove(HierarchyNode node)
         {
-            Debug.Assert(VisualStudio.Shell.ThreadHelper.CheckAccess());
+            VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread();
 
             Debug.Assert(node != null, "Called with null node");
 
@@ -66,7 +65,7 @@ namespace Microsoft.VisualStudioTools.Project
             var removeCheck = this.nodes.TryRemove(idx, out var weakRef);
 
             Debug.Assert(removeCheck, "How did we get an id, which we haven't seen before");
-            Debug.Assert(weakRef != null, "Double delete is not expected.");
+            Debug.Assert(weakRef != null, "How did we insert a null value.");
             Debug.Assert(weakRef.TryGetTarget(out var found) && object.ReferenceEquals(node, found), "The node has the wrong id, or was GC-ed before.");
 
             this.freedIds.Push(idx);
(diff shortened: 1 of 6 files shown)


Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
```diff
@@ -57,33 +47,21 @@ public void Remove(HierarchyNode node)
...
-                {
-                    throw new InvalidOperationException("Trying to retrieve a node before adding.");
-                }
+            var removeCheck = this.nodes.TryGetValue(idx, out var weakRef);
 
-                if (weakRef.TryGetTarget(out var found) && !object.ReferenceEquals(node, found))
-                {
-                    throw new InvalidOperationException("The node has the wrong id.");
-                }
+            Debug.Assert(removeCheck, "How did we get an id, which we haven't seen before");
```
amcasey: >removeCheck [](start = 25, length = 11)

Since `removeCheck == (weakRef != null)`, this also indicates a double delete, right?

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
```diff
@@ -57,33 +47,21 @@ public void Remove(HierarchyNode node)
...
+            Debug.Assert(weakRef != null, "Double delete is not expected.");
 
-                this.ids[i] = null;
-                this.freedIds.Push(i);
-            }
+            var checkReference = weakRef.TryGetTarget(out var found) && object.ReferenceEquals(node, found);
+
+            Debug.Assert(checkReference, "The node has the wrong id.");
```
amcasey: >The node has the wrong id. [](start = 42, length = 26)

Or has been GC'd?

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
```diff
@@ -1,25 +1,25 @@
...
-        private readonly object theLock = new object();
+        private readonly ConcurrentDictionary<int, WeakReference<HierarchyNode>> nodes = new ConcurrentDictionary<int, WeakReference<HierarchyNode>>();
+        private readonly ConcurrentStack<int> freedIds = new ConcurrentStack<int>();
 
         /// <summary>
         /// Must be called from the UI thread
         /// </summary>
         public uint Add(HierarchyNode node)
         {
+            VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread();
```
amcasey: >VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread(); [](start = 12, length = 55)

Do we depend on this or is it just true?  If the latter, consider asserting instead.

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
```diff
@@ -29,25 +29,15 @@ public uint Add(HierarchyNode node)
...
-                }
-                else
-                {
-                    this.ids.Add(new WeakReference<HierarchyNode>(node));
-                    // ids are 1 based
-                    return (uint)this.ids.Count;
-                }
+                idx = this.nodes.Count;
             }
+            this.nodes[idx] = new WeakReference<HierarchyNode>(node);
```
amcasey: >nodes[idx] [](start = 17, length = 10)

It seems like it might be safer to call `TryAdd`.

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
```diff
@@ -57,33 +47,21 @@ public void Remove(HierarchyNode node)
...
+            Debug.Assert(weakRef != null, "Double delete is not expected.");
 
-                this.ids[i] = null;
-                this.freedIds.Push(i);
-            }
+            var checkReference = weakRef.TryGetTarget(out var found) && object.ReferenceEquals(node, found);
+
+            Debug.Assert(checkReference, "The node has the wrong id.");
+
+            this.nodes[idx] = null;
```
amcasey: >nodes [](start = 17, length = 5)

It seems like it might be safer to call `TryRemove` (or `TryUpdate` if we want to keep the empty slot, for some reason).

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
```diff
@@ -93,12 +71,15 @@ public void Remove(HierarchyNode node)
         {
             get
             {
+                VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread();
```
amcasey: >VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread(); [](start = 16, length = 55)

This ended up being true?

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
```diff
@@ -93,12 +71,15 @@ public void Remove(HierarchyNode node)
...
+
                 var i = (int)itemId - 1;
-                if (0 <= i && i < this.ids.Count)
+                if (0 <= i && i < this.nodes.Count)
                 {
-                    var reference = this.ids[i];
+                    var reference = this.nodes[i];
```
amcasey: >nodes [](start = 41, length = 5)

`TryGetValue`?

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
paulvanbrenk: weakref may be null, because we don't know have that id in the dictionary, i.e. never added, or because we set it to null previously, i.e. double delete.

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
paulvanbrenk: yes

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
amcasey: Seems like we do depend on this.  If it's not true, then we can't just assert that `TryAdd` succeeds.

---
In reply to: [182269818](https://github.com/Microsoft/nodejstools/pull/1933#discussion_r182269818) [](ancestors = 182269818)

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
amcasey: Null no longer indicates a double-delete.

---
In reply to: [182285218](https://github.com/Microsoft/nodejstools/pull/1933#discussion_r182285218) [](ancestors = 182285218)

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
```diff
@@ -93,15 +69,19 @@ public void Remove(HierarchyNode node)
         {
             get
             {
-                var i = (int)itemId - 1;
-                if (0 <= i && i < this.ids.Count)
+                Debug.Assert(VisualStudio.Shell.ThreadHelper.CheckAccess());
+
+                var idx = (int)itemId - 1;
+                if (0 <= idx && idx < this.nodes.Count)
```
amcasey: >if (0 <= idx && idx < this.nodes.Count) [](start = 16, length = 39)

The bounds check is now redundant.

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
```diff
@@ -29,61 +29,37 @@ public uint Add(HierarchyNode node)
...
-                    // ids are 1 based
-                    return (uint)this.ids.Count;
-                }
+                idx = this.nodes.Count;
             }
+
+            var addSuccess = this.nodes.TryAdd(idx, new WeakReference<HierarchyNode>(node));
+            Debug.Assert(addSuccess, "Failed to add a new item");
+
+            return (uint)idx + 1;
```
amcasey: >+ 1 [](start = 29, length = 3)

Given that we're no longer indexing into a list, is there any value in maintaining off-by-one indices?

Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
```diff
@@ -590,7 +587,7 @@ public virtual string OutputBaseRelativePath
         /// <summary>
         /// Gets a collection of integer ids that maps to project item instances
         /// </summary>
-        internal HierarchyIdMap ItemIdMap => this.itemIdMap;
+        internal HierarchyIdMap ItemIdMap => HierarchyIdMap.Instance;
```
amcasey: >ItemIdMap [](start = 32, length = 9)

Seems like the callers could just use the constant directly, but no objection if you're minimizing the diff.

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
```diff
@@ -30,24 +34,21 @@ public uint Add(HierarchyNode node)
...
+                idx = this.NextIndex();
             }
+
+            var addSuccess = this.nodes.TryAdd(idx, new WeakReference<HierarchyNode>(node));
+            Debug.Assert(addSuccess, "Failed to add a new item");
+
+            return idx;
+        }
+
+        private uint NextIndex()
```
amcasey: >private uint NextIndex() [](start = 8, length = 24)

Since this is only correct if the free-list is empty, I would prefer to have it inlined.

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
```diff
@@ -1,25 +1,29 @@
...
         /// </summary>
         public uint Add(HierarchyNode node)
         {
+            VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread();
+            Debug.Assert(node != null, "The node added here should never be null.");
 #if DEBUG
-            foreach (var reference in this.ids)
+            foreach (var reference in this.nodes.Values)
             {
                 if (reference != null)
```
amcasey: >if (reference != null) [](start = 16, length = 22)

Is this always true?

amcasey: It's kind of weird to do this check by looping, given that they can only be equal if they have the same ID.  If we want to validate the entire structure, it would make more sense to verify that each non-GC'd value has an ID matching its key.

---
In reply to: [182613675](https://github.com/Microsoft/nodejstools/pull/1933#discussion_r182613675) [](ancestors = 182613675)

paulvanbrenk: Yes, the weakreference should never be null

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
```diff
@@ -93,15 +78,16 @@ public void Remove(HierarchyNode node)
         {
             get
             {
-                var i = (int)itemId - 1;
-                if (0 <= i && i < this.ids.Count)
+                VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread();
+
+                var idx = itemId;
```
amcasey: >idx [](start = 20, length = 3)

`idx` is pretty redundant. :)

Nodejs/Product/Nodejs/SharedProject/HierarchyIdMap.cs
```diff
@@ -93,15 +78,16 @@ public void Remove(HierarchyNode node)
...
+                if (this.nodes.TryGetValue(idx, out var reference) && reference != null && reference.TryGetTarget(out var node))
```
amcasey: >reference != null [](start = 70, length = 17)

No longer need a null check?

amcasey:
That's much cleaner.  Thanks!

paulvanbrenk:
@amcasey going to merge this now... I'll make the stylistic changes tomorrow, but this way we have a build to test.

PR #1944, opened 2018-04-25
Prevent crash when there is no pending breakpoint

paulvanbrenk:
fixes 

[Watson] clr20r3: CLR_EXCEPTION_System.Collections.Generic.KeyNotFoundException_80131577_UNKNOWN!Microsoft.NodejsTools.Debugger.DebugEngine.AD7Engine.OnBreakpointBound+1e

### DIFF
--- a/Nodejs/Product/Nodejs/Debugger/DebugEngine/AD7Engine.cs
+++ b/Nodejs/Product/Nodejs/Debugger/DebugEngine/AD7Engine.cs
@@ -1292,17 +1292,20 @@ namespace Microsoft.NodejsTools.Debugger.DebugEngine
         private void OnBreakpointBound(object sender, BreakpointBindingEventArgs e)
         {
             var pendingBreakpoint = this._breakpointManager.GetPendingBreakpoint(e.Breakpoint);
-            var breakpointBinding = e.BreakpointBinding;
-            var codeContext = new AD7MemoryAddress(this, pendingBreakpoint.DocumentName, breakpointBinding.Target.Line, breakpointBinding.Target.Column);
-            var documentContext = new AD7DocumentContext(codeContext);
-            var breakpointResolution = new AD7BreakpointResolution(this, breakpointBinding, documentContext);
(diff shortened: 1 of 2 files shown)


Nodejs/Product/Nodejs/Debugger/DebugEngine/AD7Engine.cs
```diff
@@ -1292,17 +1292,20 @@ private void OnBreakpointHit(object sender, BreakpointHitEventArgs e)
...
-            var documentContext = new AD7DocumentContext(codeContext);
-            var breakpointResolution = new AD7BreakpointResolution(this, breakpointBinding, documentContext);
-            var boundBreakpoint = new AD7BoundBreakpoint(breakpointBinding, pendingBreakpoint, breakpointResolution, breakpointBinding.Enabled);
-            this._breakpointManager.AddBoundBreakpoint(breakpointBinding, boundBreakpoint);
-            Send(
-                new AD7BreakpointBoundEvent(pendingBreakpoint, boundBreakpoint),
-                AD7BreakpointBoundEvent.IID,
-                null
-            );
+            if (pendingBreakpoint != null)
```
billti: Why did this regress? I don't understand why in the OnBreakpointBound handler GetPendingBreakpoint would return null. When does this occur? Did it not used to happen before?

paulvanbrenk: There seems to be a case that there is no pending breakpoint added to the breakpoint manager, before querying for that breakpoint. 
I have not been able to reproduce this on various versions of Node 6, this chance prevents from VS crashing, and I have requested additional information.

amcasey: It definitely seems like this indicates a bug in an underlying component.  The other callers I can find don't check for null.  Also, it seems strange that a pending breakpoint is required at all - all the information required for error reporting seems to be in the event args.

paulvanbrenk: As discussed. I'll add a comment to explain more.

PR #1956, opened 2018-05-11
Get Node.js and npm paths from VS package

minestarks:
Attempt to find bundled node.exe and npm.cmd from Visual Studio

@paulvanbrenk Please note I didn't test this.


paulvanbrenk:
thanks

minestarks:
@paulvanbrenk argh. The second argument to `RegistryRoot` should have been `false` (for `writable`)

paulvanbrenk:
Had to roll this back since it doesn't work for the npm tooling (no reference to VS Shell). Looks like more work, I'll do this next week.

PR #1958, opened 2018-05-14
Add manifest build version to make insertions into VS easier

paulvanbrenk:
This adds the following entry to the manifest, which is used for insertions into VS and should make merge conflict resolution easier.

image

### DIFF
--- a/Nodejs/Setup/NodejsTools.vsmanproj
+++ b/Nodejs/Setup/NodejsTools.vsmanproj
@@ -14,13 +14,13 @@
     <!-- Define properties that drive the manifest creation here. -->
     <FinalizeManifest>true</FinalizeManifest>
     <FinalizeSkipLayout>true</FinalizeSkipLayout>
-    <BuildNumber>$(VSTarget).$(BuildNumber)</BuildNumber>
     <TargetName>$(MSBuildProjectName)</TargetName>
   </PropertyGroup>
   <PropertyGroup>
     <!-- NTVS uses BuildVersion as 'major.minor.build.revision' and BuildNumber is 'build.revision'.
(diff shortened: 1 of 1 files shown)


Nodejs/Setup/NodejsTools.vsmanproj
```diff
@@ -14,13 +14,13 @@
     <!-- Define properties that drive the manifest creation here. -->
     <FinalizeManifest>true</FinalizeManifest>
     <FinalizeSkipLayout>true</FinalizeSkipLayout>
-    <BuildNumber>$(VSTarget).$(BuildNumber)</BuildNumber>
```
paulvanbrenk: removed since it is overridden on line 23

Nodejs/Setup/NodejsTools.vsmanproj
```diff
@@ -14,13 +14,13 @@
...
     <TargetName>$(MSBuildProjectName)</TargetName>
   </PropertyGroup>
   <PropertyGroup>
     <!-- NTVS uses BuildVersion as 'major.minor.build.revision' and BuildNumber is 'build.revision'.
              But VS Manifest task wants BuildNumber to be 'major.minor.build.revision' -->
     <BuildNumber>$(BuildVersion)</BuildNumber>
+    <ManifestBuildVersion>$(BuildVersion)</ManifestBuildVersion>
```
minestarks: Where is BuildVersion set?
Where is BuildNumber used?

paulvanbrenk: `BuildVersion` is set in `fileVersion.targers` which is imported through `SetupProjectBefore.settings`. Looks like `BuildNumber` is used as part of the id field in the vsman json.

PR #1959, opened 2018-05-16
Open Folder Unit Test

paulvanbrenk:
Adds support for running unit tests in "Open Folder" scenarios.

TODO: debugging and Telemetry, but the current changes implement discovery and running tests.

### DIFF
--- a/Nodejs/Product/Nodejs/Workspace/PackageJsonScannerFactory.cs
+++ b/Nodejs/Product/Nodejs/Workspace/PackageJsonScannerFactory.cs
@@ -87,6 +87,13 @@ namespace Microsoft.NodejsTools.Workspace
                     // (See Microsoft.VisualStudio.Workspace.VSIntegration.UI.FileContextActionsCommandHandlersProvider.Provider.GetActionProviderForProjectConfiguration)
                     fileDataValues.Add(new FileDataValue(DebugLaunchActionContext.ContextTypeGuid, main, null, target: null));
                 }
+
+                var testRoot = packageJson.TestRoot;
+                if (!string.IsNullOrEmpty(testRoot))
+                {
+                    fileDataValues.Add(new FileDataValue(NodejsConstants.TestRootDataValueGuid, "TestRoot", testRoot));
+                }
+
                 return Task.FromResult(fileDataValues);
             }
         }
--- a/Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainer.cs
+++ b/Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainer.cs
@@ -1,0 +1,78 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Microsoft.VisualStudio.TestWindow.Extensibility;
+using Microsoft.VisualStudio.TestWindow.Extensibility.Model;
+
+namespace Microsoft.NodejsTools.Workspace
+{
+    public sealed class PackageJsonTestContainer : ITestContainer
+    {
+        private int version;
+
+        public PackageJsonTestContainer(ITestContainerDiscoverer discoverer, string source, string testRoot)
+        {
+            this.Discoverer = discoverer;
+            this.Source = source;
+            this.TestRoot = testRoot;
+        }
+
+        private PackageJsonTestContainer(PackageJsonTestContainer other) :
+            this(other.Discoverer, other.Source, other.TestRoot)
+        {
+            this.version = other.version;
+        }
+
+        public ITestContainerDiscoverer Discoverer { get; }
+
+        public string Source { get; }
+
+        public string TestRoot { get; }
+
+        public IEnumerable<Guid> DebugEngines { get { yield break; } }
+
+        public FrameworkVersion TargetFramework { get; } = FrameworkVersion.None;
+
+        public Architecture TargetPlatform { get; } = Architecture.X86;
+
+        public bool IsAppContainerTestContainer { get; } = false;
+
+        public int CompareTo(ITestContainer other)
+        {
+            if (other == null || !(other is PackageJsonTestContainer testContainer))
+            {
+                return -1;
+            }
+
+            if (this.version != testContainer.version)
+            {
+                return this.version - testContainer.version;
+            }
+
+            var sourceCompare = StringComparer.OrdinalIgnoreCase.Compare(this.Source, testContainer.Source);
+
+            return sourceCompare == 0 ? sourceCompare : StringComparer.OrdinalIgnoreCase.Compare(this.TestRoot, testContainer.TestRoot);
+        }
+
+        public IDeploymentData DeployAppContainer() => null;
+
+        public ITestContainer Snapshot() => new PackageJsonTestContainer(this);
+
+        public bool IsContained(string javaScriptFilePath)
+        {
+            Debug.Assert(Path.IsPathRooted(javaScriptFilePath) && !string.IsNullOrEmpty(javaScriptFilePath), "Expected a rooted path.");
+
+            return javaScriptFilePath.StartsWith(this.TestRoot, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public void IncreaseVersion()
+        {
+            Interlocked.Increment(ref this.version);
+        }
+    }
+}
--- a/Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
+++ b/Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
@@ -1,0 +1,147 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestWindow.Extensibility;
+using Microsoft.VisualStudio.Workspace;
+using Microsoft.VisualStudio.Workspace.Indexing;
+using Microsoft.VisualStudio.Workspace.VSIntegration.Contracts;
+
+namespace Microsoft.NodejsTools.Workspace
+{
+    [Export(typeof(ITestContainerDiscoverer))]
+    public sealed class PackageJsonTestContainerDiscoverer : ITestContainerDiscoverer
+    {
+        private readonly IVsFolderWorkspaceService workspaceService;
+
+        private readonly List<PackageJsonTestContainer> containers = new List<PackageJsonTestContainer>();
+        private readonly object containerLock = new object();
+
+        private IWorkspace currentWorkspace;
+
+        [ImportingConstructor]
+        public PackageJsonTestContainerDiscoverer(IVsFolderWorkspaceService workspaceService)
+        {
+            this.workspaceService = workspaceService;
+            this.workspaceService.OnActiveWorkspaceChanged += this.OnActiveWorkspaceChangedAsync;
+
+            if (this.workspaceService.CurrentWorkspace != null)
+            {
+                this.currentWorkspace = this.workspaceService.CurrentWorkspace;
+
+                var fileWatcherService = this.currentWorkspace.GetFileWatcherService();
+                fileWatcherService.OnFileSystemChanged += this.FileSystemChangedAsync;
+
+                var indexService = this.currentWorkspace.GetIndexWorkspaceService();
+                indexService.OnFileScannerCompleted += this.FileScannerCompletedAsync;
+
+                this.currentWorkspace.JTF.RunAsync(async () =>
+                {
+                    // Yield so we don't do this now. Don't want to block the constructor.
+                    await Task.Yield();
+
+                    // See if we have an update
+                    await AttemptUpdateAsync();
+                });
+            }
+        }
+
+        public Uri ExecutorUri { get; } = NodejsConstants.PackageJsonExecutorUri;
+
+        public IEnumerable<ITestContainer> TestContainers
+        {
+            get
+            {
+                lock (this.containerLock)
+                {
+                    return this.containers.ToArray();
+                }
+            }
+        }
+
+        public event EventHandler TestContainersUpdated;
+
+        private async Task AttemptUpdateAsync()
+        {
+            if (this.currentWorkspace != null)
+            {
+                var indexServce = this.currentWorkspace.GetIndexWorkspaceService();
+                var filesDataValues = await indexServce.GetFilesDataValuesAsync<string>(NodejsConstants.TestRootDataValueGuid);
+
+                lock (this.containerLock)
+                {
+                    this.containers.Clear();
+                    foreach (var dataValue in filesDataValues)
+                    {
+                        var rootFilePath = this.currentWorkspace.MakeRooted(dataValue.Key);
+                        var testRoot = dataValue.Value.Where(f => f.Name == "TestRoot").Select(f => f.Value).FirstOrDefault();
+
+                        if (!string.IsNullOrEmpty(testRoot))
+                        {
+                            var testRootPath = this.currentWorkspace.MakeRooted(testRoot);
+                            this.containers.Add(new PackageJsonTestContainer(this, rootFilePath, testRootPath));
+                        }
+                    }
+                }
+            }
+            this.TestContainersUpdated?.Invoke(this, EventArgs.Empty);
+        }
+
+        private async Task OnActiveWorkspaceChangedAsync(object sender, EventArgs e)
+        {
+            this.currentWorkspace = this.workspaceService.CurrentWorkspace;
+
+            await AttemptUpdateAsync();
+        }
+
+        private Task FileSystemChangedAsync(object sender, FileSystemEventArgs args)
+        {
+            // We only need to raise the containers updated event in case a js file contained in the
+            // test root is updated.
+            // Any changes to the 'package.json' will be handled by the FileScannerCompleted event.
+            if (IsJavaScriptFile(args.FullPath))
+            {
+                // use a flag so we don't deadlock
+                var testsUpdated = false;
+                lock (this.containerLock)
+                {
+                    foreach (var container in this.containers)
+                    {
+                        if (container.IsContained(args.FullPath))
+                        {
+                            container.IncreaseVersion();
+                            testsUpdated = true;
+                            break;
+                        }
+                    }
+                }
+                if (testsUpdated)
+                {
+                    this.TestContainersUpdated?.Invoke(this, EventArgs.Empty);
+                }
+            }
+
+            return Task.CompletedTask;
+        }
+
+        private async Task FileScannerCompletedAsync(object sender, FileScannerEventArgs args)
+        {
+            await AttemptUpdateAsync();
+        }
+
+        private static bool IsJavaScriptFile(string path)
+        {
+            var ext = Path.GetExtension(path);
+            if (StringComparer.OrdinalIgnoreCase.Equals(ext, ".js") || StringComparer.OrdinalIgnoreCase.Equals(ext, ".jsx"))
+            {
+                return true;
+            }
+
+            return false;
+        }
+    }
+}
--- a/Nodejs/Product/TestAdapter/JavaScriptTestExecutor.cs
+++ b/Nodejs/Product/TestAdapter/JavaScriptTestExecutor.cs
@@ -198,7 +198,7 @@ namespace Microsoft.NodejsTools.TestAdapter
             // May be null, but this is handled by RunTestCase if it matters.
             // No VS instance just means no debugging, but everything else is
             // okay.
-            if (tests.Count() == 0)
+            if (!tests.Any())
             {
                 return;
             }
@@ -454,7 +454,7 @@ namespace Microsoft.NodejsTools.TestAdapter
             this.currentTests.Remove(test);
         }
 
-        private sealed class TestExecutionRedirector : Redirector
+        internal sealed class TestExecutionRedirector : Redirector
         {
             private readonly Action<string> writer;
 
@@ -470,7 +470,7 @@ namespace Microsoft.NodejsTools.TestAdapter
             public override bool CloseStandardInput() => false;
         }
 
-        private sealed class TestReceiver : ITestCaseDiscoverySink
+        internal sealed class TestReceiver : ITestCaseDiscoverySink
         {
             public readonly List<TestCase> Tests = new List<TestCase>();
 
--- a/Nodejs/Product/TestAdapter/PackageJsonTestDiscoverer.cs
+++ b/Nodejs/Product/TestAdapter/PackageJsonTestDiscoverer.cs
@@ -1,0 +1,97 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.NodejsTools.Npm;
+using Microsoft.NodejsTools.TestAdapter.TestFrameworks;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
+
+namespace Microsoft.NodejsTools.TestAdapter
+{
+    [FileExtension(".json")]
+    [DefaultExecutorUri(NodejsConstants.PackageJsonExecutorUriString)]
+    public sealed class PackageJsonTestDiscoverer : ITestDiscoverer
+    {
+        public void DiscoverTests(IEnumerable<string> sources, IDiscoveryContext discoveryContext, IMessageLogger logger, ITestCaseDiscoverySink discoverySink)
+        {
+            foreach (var source in sources)
+            {
+                // we're only interested in package.json files here.
+                if (PackageJsonFactory.IsPackageJsonFile(source))
+                {
+                    this.DiscoverTestFiles(source, logger, discoverySink);
+                }
+            }
+        }
+
+        private void DiscoverTestFiles(string packageJsonPath, IMessageLogger logger, ITestCaseDiscoverySink discoverySink)
+        {
+            logger.SendMessage(TestMessageLevel.Informational, $"Parsing '{packageJsonPath}'.");
+
+            var packageJson = PackageJsonFactory.Create(packageJsonPath);
+            if (string.IsNullOrEmpty(packageJson.TestRoot))
+            {
+                logger.SendMessage(TestMessageLevel.Informational, "No vsTestOptions|testRoot specified.");
+                return;
+            }
+
+            var workingDir = Path.GetDirectoryName(packageJsonPath);
+            var testFolderPath = Path.Combine(workingDir, packageJson.TestRoot);
+            TestFramework testFx = null;
+
+            foreach (var dep in packageJson.AllDependencies)
+            {
+                testFx = FrameworkDiscover.Intance.Get(dep.Name);
+                if (testFx != null)
+                {
+                    break;
+                }
+            }
+            testFx = testFx ?? FrameworkDiscover.Intance.Get("ExportRunner");
+
+            var nodeExePath = Nodejs.GetPathToNodeExecutableFromEnvironment();
+
+            if (!File.Exists(nodeExePath))
+            {
+                logger.SendMessage(TestMessageLevel.Error, "Node.exe was not found. Please install Node.js before running tests.");
+                return;
+            }
+
+            var fileList = Directory.GetFiles(testFolderPath, "*.js", SearchOption.AllDirectories);
+            var files = string.Join(";", fileList);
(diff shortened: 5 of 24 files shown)


Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
```diff
@@ -0,0 +1,147 @@
...
+                        }
+                    }
+                }
+            }
+            this.TestContainersUpdated?.Invoke(this, EventArgs.Empty);
+        }
+
+        private async Task OnActiveWorkspaceChangedAsync(object sender, EventArgs e)
+        {
+            this.currentWorkspace = this.workspaceService.CurrentWorkspace;
```
vukelich: You're not unsubscribing from the previous workspace's event handlers. Maybe that causes leaks, idk for certain. You will also want to subscribe to the new workspace's file watcher and index service. Each IWorkspace spins up a new instance of these services.

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
```diff
@@ -0,0 +1,147 @@
...
+
+                        if (!string.IsNullOrEmpty(testRoot))
+                        {
+                            var testRootPath = this.currentWorkspace.MakeRooted(testRoot);
+                            this.containers.Add(new PackageJsonTestContainer(this, rootFilePath, testRootPath));
+                        }
+                    }
+                }
+            }
+            this.TestContainersUpdated?.Invoke(this, EventArgs.Empty);
```
vukelich: This is being invoked outside of the `lock (this.containerLock){ }`. Is that intentional? It's in the lock for `FileSystemChangedAsync`

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        private async Task FileScannerCompletedAsync(object sender, FileScannerEventArgs args)
+        {
+            await AttemptUpdateAsync();
+        }
+
+        private static bool IsJavaScriptFile(string path)
+        {
+            var ext = Path.GetExtension(path);
```
vukelich: You're not checking if this is a directory.

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
paulvanbrenk: looks like it's outside the lock in both cases.

Nodejs/Product/Nodejs/Workspace/PackageJsonScannerFactory.cs
```diff
@@ -87,6 +87,13 @@ protected override Task<List<FileDataValue>> ComputeFileDataValuesAsync(string f
                     // (See Microsoft.VisualStudio.Workspace.VSIntegration.UI.FileContextActionsCommandHandlersProvider.Provider.GetActionProviderForProjectConfiguration)
                     fileDataValues.Add(new FileDataValue(DebugLaunchActionContext.ContextTypeGuid, main, null, target: null));
                 }
+
+                var testRoot = packageJson.TestRoot;
+                if (!string.IsNullOrEmpty(testRoot))
+                {
+                    fileDataValues.Add(new FileDataValue(NodejsConstants.TestRootDataValueGuid, "TestRoot", testRoot));
```
amcasey: >"TestRoot" [](start = 96, length = 10)

Question, not suggestion: should this be a constant?

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainer.cs
```diff
@@ -0,0 +1,78 @@
...
+            this.version = other.version;
+        }
+
+        public ITestContainerDiscoverer Discoverer { get; }
+
+        public string Source { get; }
+
+        public string TestRoot { get; }
+
+        public IEnumerable<Guid> DebugEngines { get { yield break; } }
```
amcasey: >yield break; [](start = 54, length = 12)

Unless there's an optimization I don't know about, this is way more expensive than `Array.Empty`.

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainer.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        public FrameworkVersion TargetFramework { get; } = FrameworkVersion.None;
+
+        public Architecture TargetPlatform { get; } = Architecture.X86;
+
+        public bool IsAppContainerTestContainer { get; } = false;
+
+        public int CompareTo(ITestContainer other)
+        {
+            if (other == null || !(other is PackageJsonTestContainer testContainer))
```
amcasey: >is [](start = 41, length = 2)

`is` includes a null check.

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainer.cs
```diff
@@ -0,0 +1,78 @@
...
+            {
+                return -1;
```
amcasey: >-1 [](start = 23, length = 2)

This doesn't actually happen, does it?  If it does, it seems unlikely that the comparison produces a consistent result when the operands are reversed.

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainer.cs
```diff
@@ -0,0 +1,78 @@
...
+            }
+
+            if (this.version != testContainer.version)
+            {
+                return this.version - testContainer.version;
+            }
+
+            var sourceCompare = StringComparer.OrdinalIgnoreCase.Compare(this.Source, testContainer.Source);
+
+            return sourceCompare == 0 ? sourceCompare : StringComparer.OrdinalIgnoreCase.Compare(this.TestRoot, testContainer.TestRoot);
```
amcasey: >== [](start = 33, length = 2)

`!= 0`?

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
```diff
@@ -0,0 +1,147 @@
...
+
+namespace Microsoft.NodejsTools.Workspace
+{
+    [Export(typeof(ITestContainerDiscoverer))]
+    public sealed class PackageJsonTestContainerDiscoverer : ITestContainerDiscoverer
+    {
+        private readonly IVsFolderWorkspaceService workspaceService;
+
+        private readonly List<PackageJsonTestContainer> containers = new List<PackageJsonTestContainer>();
+        private readonly object containerLock = new object();
```
amcasey: >containerLock [](start = 32, length = 13)

I'm probably forgetting some framework guideline, but why not lock on `containers`?

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
```diff
@@ -0,0 +1,147 @@
...
+                }
+            }
+            this.TestContainersUpdated?.Invoke(this, EventArgs.Empty);
+        }
+
+        private async Task OnActiveWorkspaceChangedAsync(object sender, EventArgs e)
+        {
+            this.currentWorkspace = this.workspaceService.CurrentWorkspace;
+
+            await AttemptUpdateAsync();
```
amcasey: >await AttemptUpdateAsync(); [](start = 12, length = 27)

This seems to do a lot less work than the ctor.  Maybe they should share a helper?

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
```diff
@@ -0,0 +1,147 @@
...
+                    // Yield so we don't do this now. Don't want to block the constructor.
+                    await Task.Yield();
+
+                    // See if we have an update
+                    await AttemptUpdateAsync();
+                });
+            }
+        }
+
+        public Uri ExecutorUri { get; } = NodejsConstants.PackageJsonExecutorUri;
```
amcasey: >{ get; } = [](start = 31, length = 10)

`=>`?

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+
+        public event EventHandler TestContainersUpdated;
+
+        private async Task AttemptUpdateAsync()
+        {
+            if (this.currentWorkspace != null)
+            {
+                var indexServce = this.currentWorkspace.GetIndexWorkspaceService();
```
amcasey: Typo

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            return Task.CompletedTask;
+        }
+
+        private async Task FileScannerCompletedAsync(object sender, FileScannerEventArgs args)
+        {
+            await AttemptUpdateAsync();
+        }
+
+        private static bool IsJavaScriptFile(string path)
```
amcasey: >IsJavaScriptFile [](start = 28, length = 16)

I'm skeptical that there isn't a helper for this or, at least, constants for known file extensions.

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
```diff
@@ -0,0 +1,147 @@
...
+{
+    [Export(typeof(ITestContainerDiscoverer))]
+    public sealed class PackageJsonTestContainerDiscoverer : ITestContainerDiscoverer
+    {
+        private readonly IVsFolderWorkspaceService workspaceService;
+
+        private readonly List<PackageJsonTestContainer> containers = new List<PackageJsonTestContainer>();
+        private readonly object containerLock = new object();
+
+        private IWorkspace currentWorkspace;
```
amcasey: >currentWorkspace [](start = 27, length = 16)

If `containers` is guarded by a lock, should this be as well?

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
```diff
@@ -0,0 +1,147 @@
...
+                var indexServce = this.currentWorkspace.GetIndexWorkspaceService();
+                var filesDataValues = await indexServce.GetFilesDataValuesAsync<string>(NodejsConstants.TestRootDataValueGuid);
+
+                lock (this.containerLock)
+                {
+                    this.containers.Clear();
+                    foreach (var dataValue in filesDataValues)
+                    {
+                        var rootFilePath = this.currentWorkspace.MakeRooted(dataValue.Key);
+                        var testRoot = dataValue.Value.Where(f => f.Name == "TestRoot").Select(f => f.Value).FirstOrDefault();
```
amcasey: >"TestRoot" [](start = 76, length = 10)

Constant for `"TestRoot"`?

amcasey: >.Select(f => f.Value).FirstOrDefault() [](start = 87, length = 38)

Alternatively, `.FirstOrDefault()?.Value`.

Nodejs/Product/TestAdapter/PackageJsonTestDiscoverer.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.IO;
+using Microsoft.NodejsTools.Npm;
+using Microsoft.NodejsTools.TestAdapter.TestFrameworks;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
+
+namespace Microsoft.NodejsTools.TestAdapter
+{
+    [FileExtension(".json")]
```
amcasey: >".json" [](start = 19, length = 7)

Constant?

Nodejs/Product/TestAdapter/PackageJsonTestDiscoverer.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            foreach (var dep in packageJson.AllDependencies)
+            {
+                testFx = FrameworkDiscover.Intance.Get(dep.Name);
+                if (testFx != null)
+                {
+                    break;
+                }
+            }
+            testFx = testFx ?? FrameworkDiscover.Intance.Get("ExportRunner");
```
amcasey: >Intance [](start = 49, length = 7)

Typo: "Instance"

amcasey: >FrameworkDiscover.Intance.Get("ExportRunner") [](start = 31, length = 45)

Is there some reason to believe this isn't null?

Nodejs/Product/TestAdapter/PackageJsonTestDiscoverer.cs
```diff
@@ -0,0 +1,97 @@
...
+            var nodeExePath = Nodejs.GetPathToNodeExecutableFromEnvironment();
+
+            if (!File.Exists(nodeExePath))
+            {
+                logger.SendMessage(TestMessageLevel.Error, "Node.exe was not found. Please install Node.js before running tests.");
+                return;
+            }
+
+            var fileList = Directory.GetFiles(testFolderPath, "*.js", SearchOption.AllDirectories);
+            var files = string.Join(";", fileList);
```
amcasey: >var files = string.Join(";", fileList); [](start = 12, length = 39)

This seems likely to be expensive.  Consider checking the logging level before evaluating it.

Nodejs/Product/TestAdapter/PackageJsonTestDiscoverer.cs
```diff
@@ -0,0 +1,97 @@
...
+                    DisplayName = discoveredTest.TestName
+                };
+
+                testcase.SetPropertyValue(JavaScriptTestCaseProperties.TestFramework, testFx.Name);
+                testcase.SetPropertyValue(JavaScriptTestCaseProperties.WorkingDir, workingDir);
+                testcase.SetPropertyValue(JavaScriptTestCaseProperties.NodeExePath, nodeExePath);
+
+                discoverySink.SendTestCase(testcase);
+            }
+            logger.SendMessage(TestMessageLevel.Informational, "Processing finished for framework 'mocha'.");
```
amcasey: >mocha [](start = 99, length = 5)

What's the connection to mocha?

Nodejs/Product/TestAdapter/PackageJsonTestExecutor.cs
```diff
@@ -0,0 +1,187 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.NodejsTools.TestAdapter
+{
+    [ExtensionUri(NodejsConstants.PackageJsonExecutorUriString)]
+    public sealed class PackageJsonTestExecutor : ITestExecutor
+    {
+        private List<TestCase> currentTests;
+        private IFrameworkHandle frameworkHandle;
+        private TestResult currentResult = null;
```
amcasey: > = null [](start = 40, length = 7)

How come some of the fields are initialized with their default values?

Nodejs/Product/TestAdapter/PackageJsonTestExecutor.cs
```diff
@@ -0,0 +1,187 @@
...
+        public void Cancel()
+        {
+            // We don't support cancellation
+        }
+
+        public void RunTests(IEnumerable<string> sources, IRunContext runContext, IFrameworkHandle frameworkHandle)
+        {
+            var receiver = new JavaScriptTestExecutor.TestReceiver();
+            var discoverer = new PackageJsonTestDiscoverer();
+            discoverer.DiscoverTests(sources, null, frameworkHandle, receiver);
```
amcasey: >null [](start = 46, length = 4)

A named argument might help.

Nodejs/Product/TestAdapter/PackageJsonTestExecutor.cs
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        private void ProcessTestRunnerEmit(string line)
+        {
+            try
+            {
+                var testEvent = JsonConvert.DeserializeObject<TestEvent>(line);
+                // Extract test from list of tests
+                var tests = this.currentTests.Where(n => n.DisplayName == testEvent.title);
+                if (tests.Count() > 0)
```
amcasey: >tests [](start = 20, length = 5)

We expect either 0 or 1 elements?

Nodejs/Product/TestAdapter/PackageJsonTestExecutor.cs
```diff
@@ -0,0 +1,187 @@
...
+            else
+            {
+                result.EndTime = DateTimeOffset.Now;
+                result.Duration = result.EndTime - result.StartTime;
+                result.Outcome = resultObject.passed ? TestOutcome.Passed : TestOutcome.Failed;
+            }
+
+            result.Messages.Add(new TestResultMessage(TestResultMessage.StandardOutCategory, string.Join(Environment.NewLine, standardOutputLines)));
+            result.Messages.Add(new TestResultMessage(TestResultMessage.StandardErrorCategory, string.Join(Environment.NewLine, standardErrorLines)));
+            result.Messages.Add(new TestResultMessage(TestResultMessage.AdditionalInfoCategory, string.Join(Environment.NewLine, standardErrorLines)));
```
amcasey: >standardErrorLines [](start = 129, length = 18)

We're printing this to two different categories?  If that's correct, consider extracting a local for the joined text.

Nodejs/Product/TestAdapter/PackageJsonTestExecutor.cs
```diff
@@ -0,0 +1,187 @@
...
+            frameworkHandle.RecordResult(result);
+            frameworkHandle.RecordEnd(test, result.Outcome);
+            this.currentTests.Remove(test);
```
amcasey: >Remove [](start = 30, length = 6)

It seems strange to remove from a random order list.  Is this the correct collection type?

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
paulvanbrenk: In this case locking on container would be fine, since it's not a public property. But I prefer to use the well-known pattern to avoid confusion in the future.

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
paulvanbrenk: I wanted to prevent the call to `ToArray` in the `TestContainers` property to now fail, because we would be changing the collection while converting it to an array.

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
paulvanbrenk: fixed

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
paulvanbrenk: changed

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
paulvanbrenk: fixed

Nodejs/Product/TestAdapter/PackageJsonTestDiscoverer.cs
paulvanbrenk: fixed

paulvanbrenk: Yes, this is installed by us and doesn't require additional `npm` packages.

Nodejs/Product/TestAdapter/PackageJsonTestDiscoverer.cs
paulvanbrenk: All log levels are always outputted, they only change the color of the output.

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
amcasey: I understood that pattern was for value types or public members of reference types, but I don't feel strongly about it.

---
In reply to: [192169424](https://github.com/Microsoft/nodejstools/pull/1959#discussion_r192169424) [](ancestors = 192169424)

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
amcasey: There really isn't a helper for detecting JS files?

---
In reply to: [188796597](https://github.com/Microsoft/nodejstools/pull/1959#discussion_r188796597) [](ancestors = 188796597)

paulvanbrenk: Nope. 

I'll refactor that in another PR

Nodejs/Product/TestAdapter/TestReceiver.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using System.Collections.Generic;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
+
+namespace Microsoft.NodejsTools.TestAdapter
+{
+    internal sealed class TestReceiver : ITestCaseDiscoverySink
+    {
+        public readonly List<TestCase> Tests = new List<TestCase>();
```
amcasey: Arguably, this should be exposed as an `IReadOnlyList`, but it's internal so I don't care very much.

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
```diff
@@ -0,0 +1,390 @@
...
+        private static string GetDebugArgs(Version nodeVersion, out int port)
+        {
+            port = GetFreePort();
+
+            return nodeVersion >= Node8Version ? $"--inspect-brk={port}" : $"--debug-brk={port}";
+        }
+
+        private void KillNodeProcess()
+        {
+            this.nodeProcess?.Kill();
```
amcasey: >this.nodeProcess?.Kill(); [](start = 12, length = 25)

Does this wait for it to exit?  Should it?

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
```diff
@@ -0,0 +1,390 @@
...
+namespace Microsoft.NodejsTools.TestAdapter
+{
+    internal sealed class TestExecutorWorker
+    {
+        private static readonly Version Node8Version = new Version(8, 0);
+
+        //get from NodeRemoteDebugPortSupplier::PortSupplierId
+        private static readonly Guid NodejsRemoteDebugPortSupplierUnsecuredId = new Guid("{9E16F805-5EFC-4CE5-8B67-9AE9B643EF80}");
+
+        private readonly ManualResetEvent cancelRequested = new ManualResetEvent(false);
```
amcasey: >cancelRequested [](start = 42, length = 15)

I'm probably missing something, but this doesn't appear to be used for blocking.  If it's just a flag, consider using a simpler type.

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
```diff
@@ -0,0 +1,390 @@
...
+            this.cancelRequested.Set();
+        }
+
+        /// <summary>
+        /// This is the equivalent of "RunAll" functionality
+        /// </summary>
+        /// <param name="sources">Refers to the list of test sources passed to the test adapter from the client.  (Client could be VS or command line)</param>
+        /// <param name="runContext">Defines the settings related to the current run</param>
+        /// <param name="frameworkHandle">Handle to framework.  Used for recording results</param>
+        public void RunTests(IEnumerable<string> sources, ITestDiscoverer discoverer)
```
amcasey: Is there some reason to believe a run isn't already in progress?

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
```diff
@@ -0,0 +1,390 @@
...
+            {
+                if (!fileToTests.ContainsKey(test.CodeFilePath))
+                {
+                    fileToTests[test.CodeFilePath] = new List<TestCase>();
+                }
+                fileToTests[test.CodeFilePath].Add(test);
+            }
+
+            // where key is the file and value is a list of tests
+            foreach (var entry in fileToTests)
```
amcasey: If the key isn't used, consider iterating over `fileToTests.Values`.

Nodejs/Product/TestAdapter/JavaScriptTestExecutor.cs
```diff
@@ -1,483 +1,33 @@
...
-
-        private ProcessOutput nodeProcess;
-        private readonly object syncObject = new object();
-        private List<TestCase> currentTests;
-        private IFrameworkHandle frameworkHandle;
-        private TestResult currentResult = null;
-        private ResultObject currentResultObject = null;
+        private TestExecutorWorker worker;
 
         public void RunTests(IEnumerable<TestCase> tests, IRunContext runContext, IFrameworkHandle frameworkHandle)
```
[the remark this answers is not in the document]
paulvanbrenk: Done

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
paulvanbrenk: It doesn't, and it shouldn't since we only call this for cancel.. I'll move it to the cancel method

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
paulvanbrenk: Added calls to cancel

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
paulvanbrenk: done

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
paulvanbrenk: Sure, but don't really want to touch this at this point.

Nodejs/Product/TestAdapter/TestReceiver.cs
paulvanbrenk: Made class private

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
```diff
@@ -0,0 +1,392 @@
...
+        private void ProcessTestRunnerEmit(string line)
+        {
+            try
+            {
+                var testEvent = JsonConvert.DeserializeObject<TestEvent>(line);
+                // Extract test from list of tests
+                var tests = this.currentTests.Where(n => n.DisplayName == testEvent.title);
+                if (tests.Any())
+                {
+                    switch (testEvent.type)
```
amcasey: >type [](start = 38, length = 4)

Just to confirm, we're ignoring other event types?

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
```diff
@@ -0,0 +1,392 @@
...
+                else
+                {
+                    this.currentResult.EndTime = DateTimeOffset.Now;
+                    this.currentResult.Duration = this.currentResult.EndTime - this.currentResult.StartTime;
+                    this.currentResult.Outcome = resultObject.passed ? TestOutcome.Passed : TestOutcome.Failed;
+                }
+
+                this.currentResult.Messages.Add(new TestResultMessage(TestResultMessage.StandardOutCategory, string.Join(Environment.NewLine, standardOutputLines)));
+                this.currentResult.Messages.Add(new TestResultMessage(TestResultMessage.StandardErrorCategory, string.Join(Environment.NewLine, standardErrorLines)));
+                this.currentResult.Messages.Add(new TestResultMessage(TestResultMessage.AdditionalInfoCategory, string.Join(Environment.NewLine, standardErrorLines)));
```
amcasey: If writing the same thing to two different streams is intentional, consider extracting a local.

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
paulvanbrenk: yes

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
```diff
@@ -0,0 +1,177 @@
...
+{
+    [Export(typeof(ITestContainerDiscoverer))]
+    public sealed class PackageJsonTestContainerDiscoverer : ITestContainerDiscoverer
+    {
+        private readonly IVsFolderWorkspaceService workspaceService;
+
+        private readonly List<PackageJsonTestContainer> containers = new List<PackageJsonTestContainer>();
+        private readonly object containerLock = new object();
+
+        private IWorkspace currentWorkspace;
```
amcasey: Why store this?  Why not use `this.workspaceService.CurrentWorkspace`?  Does that value change more often than we want?

l14zj:
> I have some thread-safety concerns, but this looks generally correct.

https://help.instagram.com/293775921768331/?helpref=uf_share

PR #1989, opened 2018-06-12
Skip discovery

paulvanbrenk:
Fixes:

> In the latest dogfood (15.8 preview) build, the unit test discovery output has started saying:
> Node.exe was not found. Please install Node.js before running tests.
> It says it several times (~47 for my solution).
> My tests and projects are all .NET (although, there are two Web projects which might have some Javascript in them).

### DIFF
--- a/Nodejs/Product/TestAdapter/JavaScriptTestDiscoverer.cs
+++ b/Nodejs/Product/TestAdapter/JavaScriptTestDiscoverer.cs
@@ -94,7 +94,7 @@ namespace Microsoft.NodejsTools.TestAdapter
                             {
                                 //Check to see if this is a TestCase
                                 testFrameworkName = item.GetMetadataValue("TestFramework");
-                                if (!TestFrameworks.TestFramework.IsValidTestFramework(testFrameworkName))
+                                if (!TestFramework.IsValidTestFramework(testFrameworkName))
                                 {
                                     continue;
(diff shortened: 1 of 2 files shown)


Nodejs/Setup/NodejsTools.vsmanproj
```diff
@@ -48,6 +48,7 @@
   <Target Name="Build" />
   <Target Name="Clean" />
   <Target Name="ReBuild" />
+  <Target Name="BuiltProjectOutputGroup" />
```
paulvanbrenk: This fixes a new build error I'm seeing on my dev machine, this doesn't change the output for this project.

Nodejs/Product/TestAdapter/JavaScriptTestDiscoverer.cs
```diff
@@ -203,10 +206,6 @@ private void DiscoverTests(Dictionary<string, HashSet<TestFileEntry>> testItems,
                 }
                 logger.SendMessage(TestMessageLevel.Informational, string.Format(CultureInfo.CurrentCulture, "Processing finished for framework '{0}'.", testFx));
             }
-            if (testCount == 0)
```
paulvanbrenk: The  test host already prints a warning when no tests are found. So no need to duplicate.

PR #1992, opened 2018-06-27
Set Assembly attributes in targets file

paulvanbrenk:
This way they'll be consistent through out the build, including in
.NET Core/Standard assemblies.


uniqueiniquity:
oops I merged it

PR #2006, opened 2018-07-17
Fix reporting for tape.js

paulvanbrenk:
Since tape doesn't report results in order for the first and last test
we have to keep a list of tests and handle them as their results become
available.


Nodejs/Product/TestAdapter/TestFrameworks/Tape/tape.js
```diff
@@ -57,45 +51,55 @@ function run_tests(testInfo, callback) {
...
+                // Test is starting. Reset the result object. Send a "test start" event.
                 callback({
                     'type': 'test start',
                     'title': result.title,
                     'result': result
                 });
                 break;
             case 'assert':
-                if (!capture) break;
+                var result = testState[evt.test];
```
armanio123: I think this case is incorrect. Based on the Tape documentation a test can contain multiple assertions and tape well report the result for each one of them.
This should be change to either store all of the assertions and later check its result during the 'end' case; or assume the test is successful until any failed assertion proves it otherwise (which the previous revision was doing).

PR #2015, opened 2018-07-30
Block debug during npm install

armanio123:
When starting debug during npm install. Visual studio display the message:

Attribute 'program' does not exists (<JavaScript_file>).

This PR aborts starting debug during the initial npm install and display an error message to the user. Note that it doesn't block any subsequent installs.

### DIFF
--- a/Nodejs/Product/Nodejs/Nodejs.cs
+++ b/Nodejs/Product/Nodejs/Nodejs.cs
@@ -179,6 +179,16 @@ namespace Microsoft.NodejsTools
             );
         }
 
+        public static void ShowNpmIsInstalling()
+        {
+            MessageBox.Show(
+                SR.GetString(SR.NpmIsInstalling),
+                SR.ProductName,
+                MessageBoxButtons.OK,
+                MessageBoxIcon.Error
+            );
+        }
+
         public static void ShowNodejsPathNotFound(string path)
         {
             MessageBox.Show(
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
@@ -59,7 +59,7 @@ namespace Microsoft.NodejsTools.Project
             if(NpmInstallCommand.IsInstalling)
             {
                 Nodejs.ShowNpmIsInstalling();
-                return VSConstants.E_ABORT;
+                return VSConstants.S_OK;
             }
 
             if (nodePath == null)
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
@@ -16,6 +16,7 @@ using System.Web;
 using System.Windows.Forms;
 using Microsoft.NodejsTools.Debugger;
 using Microsoft.NodejsTools.Debugger.DebugEngine;
+using Microsoft.NodejsTools.Npm.SPI;
 using Microsoft.NodejsTools.Options;
 using Microsoft.NodejsTools.Telemetry;
 using Microsoft.NodejsTools.TypeScript;
@@ -54,6 +55,12 @@ namespace Microsoft.NodejsTools.Project
         public int LaunchFile(string file, bool debug)
         {
             var nodePath = GetNodePath();
+
+            if(NpmInstallCommand.IsInstalling)
+            {
+                Nodejs.ShowNpmIsInstalling();
+                return VSConstants.E_ABORT;
+            }
 
             if (nodePath == null)
             {
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
@@ -56,9 +56,11 @@ namespace Microsoft.NodejsTools.Project
         {
             var nodePath = GetNodePath();
 
-            if(NpmInstallCommand.IsInstalling)
+            if(this._project.IsInstallingMissingModules)
             {
                 Nodejs.ShowNpmIsInstalling();
+                this._project.NpmOutputPane?.Show();
+
                 return VSConstants.S_OK;
             }
 
--- a/Nodejs/Product/Nodejs/Project/ProjectResources.cs
+++ b/Nodejs/Product/Nodejs/Project/ProjectResources.cs
@@ -13,6 +13,7 @@ namespace Microsoft.NodejsTools.Project
 
         internal const string NodeExeDoesntExist = "NodeExeDoesntExist";
         internal const string NodejsNotInstalled = "NodejsNotInstalled";
+        internal const string NpmIsInstalling = "NpmIsInstalling";
 
         internal const string TestFramework = "TestFramework";
 
--- a/Nodejs/Product/Npm/SPI/AbstractNpmLogSource.cs
+++ b/Nodejs/Product/Npm/SPI/AbstractNpmLogSource.cs
@@ -4,7 +4,7 @@ using System;
 
 namespace Microsoft.NodejsTools.Npm.SPI
 {
-    public abstract class AbstractNpmLogSource : INpmLogSource
+    internal abstract class AbstractNpmLogSource : INpmLogSource
     {
         public event EventHandler CommandStarted;
 
--- a/Nodejs/Product/Npm/SPI/AbstractNpmLogSource.cs
+++ b/Nodejs/Product/Npm/SPI/AbstractNpmLogSource.cs
@@ -1,10 +1,10 @@
-// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
 
 namespace Microsoft.NodejsTools.Npm.SPI
 {
-    internal abstract class AbstractNpmLogSource : INpmLogSource
+    public abstract class AbstractNpmLogSource : INpmLogSource
     {
         public event EventHandler CommandStarted;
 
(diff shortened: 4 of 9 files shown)


armanio123:
This is how the error message looks like: 

image

I checked how .net core projects handle this scenario (for having a uniform experience) but they don't handle it. Instead you cannot compile the project as it wont find the necessary libraries. So I decided on displaying a pop up  message as NTVS normally does.

Also, we can improve this scenario by reporting the background task on the lower left corner. We discuss more about this.

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -55,6 +56,12 @@ public int LaunchFile(string file, bool debug)
         {
             var nodePath = GetNodePath();
 
+            if(NpmInstallCommand.IsInstalling)
+            {
+                Nodejs.ShowNpmIsInstalling();
+                return VSConstants.E_ABORT;
```
amcasey: It looks like the error case below returns `S_OK`.

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -55,6 +56,12 @@ public int LaunchFile(string file, bool debug)
         {
             var nodePath = GetNodePath();
 
+            if(NpmInstallCommand.IsInstalling)
+            {
+                Nodejs.ShowNpmIsInstalling();
```
amcasey: >ShowNpmIsInstalling [](start = 23, length = 19)

Are we on an appropriate thread?  From below, I assume so, but it might be nice to document.

Nodejs/Product/Nodejs/Nodejs.cs
```diff
@@ -179,6 +179,16 @@ public static void ShowNodejsNotInstalled()
             );
         }
 
+        public static void ShowNpmIsInstalling()
```
amcasey: >ShowNpmIsInstalling [](start = 27, length = 19)

I'm a little surprised we have a separate method for each error message.

Nodejs/Product/Nodejs/Project/ProjectResources.cs
```diff
@@ -13,6 +13,7 @@ internal class SR : CommonSR
 
         internal const string NodeExeDoesntExist = "NodeExeDoesntExist";
         internal const string NodejsNotInstalled = "NodejsNotInstalled";
+        internal const string NpmIsInstalling = "NpmIsInstalling";
```
amcasey: >NpmIsInstalling [](start = 30, length = 15)

What are these for?

Nodejs/Product/Npm/SPI/AbstractNpmLogSource.cs
```diff
@@ -1,10 +1,10 @@
-// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
 
 namespace Microsoft.NodejsTools.Npm.SPI
 {
-    internal abstract class AbstractNpmLogSource : INpmLogSource
+    public abstract class AbstractNpmLogSource : INpmLogSource
```
amcasey: >public [](start = 4, length = 6)

Why did this change?

Nodejs/Product/Npm/SPI/NpmInstallCommand.cs
```diff
@@ -2,15 +2,20 @@
...
         public NpmInstallCommand(
             string fullPathToRootPackageDirectory,
             string pathToNpm = null,
             bool useFallbackIfNpmNotFound = true)
             : base(fullPathToRootPackageDirectory, showConsole: false, pathToNpm: pathToNpm)
         {
             this.Arguments = "install";
+
+            this.CommandStarted += (sender, eventArgs) => { IsInstalling = true; };
+            this.CommandStarted += (sender, eventArgs) => { IsInstalling = false; };
```
amcasey: >CommandStarted [](start = 17, length = 14)

Was this supposed to be `CommandStopped`?

Nodejs/Product/Npm/SPI/NpmInstallCommand.cs
```diff
@@ -2,15 +2,20 @@
...
+
         public NpmInstallCommand(
             string fullPathToRootPackageDirectory,
             string pathToNpm = null,
             bool useFallbackIfNpmNotFound = true)
             : base(fullPathToRootPackageDirectory, showConsole: false, pathToNpm: pathToNpm)
         {
             this.Arguments = "install";
+
+            this.CommandStarted += (sender, eventArgs) => { IsInstalling = true; };
```
amcasey: >CommandStarted [](start = 17, length = 14)

When is this unsubscribed?  Does it matter?

amcasey: It seems like it would be more natural for the consumer of the state changes to subscribe to the event, but I see that the event doesn't expose *which* command started.

---
In reply to: [206370382](https://github.com/Microsoft/nodejstools/pull/2015#discussion_r206370382) [](ancestors = 206370382,206369659)

Nodejs/Product/Npm/SPI/NpmInstallCommand.cs
```diff
@@ -2,15 +2,20 @@
 
 namespace Microsoft.NodejsTools.Npm.SPI
 {
-    internal class NpmInstallCommand : NpmCommand
+    public class NpmInstallCommand : NpmCommand
     {
+        public static bool IsInstalling { get; private set; } = false;
```
amcasey: >IsInstalling [](start = 27, length = 12)

What happens if the debugging session checks the flag and *then* npm starts installing?

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
armanio123: Changed.

Nodejs/Product/Nodejs/Project/ProjectResources.cs
armanio123: This are the resource key, for the error message.

Nodejs/Product/Npm/SPI/AbstractNpmLogSource.cs
armanio123: NpmInstallCommand needs to be public in order to be accessed on the Nodejs assembly.

Nodejs/Product/Npm/SPI/NpmInstallCommand.cs
armanio123: Good, catch those where last minute changes and I didn't noticed that I have used the incorrect handler. 
Also the code that executed the event handlers didn't handle exceptions correctly. I have refactored that code as well. Check `NpmCommand.cs` file.

Nodejs/Product/Npm/SPI/NpmInstallCommand.cs
armanio123: This is not unsubscribed, I don't think it matters tho.

I understood the `AbstractNpmLogSource` as a template class that provides a "hook" for when a command has started or finished. In consequence the `NpmInstallCommand` provides the implementation. 

The other place that might be a good option for this would be the class `NpmCommander`, which is actually the one that consumes the command. Seems like this is not the correct place to me though, as I think this should be just dumb and don't know the implementation details of the command.

Let me know what you think.

Nodejs/Product/Npm/SPI/NpmInstallCommand.cs
armanio123: This scenario cannot happen because starting npm install happens before loading the project. Check comment on `NpmWizardExtension.cs:27`

```
// Create the "node_modules/@types" folder before opening any files (which creates the
// context for the project). This allows tsserver to start watching for type definitions
// before any are installed (which is needed as "npm install" runs async, so the modules
// usually aren't installed before the project context is loaded).
```

Nodejs/Product/Npm/SPI/AbstractNpmLogSource.cs
armanio123: The latest changes doesn't require this anymore. Rollback.

Nodejs/Product/Npm/SPI/NpmCommand.cs
```diff
@@ -74,38 +74,46 @@ public void CancelCurrentTask()
...
-            }
-            catch (NpmNotFoundException)
-            {
-                redirector?.WriteErrorLine(Resources.CouldNotFindNpm);
-                return false;
-            }
-            redirector?.WriteLine(
-                string.Format(CultureInfo.InvariantCulture, "===={0}====\r\n\r\n",
-                string.Format(CultureInfo.InvariantCulture, Resources.ExecutingCommand, this.Arguments)));
+                this.GetPathToNpm();
```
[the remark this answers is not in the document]
amcasey: It looks like it's just for the `NpmNotFoundException`.  Can you please add a comment to that effect?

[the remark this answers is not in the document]
amcasey: Also, consider storing the result in a local, rather than refetching it below.

Nodejs/Product/Nodejs/Project/ProjectResources.cs
amcasey: It might be nice to use `nameof`.

Nodejs/Product/Npm/SPI/NpmCommand.cs
```diff
@@ -74,38 +74,46 @@ public void CancelCurrentTask()
...
+                return false;
+            }
             catch (OperationCanceledException)
             {
                 cancelled = true;
             }
-            OnCommandCompleted(this.Arguments, redirector?.HasErrors ?? false, cancelled);
+            catch (Exception)
+            {
+                // Catch exception to run the finally block during an unhandled exception.
```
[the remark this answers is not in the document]
armanio123: Fixed.

Nodejs/Product/Npm/SPI/NpmCommand.cs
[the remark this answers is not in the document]
armanio123: Comment and storing the result added.

armanio123:
Yes, I'm looking at that right now I'll send another PR as soon as is ready.

PR #2017, opened 2018-07-31
Fix unhandled cast

armanio123:
Fixes issue with an invalid project cast


billti:
Be sure to port this to the release-15.9 and master branches also to ensure it doesn't pop up again. Thanks!

PR #1996, opened 2018-07-03
Add jest support

armanio123:
Adds the Jest test adapter.

PR #2019, opened 2018-07-31
Revert "Add jest support" on release-15.9

armanio123:
Reverts Microsoft/nodejstools#1996


armanio123:
Jest is going to be included on 15.9. No need to revert this.

PR #2040, opened 2018-08-11
Removed process.exit for process.exitCode on Mocha test adapter

armanio123:
Fixes an issue when the process.stdout has not finished reporting the test result and the node process is killed.

With this the node process will wait until the I/O has finished all operations to be closed.

### DIFF
--- a/Nodejs/Product/TestAdapter/TestFrameworks/mocha/mocha.js
+++ b/Nodejs/Product/TestAdapter/TestFrameworks/mocha/mocha.js
@@ -111,7 +111,7 @@ var run_tests = function (testCases, callback) {
     mocha.addFile(testCases[0].testFile);
 
     var runner = mocha.run(function (code) {
-        process.exit(code);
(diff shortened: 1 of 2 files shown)


Nodejs/Product/TestAdapter/TestFrameworks/mocha/mocha.js
```diff
@@ -111,7 +111,7 @@ var run_tests = function (testCases, callback) {
     mocha.addFile(testCases[0].testFile);
 
     var runner = mocha.run(function (code) {
-        process.exit(code);
+        process.exitCode = code ? code : 0;
```
billti: So this no longer exits the process when called, just sets the exit code for when it exits later by some other mechanism. Is that right?

armanio123: That's correct. With this NodeJs will close itself after all operations have finished instead of killing and truncating the stdout.

Issue #2055, opened 2018-08-29
TestAdapter: Test discovery fails on test projects that contain a large number of test files


lukedgr:
We have Mocha Unit Test projects with 100-150 test files each, and test discovery now fails because the command line to execute node.js test discovery from C# exceeds the windows command line length limit of 8191 characters.

Tests were discovered without issue for the same test projects using Visual Studio Version 15.7.6 and NTVS 1.4.11027.3

##### Expected Behavior 
All Tests are discovered

##### Actual Behavior 
[8/29/2018 7:22:38 AM Error] The command line is too long. And no tests are visible for that project in the test explorer.

- NTVS Version: 1.4.20802.1 Commit Hash:97e1085d8b4b8e3e51c398e910177f87e86d135e
- Visual Studio Version: 15.8.2
- Node.js Version: All

##### Steps to Reproduce
1. Create a unit test project with enough test files so that the command line to execute node.js test discovery is > 8191 characters

##### Potential Fix: 
Write the list of test files to a file and pass that file path to the node.js process instead of passing all of the test files to the node.js process through the command line.

armanio123:
Issue has been resolved. Thanks @lukedgr for fixing it.

PR #2064, opened 2018-09-07
2055: TestAdapter: Test discovery fails on test projects that contain a large number of test files

lukedgr:
Issue #
 #2055 
##### Bug
Test discovery fails for unit test projects with 100+ test files because the command line to execute node.js test discovery from C# exceeds the windows command line length limit of 8191 characters.

##### Fix
Updated the communication between the TestFramework class in C# and find_tests.js to provide the file path for a temp file that contains the list of test files in which to discover tests.  find_tests.js then reads the contents of the provided file to get the list of test files.  This prevents exceeding the command line length maximum for projects with a large number of test files.

##### Testing
Validated that test discovery works on large projects where it failed previously

### DIFF
--- a/Nodejs/Product/TestAdapter/TestFrameworks/TestFramework.cs
+++ b/Nodejs/Product/TestAdapter/TestFrameworks/TestFramework.cs
@@ -39,7 +39,7 @@ namespace Microsoft.NodejsTools.TestAdapter.TestFrameworks
             // Create a temp file with the list of test files.
             var testFilesString = string.Join(";", testFiles);
             var testFilesListFile = Path.GetTempFileName();
-            File.AppendAllText(testFilesListFile, testFilesString, System.Text.Encoding.UTF8);
+            File.WriteAllText(testFilesListFile, testFilesString, System.Text.Encoding.UTF8);
 
             try
             {
--- a/Nodejs/Product/TestAdapter/TestFrameworks/TestFramework.cs
+++ b/Nodejs/Product/TestAdapter/TestFrameworks/TestFramework.cs
@@ -35,9 +35,15 @@ namespace Microsoft.NodejsTools.TestAdapter.TestFrameworks
         {
             var testInfo = string.Empty;
             var discoverResultFile = Path.GetTempFileName();
+
+            // Create a temp file with the list of test files.
+            var testFilesString = string.Join(";", testFiles);
+            var testFilesListFile = Path.GetTempFileName();
+            File.AppendAllText(testFilesListFile, testFilesString, System.Text.Encoding.UTF8);
+
             try
             {
-                var stdout = EvaluateJavaScript(nodeExe, string.Join(";", testFiles), discoverResultFile, logger, projectRoot);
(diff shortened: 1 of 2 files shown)


armanio123:
@minestarks @amcasey This PR fixes an issue with the Test Adapter where tests discovery and execution is blocked if the tests path exceeds the command line limit.

Is this something we want to push for 15.9 ?

Nodejs/Product/TestAdapter/TestFrameworks/TestFramework.cs
```diff
@@ -35,9 +35,15 @@ public TestFramework(string vsixScriptFolder)
         {
             var testInfo = string.Empty;
             var discoverResultFile = Path.GetTempFileName();
+
+            // Create a temp file with the list of test files.
+            var testFilesString = string.Join(";", testFiles);
+            var testFilesListFile = Path.GetTempFileName();
+            File.AppendAllText(testFilesListFile, testFilesString, System.Text.Encoding.UTF8);
```
amcasey: I would probably `WriteAllText`.  I realize the file should be empty, but I think it's a clearer way to express that you intend to completely determine its contents.

Nodejs/Product/TestAdapter/TestFrameworks/find_tests.js
```diff
@@ -6,7 +7,19 @@ try {
     process.exit(1);
 }
 try {
-    framework.find_tests(process.argv[3], process.argv[4], process.argv[5]);
+    var testFilesListInputFile = process.argv[3];
+    if (!fs.existsSync(testFilesListInputFile)) {
+        throw new Error("testFilesListInputFile '" + testFilesListInputFile + "' does not exist.");
```
amcasey: It seems strange to throw within a try block.

Nodejs/Product/TestAdapter/TestFrameworks/TestFramework.cs
```diff
@@ -35,9 +35,15 @@ public TestFramework(string vsixScriptFolder)
...
+
             try
             {
-                var stdout = EvaluateJavaScript(nodeExe, string.Join(";", testFiles), discoverResultFile, logger, projectRoot);
+                var stdout = EvaluateJavaScript(nodeExe, testFilesListFile, discoverResultFile, logger, projectRoot);
```
amcasey: Is this the only call to `EvaluateJavaScript`?

armanio123: Yes it's the only call, I searched for all references in the project.

Nodejs/Product/TestAdapter/TestFrameworks/TestFramework.cs
armanio123: Changed.

armanio123:
Thanks @lukedgr for fixing the issue.

armanio123:
Ported to 15.9. PR #2070

PR #2081, opened 2018-09-25
Fix chrome admin issue

armanio123:
Chrome fails to open web pages when launched in Admin context. 

This fix will always launch chrome un-elevated.


Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -468,6 +464,29 @@ private bool ShouldStartBrowser()
             return true;
         }
 
+        private string GetBrowserPath()
+        {
+            ThreadHelper.ThrowIfNotOnUIThread(nameof(GetBrowserPath));
```
amcasey: That function uses `CallerMemberNameAttribute` so the argument will be provided automatically.

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -468,6 +464,29 @@ private bool ShouldStartBrowser()
...
+
+            return defaultBrowserPath;
+        }
+
+        private static void OpenBrowser(string browserPath, string webBrowserUrl)
+        {
+            // Chrome has known issues with being launched as admin.
+            if (!string.IsNullOrEmpty(browserPath) && browserPath.EndsWith(NodejsConstants.ChromeApplicationName))
+            {
+                SystemUtility.ExecuteProcessUnElevated(browserPath, webBrowserUrl);
```
amcasey: Any reason not to launch all browsers unelevated?

armanio123: Since I just tested this for Chrome, I decided it's a good idea to change only the behavior we know is broken. In any case, if you feel like we should always launch un-elevated I can do that.

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
armanio123: Fixed.

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -468,6 +464,29 @@ private bool ShouldStartBrowser()
...
+            }
+            else
+            {
+                VsShellUtilities.OpenBrowser(webBrowserUrl, (uint)__VSOSPFLAGS.OSP_LaunchNewBrowser);
```
billti: I assume we already don't support attaching for client-side script debugging on launch, and thus this has no impact on debugging client side script (though may complicate matters if we want to support that later).

Nodejs/Product/Nodejs/SharedProject/SystemUtilities.cs
```diff
@@ -0,0 +1,142 @@
...
+        /// </summary>
+        internal static void ExecuteProcessUnElevated(string process, string args, string currentDirectory = "")
+        {
+            var shellWindows = (IShellWindows)new CShellWindows();
+
+            // Get the desktop window
+            object loc = CSIDL_Desktop;
+            object unused = new object();
+            int hwnd;
+            var serviceProvider = (IServiceProvider)shellWindows.FindWindowSW(ref loc, ref unused, SWC_DESKTOP, out hwnd, SWFO_NEEDDISPATCH);
```
billti: I spent a while trying to read the docs to understand these interfaces, methods, and args, and.... 😕 

If this is taken from other working code, and you've tested it thoroughly, then seems OK. But it's not the most intuitive or readable code we've checked in.

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -468,6 +464,29 @@ private bool ShouldStartBrowser()
             return true;
         }
 
+        private string GetBrowserPath()
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
+            var uiShellOpenDocument = (IVsUIShellOpenDocument)ServiceProvider.GlobalProvider.GetService(typeof(SVsUIShellOpenDocument));
+            uiShellOpenDocument.GetFirstDefaultPreviewer(out var defaultBrowserPath, out _, out _);
```
billti: The docs state this method: "Returns S_OK if a default previewer is defined. Returns S_FALSE if there are no default previewers". You don't seem to allow for the latter case here.

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -468,6 +464,29 @@ private bool ShouldStartBrowser()
...
+
+            return defaultBrowserPath;
+        }
+
+        private static void OpenBrowser(string browserPath, string webBrowserUrl)
+        {
+            // Chrome has known issues with being launched as admin.
+            if (!string.IsNullOrEmpty(browserPath) && browserPath.EndsWith(NodejsConstants.ChromeApplicationName))
```
billti: I guess you do check here if the browserPath was set correctly, so the previous comment about not handling failure on the call to GetFirstDefaultPreviewer is likely benign.

Nodejs/Product/Nodejs/SharedProject/SystemUtilities.cs
armanio123: I agree, I tested the most common browsers and through different launch scenarios and everything looked good. 

The code has been taken from the Visual Studio installer and it's also the same code Diagnostics has recently adopted for this same Chrome issue.

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
armanio123: This is as intended as we later check if the browserPath exists.

armanio123:
I'll make sure to explicitly request CTI the tests you mentioned on the next round of tests.

PR #2083, opened 2018-09-27
Code defensively when casting IVsBuildPropertyStorage

armanio123:
This code fixes an unhandled exception when trying to cast to IVsBuildPropertyStorage.

### DIFF
--- a/Nodejs/Product/TypeScript/TypeScriptHelpers.cs
+++ b/Nodejs/Product/TypeScript/TypeScriptHelpers.cs
@@ -61,8 +61,16 @@ namespace Microsoft.NodejsTools.TypeScript
         internal static string GetTypeScriptBackedJavaScriptFile(IVsProject project, string pathToFile)
         {
             //Need to deal with the format being relative and explicit
-            var props = (IVsBuildPropertyStorage)project;
-            ErrorHandler.ThrowOnFailure(props.GetPropertyValue(NodeProjectProperty.TypeScriptOutDir, null, (uint)_PersistStorageType.PST_PROJECT_FILE, out var outDir));
+            string outDir = null;
+
+            if (project is IVsBuildPropertyStorage props)
+            {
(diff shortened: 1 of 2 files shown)


Nodejs/Product/TypeScript/TypeScriptHelpers.cs
```diff
@@ -61,8 +61,16 @@ private static string GetTypeScriptBackedJavaScriptFile(string projectHome, stri
...
+            string outDir = null;
+
+            if (project is IVsBuildPropertyStorage props)
+            {
+                ErrorHandler.ThrowOnFailure(props.GetPropertyValue(NodeProjectProperty.TypeScriptOutDir, null, (uint)_PersistStorageType.PST_PROJECT_FILE, out outDir));
+            }
+            else
+            {
+                Debug.Fail($"Why is {nameof(project)} not of type {nameof(IVsBuildPropertyStorage)}?");
+            }
```
billti: So isn't it going to fall through to line 77 if it couldn't cast 'props', and thus outdir will be null in the call in line 77 (as line 68 wouldn't have been called)? Seems like that would just cause other failures down the line.

billti: (Though arguably is already "ErrorHandler.ThrowOnFailure" if it can't get that property, so maybe should also throw at that point if the cast fails)

armanio123: GetTypeScriptBackedJavaScriptFile handles when outDir is null, it considers it as if no setting is configured so it will set the js file in the same path as the ts file. Arguable this might be an error as the user might expect the js file to be on another place or to not exist at all.

GetTypeScriptBackedJavaScriptFile(IVsProject, string) has two callers. Only one has thrown this exception (TestAdapterImpl) and it already handles the case where the js returned doesn't exists by ignoring the result and indicating that it shouldn't not discover test files. Which sounds like the correct result to me. If we go this way we might want to reconsider throwing an exception when it can't get the outDir as well.

Said all that, I do agree that this is changing the current logic and minimizing the changes is a safer bet but throwing an exception doesn't seem to fix the issue.

Let me know what you think.

PR #2116, opened 2018-12-10
Added FailFast to OnItemAdded

armanio123:
Added a FailFast to get better dumps whenever a "parent" node is null or whitespace. The condition should never be true. Nevertheless, it's been added to get information of a crash happening when updating the Hierarchy nodes.

### DIFF
--- a/Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
@@ -6476,6 +6476,8 @@ If the files in the existing folder have the same names as files in the folder y
 
             if (string.IsNullOrWhiteSpace(parent.GetItemName()))
             {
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
```diff
@@ -6473,6 +6473,14 @@ internal void OnItemAdded(HierarchyNode parent, HierarchyNode child, HierarchyNo
 
             var prev = previousVisible ?? child.PreviousVisibleSibling;
             var prevId = (prev != null) ? prev.HierarchyId : VSConstants.VSITEMID_NIL;
+
+            if (string.IsNullOrWhiteSpace(parent.GetItemName()))
```
amcasey: I'd be happier checking the HRESULT - I'm worried that null and empty might be valid values.  We're just trying to check for the absence of a node.

We don't have a way to simulate the failure?

armanio123:
Closing as we decided to go for etw logging. Implemented in this PR #2118

PR #2120, opened 2018-12-21
Fix VS crash in test file discovery upon file change notifications

minestarks:
Fixes https://devdiv.visualstudio.com/DevDiv/_queries/edit/731590 (more details in bug)

`GetTestProjectFromFile` can return projects that are not NTVS projects. Later we try to get the `ProjectHome` property from that project, which throws because that property isn't necessarily defined on every project. Per contract `EnvDTE.Properties.Item` throws `ArgumentException` if the property isn't found. 

The exception _can_ propagate and cause VS to crash, depending on context. In most cases it seems to be swallowed further up the stack, but it really should be handled at the originating site.

A similar thing will happen when getting the `TypeScriptOutDir` property. Some project types (e.g. ASP.NET with .NET Framework) will throw a `COMException` when the property isn't found. This should also be handled.

To handle these exceptions, the most sensible thing to do in this context would be to ignore the notification since the project clearly isn't supported. There is probably a smarter solution here that would prevent the notifications from getting fired in the first place, but I'm not familiar enough with the design to think of it.

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
@@ -522,6 +522,11 @@ namespace Microsoft.NodejsTools.Project
             if (TypeScriptHelpers.IsTypeScriptFile(startupFile))
             {
                 startupFile = TypeScriptHelpers.GetTypeScriptBackedJavaScriptFile(this._project, startupFile);
+                if (startupFile == null)
+                {
+                    // Expected to find a JS file
+                    throw new ArgumentException();
+                }
             }
             return startupFile;
         }
--- a/Nodejs/Product/TypeScript/TypeScriptHelpers.cs
+++ b/Nodejs/Product/TypeScript/TypeScriptHelpers.cs
@@ -89,12 +89,9 @@ namespace Microsoft.NodejsTools.TypeScript
                 {
                     ErrorHandler.ThrowOnFailure(props.GetPropertyValue(NodeProjectProperty.TypeScriptOutDir, null, (uint)_PersistStorageType.PST_PROJECT_FILE, out outDir));
                 }
-                catch (System.Runtime.InteropServices.COMException e)
+                catch (System.Runtime.InteropServices.COMException e) when (e.ErrorCode == ERR_XML_ATTRIBUTE_NOT_FOUND)
                 {
-                    if (e.ErrorCode == ERR_XML_ATTRIBUTE_NOT_FOUND)
-                    {
-                        return null;
-                    }
+                    return null;
                 }
             }
             else
--- a/Nodejs/Product/TypeScript/TypeScriptHelpers.cs
+++ b/Nodejs/Product/TypeScript/TypeScriptHelpers.cs
@@ -82,7 +82,20 @@ namespace Microsoft.NodejsTools.TypeScript
 
             if (project is IVsBuildPropertyStorage props)
             {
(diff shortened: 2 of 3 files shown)


Nodejs/Product/TypeScript/TypeScriptHelpers.cs
```diff
@@ -82,7 +82,20 @@ internal static string GetTypeScriptBackedJavaScriptFile(IVsProject project, str
...
+                // GetProperty can return this error code if the property doesn't exist
+                const int ERR_XML_ATTRIBUTE_NOT_FOUND = unchecked((int)0x8004C738);
+
+                try
+                {
+                    ErrorHandler.ThrowOnFailure(props.GetPropertyValue(NodeProjectProperty.TypeScriptOutDir, null, (uint)_PersistStorageType.PST_PROJECT_FILE, out outDir));
+                }
+                catch (System.Runtime.InteropServices.COMException e)
+                {
+                    if (e.ErrorCode == ERR_XML_ATTRIBUTE_NOT_FOUND)
```
armanio123: if the `ErrorCode` is not `ERR_XML_ATTRIBUTE_NOT_FOUND`, should it `throw`?

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -522,6 +522,11 @@ private string ResolveStartupFile()
             if (TypeScriptHelpers.IsTypeScriptFile(startupFile))
             {
                 startupFile = TypeScriptHelpers.GetTypeScriptBackedJavaScriptFile(this._project, startupFile);
+                if (startupFile == null)
```
armanio123: I'm concerned that returning `null` affect the scenarios where is an NTVS project that has not defined `TypeScriptOutDir` but the user expect to have a `js` file created on the same place as the `ts` file.

Nodejs/Product/TypeScript/TypeScriptHelpers.cs
```diff
@@ -82,7 +82,20 @@ internal static string GetTypeScriptBackedJavaScriptFile(IVsProject project, str
...
             {
-                ErrorHandler.ThrowOnFailure(props.GetPropertyValue(NodeProjectProperty.TypeScriptOutDir, null, (uint)_PersistStorageType.PST_PROJECT_FILE, out outDir));
+                // GetProperty can return this error code if the property doesn't exist
+                const int ERR_XML_ATTRIBUTE_NOT_FOUND = unchecked((int)0x8004C738);
+
+                try
+                {
+                    ErrorHandler.ThrowOnFailure(props.GetPropertyValue(NodeProjectProperty.TypeScriptOutDir, null, (uint)_PersistStorageType.PST_PROJECT_FILE, out outDir));
+                }
+                catch (System.Runtime.InteropServices.COMException e)
```
armanio123: Instead add `when` to the `catch`.

Nodejs/Product/TypeScript/TypeScriptHelpers.cs
minestarks: Oops, that's what I meant to do. Good catch!

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
minestarks: Sure, we may want to handle those scenarios, bu today those scenarios throw/crash. So my claim is I'm only making it better :)  In this PR my intention is to gracefully handle the cases where we throw known exceptions today.

Note that the existing code did handle `outDir` being null in cases where we didn't throw.

minestarks:
Thanks for the catch, updated. Look good?

PR #2122, opened 2018-12-28
Make VSHPROPID_IsNonSearchable return true for project files

minestarks:
This fix is copied from https://github.com/Microsoft/PTVS/pull/4910 and fixes the exact same issue.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/756502 among others (described in comment and linked PTVS issue).


minestarks:
@armanio123 please review when you get the change -- I will merge this now so we can get a build out ASAP.

PR #2123, opened 2019-01-09
Use TypeScript compiler from the root node_modules instead of the TypeScript SDK

armanio123:
Removes some of the TypeScript SDK dependencies and uses the installed TypeScript compiler from the the root node_modules folder.


Nodejs/Product/Nodejs/NodejsConstants.cs
```diff
@@ -102,9 +100,9 @@ internal static class NodeProjectProperty
         public const string ScriptArguments = "ScriptArguments";
         public const string StartWebBrowser = "StartWebBrowser";
         public const string TypeScriptCfgProperty = "CfgPropertyPagesGuidsAddTypeScript";
-        public const string TypeScriptModuleKind = "TypeScriptModuleKind";
-        public const string TypeScriptOutDir = "TypeScriptOutDir";
-        public const string TypeScriptSourceMap = "TypeScriptSourceMap";
+        // This property is used by the SDK and should be removed. Nevertheless, this is also used by the TestAdapter to find the JS tests. 
+        // My preference is to get the ts configuration to find out the location but might be easier to just repurpose the property.
+        public const string TypeScriptOutDir = "TypeScriptOutDir";
```
armanio123: This essentially means that the user will need to specify where the outdir is located for tests files on the njsproj file instead of using the tsconfig file.

Nodejs/Product/TargetsVsix/Microsoft.NodejsTools.targets
```diff
@@ -55,9 +53,18 @@
...
+  </Target>
+  
+  <Target Name="AfterCompile">
+    <Message Text="AfterCompile executed" />
+    <Exec Condition="'$(EnableTypeScript)' == 'true'" Command="&quot;$(Tsc)&quot; --build" />
+  </Target>
 
   <Target Name="CoreCompile">
-    <Error Condition="!Exists('$(VSToolsPath)\TypeScript\Microsoft.TypeScript.targets') And '$(EnableTypeScript)' == 'true' And '$(VisualStudioVersion)' != '16.0'" Text="TypeScript is not installed but TypeScript support is enabled in this project.  Please download and install TypeScript http://go.microsoft.com/fwlink/?LinkID=393159&amp;clcid=0x409."></Error>
+    <Error Condition="'$(EnableTypeScript)' == 'true' And !Exists('$(Tsc)')" Text="TypeScript compiler is not installed. Please execute 'npm install --save-dev typescript'"></Error>
```
armanio123: Should we tell the user to use the NPM UI on Visual Studio instead?

Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureNodejsWorker/Worker.njsproj
```diff
@@ -34,13 +31,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <TypeScriptCompile Include="server.ts" />
-    <TypeScriptCompile Include="startup.ts" />
+    <Compile Include="server.ts" />
```
armanio123: The "Compile" action gets ignores by `tsc` as it's now configured by `tsconfig.json`. Regardless, it's still needed in order for VS to keep track of `dirty` files and to trigger the build if necessary.

I have not found a way to avoid this.

Nodejs/Product/TargetsVsix/Microsoft.NodejsTools.targets
```diff
@@ -55,9 +53,18 @@
...
 
   <Target Name="CopyFilesToOutputDirectory"/>
+  
+  <Target Name="AfterClean">
+    <Exec Condition="'$(EnableTypeScript)' == 'true'" Command="&quot;$(Tsc)&quot; --build --clean" />
+  </Target>
+  
+  <Target Name="AfterCompile">
+    <Message Text="AfterCompile executed" />
+    <Exec Condition="'$(EnableTypeScript)' == 'true'" Command="&quot;$(Tsc)&quot; --build" />
```
armanio123: Having `--build ` flag enabled will force the user to have a `tsconfig.json` file in the project.

Nodejs/Product/Nodejs/NodejsConstants.cs
minestarks: This has always been true, right?

---
In reply to: [246554781](https://github.com/Microsoft/nodejstools/pull/2123#discussion_r246554781) [](ancestors = 246554781)

armanio123: It has.

Nodejs/Product/TargetsVsix/Microsoft.NodejsTools.targets
```diff
@@ -55,9 +53,18 @@
   </ItemGroup>
 
   <Target Name="CopyFilesToOutputDirectory"/>
+  
+  <Target Name="AfterClean">
+    <Exec Condition="'$(EnableTypeScript)' == 'true'" Command="&quot;$(Tsc)&quot; --build --clean" />
+  </Target>
+  
+  <Target Name="AfterCompile">
+    <Message Text="AfterCompile executed" />
```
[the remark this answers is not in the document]
armanio123: It was not, fixed.

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -132,12 +132,6 @@ protected override void AddNewFileNodeToHierarchy(HierarchyNode parentNode, stri
             {
                 // enable TypeScript on the project automatically...
                 SetProjectProperty(NodeProjectProperty.EnableTypeScript, "true");
-                SetProjectProperty(NodeProjectProperty.TypeScriptSourceMap, "true");
```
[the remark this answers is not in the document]
armanio123: This properties will no longer serve a purpose because they configure the TypeScript SDK, which we no longer use.

It will be incompatible with older templates or with projects not containing a `tsconfig.json` file. We'll still be able to open the project but build will fail informing to the user the need for a `tsconfig.json` file.

I'm not sure if this clarifies your questions.

armanio123:
Based on the latest conversation with @minestarks we need to do a couple of changes to the scope:
- We no longer want to completely remove the SDK but to remove it from the templates as we want to maintain compatibility with the previous templates for now.
- Execute the build & clean from an npm script intead fore new project templates.

Nodejs/Product/Nodejs/NodejsConstants.cs
```diff
@@ -44,7 +44,12 @@ internal static class NodejsConstants
 
         public const ushort DefaultDebuggerPort = 5858;
 
+        public const string CompileItemType = "Compile";
```
amcasey: It would be nice to clean these up if it turns out we can update the targets file instead.

Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureExpressApp/TypeScriptExpressApp.njsproj
```diff
@@ -31,9 +28,9 @@
     <DebugSymbols>true</DebugSymbols>
   </PropertyGroup>
   <ItemGroup>
-    <TypeScriptCompile Include="app.ts" />
-    <TypeScriptCompile Include="routes\index.ts" />
-    <TypeScriptCompile Include="routes\user.ts" />
+    <Compile Include="app.ts" />
+    <Compile Include="routes\index.ts" />
+    <Compile Include="routes\user.ts" />
     <Content Include="tsconfig.json" />
```
amcasey: If `<Compile>` marks the file that trigger a rebuild, it seems like tsconfig.json should be `<Compile>`.

Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptConsoleApp/package.json
```diff
@@ -6,7 +6,12 @@
   "author": {
     "name": ""
   },
+  "scripts": {
+    "build": "tsc --build",
+    "clean": "tsc --build --clean"
```
amcasey: Naïve question: why no `start` for console apps?

Nodejs/Product/Nodejs/Project/ImportWizard/ImportSettings.cs
```diff
@@ -392,13 +389,7 @@ ProjectLanguage projectLanguage
 
             if (typeScriptSupport)
             {
-                writer.WriteElementString("TypeScriptSourceMap", "true");
-                writer.WriteElementString("TypeScriptModuleKind", "CommonJS");
                 writer.WriteElementString("EnableTypeScript", "true");
-                if (typeScriptVersion != null)
-                {
-                    writer.WriteElementString("TypeScriptToolsVersion", typeScriptVersion);
```
amcasey: Are we giving up the ability to fix people on a particular TS version?

Nodejs/Product/Nodejs/ProjectTemplates/VuejsApp/VuejsApp.njsproj
```diff
@@ -45,18 +45,17 @@
     <Folder Include="src\components\" />
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="babel.config.js" />
-    <Compile Include="src\main.js" />
+    <Content Include="babel.config.js" />
```
amcasey: Does this template already have an appropriate package.json file?

Nodejs/Product/TargetsVsix/Microsoft.NodejsToolsV2.targets
```diff
@@ -0,0 +1,313 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">14.0</VisualStudioVersion>
```
amcasey: Is "14.0" the appropriate value if this will only be used with new projects?

Nodejs/Product/TargetsVsix/Microsoft.NodejsToolsV2.targets
```diff
@@ -0,0 +1,313 @@
...
+    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- Specifies the directories which are ignored.  Any paths which contain these directory names
+         will be ignored.  The directory name must completely match and no wild cards are allowed
+     -->
+    <AnalysisIgnoredDirectories Condition="'$(AnalysisIgnoredDirectories)' == ''">bower_components;node_modules\core-js</AnalysisIgnoredDirectories>
+    <DockerExtraPublishProfileProperties>ExcludeFoldersFromDeployment=node_modules</DockerExtraPublishProfileProperties>
```
amcasey: node_modules but not bower_components?

Nodejs/Product/TargetsVsix/Microsoft.NodejsToolsV2.targets
```diff
@@ -0,0 +1,313 @@
...
+  <Target Name="BuiltProjectOutputGroup"
+          Returns="@(BuiltProjectOutputGroupOutput)"
+          DependsOnTargets="$(BuiltProjectOutputGroupDependsOn)">
+    <ItemGroup>
+      <!-- We can't depend upon GeneratedJavascript defined by TypeScript because it 
+           doesn't occur when we're publishing, just when building.  So we calculate
+           the paths for the generated code here.  We also don't get to rely upon 
+           TypeScriptCollectPublishFiles to include the files for publishing because 
+           Worker Roles won't invoke PublishPipeline, they invoke some targets directly -->
+      <_GeneratedJavascript Include="@(TypeScriptCompile->'%(RelativeDir)%(Filename).js');@(Compile->'%(RelativeDir)%(Filename).js')"
```
amcasey: I'm surprised the file extension is hard-coded.  There's not a variable for that?  What about jsx files?

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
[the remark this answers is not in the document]
armanio123: Comment is outdated since the change is now backwards compatible.

Nodejs/Product/Nodejs/NodejsConstants.cs
armanio123: Yeah, I'll maker sure to clean up this constants if when they are not used anymore.

Nodejs/Product/Nodejs/Project/ImportWizard/ImportSettings.cs
armanio123: Only for SDK, if the user wants a specific version of TS they need to install it through NPM since we're using whatever version is installed in node_modules. e.g. `npm install --save-dev typescript@3.0.1`

Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureExpressApp/TypeScriptExpressApp.njsproj
armanio123: Good question, I tried it and it works. It also have the side effect of always building the project regardless if any file has been modified or not. I didn't detect any other side effect on the couple of tests I executed. I will change the templates.

Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptConsoleApp/package.json
armanio123: `start` serves no purpose for any project as far as I can see. My best guess is that it might have been added for express apps long ago as part of the "express" template.

Nodejs/Product/Nodejs/ProjectTemplates/VuejsApp/VuejsApp.njsproj
armanio123: It does, but a change on the njsproj is required. Commit sent.

Nodejs/Product/TargetsVsix/Microsoft.NodejsToolsV2.targets
armanio123: This is copied from the NodejsTools.targets file. I'm not very familiar with bower it might be necessary to copy this in deployment.

Nodejs/Product/TargetsVsix/Microsoft.NodejsToolsV2.targets
armanio123: There's no variable. `jsx` and `tsx` files are transpiled to `js` eventually. Note that this will be configured by `tsconfig.json`, meaning that `jsx` will not be transpiled unless `AllowJs=true` in the `tsconfig.json` file.

Nodejs/Product/TargetsVsix/Microsoft.NodejsToolsV2.targets
armanio123: Fixed.

Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureExpressApp/TypeScriptExpressApp.njsproj
armanio123: Change of plans, Compile no longer decides if the file is "dirty". Instead, we removed that logic and make the project "dirty" all the time, to allow either the SDK or tsc to build it.

Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureNodejsWorker/Worker.njsproj
armanio123: This is no longer the case. That logic got removed and now the ts file are always "dirty". The TS SDK and tsc are now able to implement its own incremental build.

Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureExpressApp/TypeScriptExpressApp.njsproj
```diff
@@ -31,10 +28,10 @@
...
   <ItemGroup>
-    <TypeScriptCompile Include="app.ts" />
-    <TypeScriptCompile Include="routes\index.ts" />
-    <TypeScriptCompile Include="routes\user.ts" />
-    <Content Include="tsconfig.json" />
+    <None Include="app.ts" />
+    <None Include="routes\index.ts" />
+    <None Include="routes\user.ts" />
+    <None Include="tsconfig.json" />
     <Content Include="package.json" />
```
amcasey: >Content [](start = 5, length = 7)

`None`?

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
```diff
@@ -161,10 +161,12 @@ internal override string GetItemType(string filename)
                 return node.ItemNode.ItemTypeName;
             }
 
+            // We need to return TypeScriptCompile for now to maintain backwards compatibility. In the future we will return "None" once the TypeScript SDK has been removed.
```
amcasey: >// We need to return TypeScriptCompile for now to maintain backwards compatibility. In the future we will return "None" once the TypeScript SDK has been removed. [](start = 12, length = 161)

Would it be worthwhile to file an issue for this so that you can locate all the places in need of updating by searching for the issue number?

Nodejs/Product/TargetsVsix/Microsoft.NodejsToolsV2.targets
amcasey: What about `%(Extension)`?

---
In reply to: [249956829](https://github.com/Microsoft/nodejstools/pull/2123#discussion_r249956829) [](ancestors = 249956829)

Nodejs/Product/TargetsVsix/Microsoft.NodejsToolsV2.targets
```diff
@@ -0,0 +1,312 @@
...
+    DependsOnTargets="$(GetReferenceAssemblyPathsDependsOn)">
+    <!-- We have no reference assembly paths -->
+  </Target>
+
+  <ItemGroup>
+    <NodeModulesContent Include="$(ProjectHome)\node_modules\**\*.*" />
+    <TypeScriptSource Exclude="$(ProjectHome)\node_modules\**\*.*;$(ProjectHome)\obj\**\*.*;$(ProjectHome)\bin\**\*.*;$(ProjectHome)\**\*.d.ts" Include="$(ProjectHome)\**\*.ts;$(ProjectHome)\**\*.tsx;$(ProjectHome)\**\*.jsx;" />
+  </ItemGroup>
+
+  <Target Name="CopyFilesToOutputDirectory"/>
```
amcasey: ><Target Name="CopyFilesToOutputDirectory"/> [](start = 2, length = 43)

I thought we still wanted to copy JS files to the output directory.  Does that happen through a different mechanism?  Or do we consider having a separate output directory redundant?

Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureExpressApp/TypeScriptExpressApp.njsproj
armanio123: I'm inclined to say it's better to use `None` for `package.son` and `readme.md`. I'll track this on issue #2128

Nodejs/Product/Nodejs/Project/NodejsProjectNode.cs
armanio123: The issue most likely is going to be there for a long time since there is no plans on the short term to remove the backward compatibility. I prefer to track independently during the effort to remove the SDK.

Nodejs/Product/TargetsVsix/Microsoft.NodejsToolsV2.targets
armanio123: If the target is not overridden VS tries to copy the ProjectnName.exe from obj. This scenario is not useful for Node projects so it got completely removed.

The output is configured by `tsconfig.json`.

Nodejs/Product/TargetsVsix/Microsoft.NodejsTools.targets
minestarks: Either way sounds fine to me.

Is there a way of localizing these strings? Are they English only?

armanio123: A quick search didn't throw any easy way of localizing. I can merge for now as is and follow up on a different issue for localizing the targets file if you're ok with it.

PR #2132, opened 2019-02-12
Delete some unused externs

amcasey:
Security audits will be easier without them.


amcasey:
@armanio123 I could be persuaded to leave some of the unused constants as documentation / for completeness.  Thoughts?

armanio123:
I prefer to remove all of the code that's unnecessary as it adds more unnecessary complexity and maintenance.

PR #2144, opened 2019-03-29
Removed URI from CommonUtils and used as Path instead.

armanio123:
Refactored CommonUtils for removing URI in paths and used Path instead.

Fixes an issue that blocks the user to save file and directory names with some Chinese characters.

### DIFF
--- a/Common/Product/SharedProject/CommonUtils.cs
+++ b/Common/Product/SharedProject/CommonUtils.cs
@@ -83,8 +83,7 @@ namespace Microsoft.VisualStudioTools
         public static string NormalizeDirectoryPath(string path)
         {
             var normalizedPath = NormalizePath(path);
-
-            return string.IsNullOrEmpty(normalizedPath) || HasEndSeparator(normalizedPath) ? normalizedPath : normalizedPath + Path.DirectorySeparatorChar;
+            return EnsureEndSeparator(normalizedPath);
         }
 
         /// <summary>
@@ -166,10 +165,7 @@ namespace Microsoft.VisualStudioTools
         public static string GetAbsoluteDirectoryPath(string root, string relativePath)
         {
             var absolutePath = GetAbsoluteFilePath(root, relativePath);
-
-            return string.IsNullOrEmpty(absolutePath) || HasEndSeparator(absolutePath)
-                ? absolutePath
-                : absolutePath + Path.DirectorySeparatorChar;
+            return EnsureEndSeparator(absolutePath);
         }
 
         /// <summary>
@@ -185,7 +181,7 @@ namespace Microsoft.VisualStudioTools
                 : Path.Combine(root, relativePath);
 
             var split = absolutePath.Split(DirectorySeparators);
-            var segments = new List<string>();
+            var segments = new LinkedList<string>();
 
             for (var i = split.Length - 1; i >= 0; i--)
             {
@@ -197,7 +193,7 @@ namespace Microsoft.VisualStudioTools
                 }
                 else if(segment != ".")
                 {
-                    segments.Insert(0, segment);
+                    segments.AddFirst(segment);
                 }
             }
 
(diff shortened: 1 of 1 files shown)


Common/Product/SharedProject/CommonUtils.cs
```diff
@@ -320,39 +227,31 @@ public static string GetRelativeDirectoryPath(string fromDirectory, string toDir
...
-                {
-                    relPath = relUri.IsFile ? relUri.LocalPath : relUri.AbsoluteUri;
-                }
-                else
-                {
-                    relPath = Uri.UnescapeDataString(relUri.ToString());
-                }
-            }
-            catch (InvalidOperationException ex)
+            var minIndex = Math.Min(splitDirectory.Length, splitFile.Length);
```
amcasey: What about "minLength" since it's not a valid index?

Common/Product/SharedProject/CommonUtils.cs
```diff
@@ -320,39 +227,31 @@ public static string GetRelativeDirectoryPath(string fromDirectory, string toDir
         /// </summary>
         public static string GetRelativeFilePath(string fromDirectory, string toFile)
```
amcasey: Is this going to do something sensible if the drive letters don't match?

Common/Product/SharedProject/CommonUtils.cs
```diff
@@ -320,39 +227,31 @@ public static string GetRelativeDirectoryPath(string fromDirectory, string toDir
...
         {
-            var fromUri = MakeUri(fromDirectory, true, UriKind.Absolute, "fromDirectory");
-            var toUri = MakeUri(toFile, false, UriKind.Absolute, "toFile");
+            var splitDirectory = Path.GetFullPath(TrimEndSeparator(fromDirectory)).Split(DirectorySeparators);
+            var splitFile = Path.GetFullPath(toFile).Split(DirectorySeparators);
```
amcasey: Consider adding a comment about why it's okay to include the file name in the path part comparison.  (I agree that it is, it just wasn't obvious why.)

Common/Product/SharedProject/CommonUtils.cs
```diff
@@ -240,28 +180,28 @@ public static string GetAbsoluteDirectoryPath(string root, string relativePath)
         /// either path is invalid.</exception>
         public static string GetAbsoluteFilePath(string root, string relativePath)
         {
-            var rootUri = MakeUri(root, true, UriKind.Absolute, "root");
-            var relUri = MakeUri(relativePath, false, UriKind.RelativeOrAbsolute, "relativePath");
+            var absolutePath = HasStartSeparator(relativePath)
```
amcasey: If it starts with a separator, it's not relative, is it?

Common/Product/SharedProject/CommonUtils.cs
```diff
@@ -240,28 +180,28 @@ public static string GetAbsoluteDirectoryPath(string root, string relativePath)
...
+                if (segment == "..")
                 {
-                    absUri = new Uri(rootUri, relUri);
+                    i--;
                 }
-                catch (UriFormatException ex)
+                else if(segment != ".")
                 {
-                    throw new InvalidOperationException("Cannot create absolute path", ex);
+                    segments.Insert(0, segment);
```
amcasey: Consider either switching to a linked list or building the list backwards and then reversing it.

Common/Product/SharedProject/CommonUtils.cs
armanio123: It's not, its an absolute path from the root of the current drive. That's why it needs to check for at the start.

Common/Product/SharedProject/CommonUtils.cs
armanio123: Good catch, I wasn't aware that the Insert is an O(n), I didn't expect that. LinkedList makes more sense. Fixed.

armanio123:
Yes, I created a bunch of tests for comparing the before and after. But for reasons I still don't understand I wasn't able to create a UnitTest project, so I manage to do it as a Console app. I don't like it that much but did the job. 

Attached the test results
[TestResult.txt](https://github.com/Microsoft/nodejstools/files/3031287/TestResult.txt)

PR #2153, opened 2019-04-11
Added NotificationTextBox

armanio123:
Adds the capability for screen readers to read that an npm search has finished and the amount of results found.

There's a drawback I'm still figuring out. When VS is unable to query npm, it keeps searching infinitely. A screen reader wont tell a user that a search as started only that a search has finished.

### DIFF
--- a/Nodejs/Product/Nodejs/NpmUI/NpmInstallWindowResources.resx
+++ b/Nodejs/Product/Nodejs/NpmUI/NpmInstallWindowResources.resx
@@ -192,4 +192,11 @@
   <data name="LatestVersion" xml:space="preserve">
     <value>(latest)</value>
   </data>
+  <data name="OneResultFoundMessage" xml:space="preserve">
+    <value>One result found</value>
+  </data>
+  <data name="ResultsFoundMessage" xml:space="preserve">
+    <value>{0} results found</value>
+    <comment>Parameter is a number representing the amount of results.</comment>
+  </data>
 </root>
--- a/Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
+++ b/Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
@@ -53,12 +53,14 @@ namespace Microsoft.NodejsTools.NpmUI
         private readonly Timer filterTimer;
         private readonly Dispatcher dispatcher;
         private readonly NpmWorker npmWorker;
+        private readonly Action<int> searchResultCallback;
 
         private bool disposed = false;
 
         public NpmPackageInstallViewModel(
             NpmWorker npmWorker,
-            Dispatcher dispatcher
+            Dispatcher dispatcher,
+            Action<int> searchResultCallback
         )
         {
             this.dispatcher = dispatcher;
@@ -67,6 +69,7 @@ namespace Microsoft.NodejsTools.NpmUI
             this.npmWorker.CommandStarted += this.NpmWorker_CommandStarted;
             this.npmWorker.CommandCompleted += this.NpmWorker_CommandCompleted;
             this.filterTimer = new Timer(this.FilterTimer_Elapsed, null, Timeout.Infinite, Timeout.Infinite);
+            this.searchResultCallback = searchResultCallback;
         }
 
         private void NpmWorker_CommandStarted(object sender, EventArgs e)
@@ -243,6 +246,11 @@ namespace Microsoft.NodejsTools.NpmUI
                 if (filterText != GetTrimmedTextSafe(this.filterText))
                 {
                     return;
+                }
+
+                if (!string.IsNullOrWhiteSpace(filterText))
+                {
+                    this.searchResultCallback(filtered.Count());
                 }
 
                 var originalSelectedPackage = this.SelectedPackage;
--- a/Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
+++ b/Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
@@ -53,14 +53,14 @@ namespace Microsoft.NodejsTools.NpmUI
         private readonly Timer filterTimer;
         private readonly Dispatcher dispatcher;
         private readonly NpmWorker npmWorker;
-        private readonly Action<int> searchResultCallback;
 
         private bool disposed = false;
 
+        public event EventHandler<int> OnSearchResultEnded;
+
         public NpmPackageInstallViewModel(
             NpmWorker npmWorker,
-            Dispatcher dispatcher,
-            Action<int> searchResultCallback
+            Dispatcher dispatcher
         )
         {
             this.dispatcher = dispatcher;
@@ -69,7 +69,6 @@ namespace Microsoft.NodejsTools.NpmUI
             this.npmWorker.CommandStarted += this.NpmWorker_CommandStarted;
             this.npmWorker.CommandCompleted += this.NpmWorker_CommandCompleted;
             this.filterTimer = new Timer(this.FilterTimer_Elapsed, null, Timeout.Infinite, Timeout.Infinite);
-            this.searchResultCallback = searchResultCallback;
         }
 
         private void NpmWorker_CommandStarted(object sender, EventArgs e)
@@ -250,7 +249,7 @@ namespace Microsoft.NodejsTools.NpmUI
 
                 if (!string.IsNullOrWhiteSpace(filterText))
                 {
-                    this.searchResultCallback(filtered.Count());
+                    this.OnSearchResultEnded(this, filtered.Count());
                 }
 
                 var originalSelectedPackage = this.SelectedPackage;
--- a/Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
@@ -12,22 +12,6 @@ namespace Microsoft.NodejsTools.SharedProject.Wpf
         // This control's AutomationPeer is the object that actually raises the UIA Notification event.
         private NotificationTextBoxAutomationPeer notificationTextBoxAutomationPeer;
 
-        // Assume the UIA Notification event is available until we learn otherwise.
-        // If we learn that the UIA Notification event is not available, no instance
-        // of the NotificationTextBox should attempt to raise it.
-        private static bool notificationEventAvailable = true;
-        public bool NotificationEventAvailable
-        {
-            get
-            {
-                return notificationEventAvailable;
-            }
-            set
-            {
-                notificationEventAvailable = value;
-            }
-        }
-
         protected override AutomationPeer OnCreateAutomationPeer()
         {
             this.notificationTextBoxAutomationPeer = new NotificationTextBoxAutomationPeer(this);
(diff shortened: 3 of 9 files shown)


Nodejs/Product/Nodejs/NpmUI/NpmInstallWindowResources.resx
```diff
@@ -192,4 +192,11 @@
   <data name="LatestVersion" xml:space="preserve">
     <value>(latest)</value>
   </data>
+  <data name="OneResultFoundMessage" xml:space="preserve">
+    <value>One result found</value>
```
amcasey: This is pretty nitpicky, but I'm not sure special-casing one localizes well.  In particular, I believe some languages inflect the counted noun differently depending on the quantity (e.g. one, two, many) and would require additional special-cases for strict correctness.  However, I'm probably overthinking things.  If this is the usual pattern, then I'm sure it's fine.

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
```diff
@@ -245,6 +248,11 @@ private async void FilterTimer_Elapsed(object state)
                     return;
                 }
 
+                if (!string.IsNullOrWhiteSpace(filterText))
+                {
+                    this.searchResultCallback(filtered.Count());
```
amcasey: Can `searchResultCallback` be null?

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
```diff
@@ -53,12 +53,14 @@ internal static class FilterState
...
         private readonly NpmWorker npmWorker;
+        private readonly Action<int> searchResultCallback;
 
         private bool disposed = false;
 
         public NpmPackageInstallViewModel(
             NpmWorker npmWorker,
-            Dispatcher dispatcher
+            Dispatcher dispatcher,
+            Action<int> searchResultCallback
```
amcasey: It would be slightly more idiomatic to expose this as an event.

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
```diff
@@ -0,0 +1,103 @@
...
+    internal class NotificationTextBox : TextBox
+    {
+        // This control's AutomationPeer is the object that actually raises the UIA Notification event.
+        private NotificationTextBoxAutomationPeer notificationTextBoxAutomationPeer;
+
+        // Assume the UIA Notification event is available until we learn otherwise.
+        // If we learn that the UIA Notification event is not available, no instance
+        // of the NotificationTextBox should attempt to raise it.
+        private static bool notificationEventAvailable = true;
+        public bool NotificationEventAvailable
```
amcasey: I believe auto-properties can have initializers now: `public bool NotificationEventAvailable { get; set; } = true;`

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            this.notificationTextBoxAutomationPeer = new NotificationTextBoxAutomationPeer(this);
+
+            return this.notificationTextBoxAutomationPeer;
+        }
+
+        public void RaiseNotificationEvent(string notificationText, string notificationGuid)
+        {
+            // Only attempt to raise the event if we already have an AutomationPeer.
+            if (this.notificationTextBoxAutomationPeer != null)
```
amcasey: Is any synchronization required here?

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
```diff
@@ -0,0 +1,103 @@
...
+            {
+                this.notificationTextBoxAutomationPeer.RaiseNotificationEvent(notificationText, notificationGuid);
+            }
+        }
+    }
+
+    internal class NotificationTextBoxAutomationPeer : TextBoxAutomationPeer
+    {
+        private NotificationTextBox notificationTextBox;
```
amcasey: `readonly`

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
```diff
@@ -0,0 +1,103 @@
...
+            if (this.notificationTextBox.NotificationEventAvailable)
+            {
+                // If no UIA clients are listening for events, don't bother raising one.
+                if (UiaAutomationNativeMethods.UiaClientsAreListening())
+                {
+                    // Get the IRawElementProviderSimple for this AutomationPeer if we don't
+                    // have it already.
+                    if (this.rawElementProviderSimple == null)
+                    {
+                        var automationPeer = FromElement(this.notificationTextBox);
```
amcasey: I'm confused - isn't `this` the automation peer of `this.notificationTextBox`?

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
```diff
@@ -0,0 +1,103 @@
...
+        public NotificationTextBoxAutomationPeer(NotificationTextBox owner) : base(owner)
+        {
+            this.notificationTextBox = owner;
+        }
+
+        public void RaiseNotificationEvent(string notificationText, string notificationGuid)
+        {
+            // If we already know that the UIA Notification event is not available, do not
+            // attempt to raise it.
+            if (this.notificationTextBox.NotificationEventAvailable)
```
amcasey: Why is this stored on the textbox if it is only read and written by the peer?

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
```diff
@@ -0,0 +1,103 @@
...
+                                UiaAutomationNativeMethods.AutomationNotificationKind.ActionCompleted,
+                                UiaAutomationNativeMethods.AutomationNotificationProcessing.All,
+                                notificationText,
+                                notificationGuid);
+                        }
+                        catch (EntryPointNotFoundException)
+                        {
+                            // The UIA Notification event is not not available, so don't attempt
+                            // to raise it again.
+                            this.notificationTextBox.NotificationEventAvailable = false;
```
amcasey: Will this happen if narrator is inactive?  What if it's activated later in the same session?

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
```diff
@@ -0,0 +1,103 @@
...
+                if (UiaAutomationNativeMethods.UiaClientsAreListening())
+                {
+                    // Get the IRawElementProviderSimple for this AutomationPeer if we don't
+                    // have it already.
+                    if (this.rawElementProviderSimple == null)
+                    {
+                        var automationPeer = FromElement(this.notificationTextBox);
+                        if (automationPeer != null)
+                        {
+                            this.rawElementProviderSimple = this.ProviderFromPeer(automationPeer);
```
amcasey: Is this expensive?  Why cache it?

Nodejs/Product/Nodejs/NpmUI/NpmInstallWindowResources.resx
armanio123: I'm not sure if this is the pattern. I would follow up with the loc team and figure out the best approach.

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
armanio123: It cannot.

Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallViewModel.cs
armanio123: Changed.

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
armanio123: Changed.

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
armanio123: I don't think so but I might be missing something. Is there a specific scenario you think might affect this?

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
armanio123: Done.

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
armanio123: I'm not entirely sure why this is needed. Nonetheless, this was exactly the same from the documentation found regarding the notifications. I'll suggest to leave it as is. Let me know if you think otherwise.

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
amcasey: Is there some reason to believe the calls to `OnCreateAutomationPeer` and `RaiseNotificationEvent` won't overlap?

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
armanio123: Only for the intention of exposing it on the TextBox element. I'll could move it to the peer if you feel its necessary.

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
armanio123: No, I believe this should happen when clr is unable to load the UIAutomationCore.dll because the method is not found. It's a fail safe.

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
armanio123: Looking through the source it does appears that in order to get the provider it might navigate the UIA tree, which I'm not entirely sure what it means but sounds expensive.

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
amcasey: Oh, it's used in XAML too?  Good answer.  Just leave it. :)

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
amcasey: Can you please add a comment that you're following a template so no one "cleans it up" later?

Nodejs/Product/Nodejs/SharedProject/Wpf/NotificationTextBox.cs
amcasey: Offline you pointed out that it doesn't actually matter if they overlap.

PR #2159, opened 2019-04-24
Migrates all references/dependencies to Nuget

armanio123:
- Moves all references "possible" to Nuget. Also updates them to the latest version available.
- Removes unused references.
- Updates .Net framework to 4.7.2

Note that this change will make the project unsupported on Visual Studio 2017, version 2019 will be required to open the solution.

### DIFF
--- a/Build/Common.Build.CSharp.settings
+++ b/Build/Common.Build.CSharp.settings
@@ -29,7 +29,7 @@
   </PropertyGroup>
   
   <PropertyGroup>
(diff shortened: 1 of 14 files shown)


armanio123:
@amcasey I'm going to merge this as it's required for the build to succeed. Nonetheless, I still recommend to take a look and I'll address any feedback on a different pull request.

amcasey:
Pulling the VS references from nuget seems like a great change, but I'm not sure I see the benefit in pulling the BCL references from their as well.  Also, why do so many of our projects consume `Microsoft.CSharp`?

PR #2173, opened 2019-05-22
Fix .Net Core unit tests

armanio123:
Fixes the issue of js tests not being discovered/executed on .Net Core projects using Visual Studio's Test Explorer.

Also, adds Jest support to .Net Core JS unit tests.

### DIFF
--- a/Common/Product/SharedProject/CommonUtils.cs
+++ b/Common/Product/SharedProject/CommonUtils.cs
@@ -201,26 +201,6 @@ namespace Microsoft.VisualStudioTools
         }
 
         /// <summary>
-        /// Returns a normalized file path created by joining relativePath to root.
-        /// The result is not guaranteed to end with a backslash.
-        /// </summary>
-        /// <returns>True, if the absolute path is returned successfully; otherwise false, 
-        /// if root is not an absolute path, or either path is invalid.</returns>
-        public static bool TryGetAbsoluteFilePath(string root, string relativePath, out string absoluteFilePath)
-        {
-            try
-            {
-                absoluteFilePath = GetAbsoluteFilePath(root, relativePath);
-                return true;
-            }
-            catch
-            {
-                absoluteFilePath = null;
-                return false;
-            }
-        }
-
-        /// <summary>
         /// Returns a relative path from the base path to the other path. This is
         /// intended for serialization rather than UI. See CreateFriendlyDirectoryPath
         /// for UI strings.
--- a/Common/Product/SharedProject/CommonUtils.cs
+++ b/Common/Product/SharedProject/CommonUtils.cs
@@ -197,14 +197,14 @@ namespace Microsoft.VisualStudioTools
                 }
             }
 
-            return string.Join(Path.DirectorySeparatorChar.ToString(), segments);
+            return Path.Combine(segments.ToArray());
         }
 
         /// <summary>
         /// Returns a normalized file path created by joining relativePath to root.
         /// The result is not guaranteed to end with a backslash.
         /// </summary>
-        /// <returns>True, if the absolut path is returned successfully; otherwise false, 
+        /// <returns>True, if the absolute path is returned successfully; otherwise false, 
         /// if root is not an absolute path, or either path is invalid.</returns>
         public static bool TryGetAbsoluteFilePath(string root, string relativePath, out string absoluteFilePath)
         {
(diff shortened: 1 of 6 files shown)


Common/Product/SharedProject/CommonUtils.cs
```diff
@@ -200,6 +200,26 @@ public static string GetAbsoluteFilePath(string root, string relativePath)
             return string.Join(Path.DirectorySeparatorChar.ToString(), segments);
```
amcasey: >string.Join(Path.DirectorySeparatorChar.ToString(), segments) [](start = 19, length = 61)

`Path.Combine(segments)`?

Common/Product/SharedProject/CommonUtils.cs
```diff
@@ -200,6 +200,26 @@ public static string GetAbsoluteFilePath(string root, string relativePath)
...
+        /// <returns>True, if the absolut path is returned successfully; otherwise false, 
+        /// if root is not an absolute path, or either path is invalid.</returns>
+        public static bool TryGetAbsoluteFilePath(string root, string relativePath, out string absoluteFilePath)
+        {
+            try
+            {
+                absoluteFilePath = GetAbsoluteFilePath(root, relativePath);
+                return true;
+            }
+            catch
```
amcasey: >catch [](start = 12, length = 5)

Do we want to catch exceptions other than `ArgumentException`?

amcasey:
>                 throw new InvalidOperationException($"Called from unexepected assembly: '{currentAssembly.FullName}'.");

At some point, we should review these and add telemetry.

---
Refers to: Nodejs/Product/Nodejs/TestFrameworks/TestFrameworkDirectories.cs:82 in 1a3cb0e. [](commit_id = 1a3cb0e07644535153629d7afbd3e6d8be889fe4, deletion_comment = False)

Common/Product/SharedProject/CommonUtils.cs
```diff
@@ -200,6 +200,26 @@ public static string GetAbsoluteFilePath(string root, string relativePath)
             return string.Join(Path.DirectorySeparatorChar.ToString(), segments);
         }
 
+        /// <summary>
+        /// Returns a normalized file path created by joining relativePath to root.
+        /// The result is not guaranteed to end with a backslash.
+        /// </summary>
+        /// <returns>True, if the absolut path is returned successfully; otherwise false, 
+        /// if root is not an absolute path, or either path is invalid.</returns>
+        public static bool TryGetAbsoluteFilePath(string root, string relativePath, out string absoluteFilePath)
```
amcasey: >TryGetAbsoluteFilePath [](start = 27, length = 22)

This feels specialized for the only caller, so I'd consider inlining it.

Common/Product/SharedProject/CommonUtils.cs
armanio123: Makes sense to me. Changed.

Common/Product/SharedProject/CommonUtils.cs
armanio123: So far we only use it for `ArgumentException` as I have not seen it in another scenario. Nonetheless, there are multiple reasons that can make the path fail with different exceptions. These are just a couple of examples :
```
System.Security.SecurityException:
The caller does not have the required permissions.

System.ArgumentNullException:
path is null.

System.NotSupportedException:
path contains a colon (":") that is not part of a volume identifier (for example,
"c:\").

System.IO.PathTooLongException:
The specified path, file name, or both exceed the system-defined maximum length.
For example, on Windows-based platforms, paths must be less than 248 characters,
and file names must be less than 260 characters.
```

I find it's easier and more readable to just catch any exception and return `false`.

Let me know if you think otherwise.

Common/Product/SharedProject/CommonUtils.cs
armanio123: Although I agree is only used for this scenario, I feel like adding it to this class is a good practice to as it shares the same responsibility than the rest of the methods on it as well as making the code readable and cleaner. 

If you still have preference for inlining it, I'm happy to change it.

armanio123:
I agree on the telemetry. Actually, I think we should invest some time to analyze, identify important telemetry and implement it.

Common/Product/SharedProject/CommonUtils.cs
armanio123: Changed as per our offline comments.

Common/Product/SharedProject/CommonUtils.cs
armanio123: Changed as per our offline comments.

PR #2175, opened 2019-05-31
Rollback change on GetAbsoluteFilePath

armanio123:
Fixes the project creation by using string.Join instead of Path.Combine.


amcasey:
https://github.com/dotnet/coreclr/blob/fb86c0294a999b2c7bd1e13da1fdc0d3c2f701e5/src/mscorlib/shared/System/IO/Path.cs#L311 :(

PR #2183, opened 2019-06-15
Change all template files to encoding to UTF8 without BOM

armanio123:
Change all template *.js and *.ts files from Western European encoding to UTF8 with BOM.

Note that not all of the files required change as some of them where already encoded to UTF8 with BOM.


amcasey:
Is the BOM necessary?  It causes problems for some consumers.

armanio123:
It might not be necessary, I'll see to encoded it without it. 

Do you know what kind of problems can it cause?

amcasey:
It's not uncommon to find a parser or lexer that reports a syntax error on the "unexpected" characters or, worse, gets misaligned and mishandles the rest of the file.

PR #2187, opened 2019-07-09
Only end the build if it started successfully

### LINKED ISSUE #2117: Don't end build if it failed to start
https://github.com/Microsoft/nodejstools/blob/2118e8ff4dcba7cfd281f8e9d12b62cc0d96ff0f/Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs#L2625

If a build is in-progress, we throw an `InvalidOperationException`.  There's nothing inherently wrong with this, but we then try to end the build that never started (in the following finally block) and that call hangs forever since no build is in progress.

If we want to continue throwing an exception, the obvious workaround would be to add a flag tracking whether or not the build was actually started and only end it if the flag is set.

Repro: Open TypeScript-VisualStudio.sln with a VS build from the staging branch.

amcasey:
In the (rare) case that a design time build failed to start, we would
attempt to release the UI thread in `TryBeginBuild` and then again in
`EndBuild`.  The comments suggested that it was the `TryBeginBuild` that
was intended to do the cleanup and code inspection confirmed that
`EndBuild` did not perform useful work if `TryBeginBuild` failed, so
this change gives `TryBeginBuild` responsibility for cleanup and doesn't
call `EndBuild` a build was successfully started.

Fixes #2117 

Tested by manually inducing `TryBeginBuild` failures in the debugger.

### DIFF
--- a/Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
+++ b/Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
@@ -2614,20 +2614,21 @@ namespace Microsoft.VisualStudioTools.Project
             this.Site.GetUIThread().MustBeCalledFromUIThread();
 
             var result = MSBuildResult.Failed;
-            const bool designTime = true;
 
-            BuildSubmission submission = null;
-
-            try
+            // Do the actual Build
+            if (this.buildProject != null)
             {
-                // Do the actual Build
-                if (this.buildProject != null)
-                {
-                    if (!TryBeginBuild(designTime, true))
-                    {
-                        throw new InvalidOperationException("A build is already in progress.");
-                    }
+                const bool designTime = true;
+                const bool requiresUIThread = true;
 
+                if (!TryBeginBuild(designTime, requiresUIThread))
+                {
+                    throw new InvalidOperationException("A build is already in progress.");
+                }
+
+                BuildSubmission submission = null;
+                try
+                {
                     var targetsToBuild = new string[target != null ? 1 : 0];
                     if (target != null)
                     {
@@ -2647,10 +2648,10 @@ namespace Microsoft.VisualStudioTools.Project
 
                     result = (buildResult.OverallResult == BuildResultCode.Success) ? MSBuildResult.Successful : MSBuildResult.Failed;
                 }
-            }
-            finally
-            {
-                EndBuild(submission, designTime, true);
+                finally
+                {
+                    EndBuild(submission, designTime, requiresUIThread);
+                }
             }
 
             return result;
@@ -2666,8 +2667,9 @@ namespace Microsoft.VisualStudioTools.Project
         protected virtual BuildSubmission DoAsyncMSBuildSubmission(string target, Action<MSBuildResult, string> uiThreadCallback)
         {
             const bool designTime = false;
+            const bool requiresUIThread = false;
 
-            if (!TryBeginBuild(designTime, false))
+            if (!TryBeginBuild(designTime, requiresUIThread))
             {
                 if (uiThreadCallback != null)
                 {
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
```diff
@@ -2614,20 +2614,21 @@ protected virtual MSBuildResult InvokeMsBuild(string target)
...
-            try
+            // Do the actual Build
+            if (this.buildProject != null)
             {
-                // Do the actual Build
-                if (this.buildProject != null)
+                const bool designTime = true;
+                const bool requiresUIThread = true;
+
+                if (!TryBeginBuild(designTime, requiresUIThread))
```
armanio123: Either `designTime` or `requiresUIThread` is not required. The two places where `TryBeginBuild` is called, have the same value for both variables. We can remove one and refactor the method to make it a little bit more readable. I suggest removing `designTime`, as `requiresUIThread` is easy to understand its purpose.

Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
```diff
@@ -2647,10 +2648,10 @@ protected virtual MSBuildResult InvokeMsBuild(string target)
 
                     result = (buildResult.OverallResult == BuildResultCode.Success) ? MSBuildResult.Successful : MSBuildResult.Failed;
                 }
-            }
-            finally
-            {
-                EndBuild(submission, designTime, true);
+                finally
+                {
+                    EndBuild(submission, designTime, requiresUIThread);
```
armanio123: Same comment as about `designTime` and `requiresUIThread`.

Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
```diff
@@ -6026,57 +6028,44 @@ private void SetupProjectGlobalPropertiesThatAllProjectSystemsMustSet()
...
-            try
+            // If the SVsBuildManagerAccessor service is absent, we're not running within Visual Studio.
+            if (this.buildManagerAccessor != null)
             {
-                // If the SVsBuildManagerAccessor service is absent, we're not running within Visual Studio.
-                if (this.buildManagerAccessor != null)
+                if (requiresUIThread)
                 {
-                    if (requiresUIThread)
+                    var result = this.buildManagerAccessor.ClaimUIThreadForBuild();
```
armanio123: Can `ClaimUIThreadForBuild` lock the UI thread and then throw and exception?

Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
```diff
@@ -6026,57 +6028,44 @@ private void SetupProjectGlobalPropertiesThatAllProjectSystemsMustSet()
...
-                        releaseUIThread = true; // assume we need to release this immediately until we get through the whole gauntlet.
+                        // Not allowed to claim the UI thread right now. Try again later.
+                        return false;
                     }
+                }
 
-                    if (designTime)
+                if (designTime)
+                {
+                    var result = this.buildManagerAccessor.BeginDesignTimeBuild();
```
armanio123: Can this throw an exception? If so, I think the UI thread will be blocked and never released.

Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
amcasey: I'd prefer to retain `designTime` since one could reasonably conclude that `designTime` requires the UI thread, but I don't see how would deduce that `requiresUIThread` implies a design time build.

Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
amcasey: I'm not sure what you mean by "lock the UI thread".  It sets a flag indicating that it's going to use the UI thread and that others checking the flag should not.

I would also not expect it to throw since it returns an HRESULT.  It can still throw OOM, etc, but such exceptions are unrecoverable anyway.

Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
amcasey: As above, I believe it's technically possible, but I'm not convinced we should attempt to recover if it does.

Nodejs/Product/Nodejs/SharedProject/ProjectNode.cs
amcasey: I'll do this in a new PR.

Issue #2129, opened 2019-01-29
Update npm Package option does not update dev dependencies


Mikejo5000:
##### Expected Behavior 
1. Update a npm package version in package.json devDependencies
2. Right-click on the package in Solution Explorer, choose Update npm Package.
3. Package gets updated

##### Actual Behavior 
Package does not get updated

- NTVS Version: 1.5.1228.1
- Visual Studio Version: Dev16
- Node.js Version: 10.13.0

##### Steps to Reproduce
1. Update a npm package version in package.json devDependencies
2. Right-click on the package in Solution Explorer, choose Update npm Package.
3. Check whether package version was updated.

amcasey:
@uniqueiniquity Can you please check whether there's a reason we're not doing this?  It seems intuitively reasonable.

rakatyal:
@Mikejo5000: I played around with this today and could repro the behavior, however I verified that npm itself behaves the same way so it may be that 'npm update' isn't supposed to be used this way. Kindly verify my investigations by these 2 workflows and let me know if you observe the same. 

1. If I do not update the package.json file (keep it same as the version installed) to the new version and then run 'Update npm package' from the solution explorer, it does update the package to the latest version and also update the 'package.json' file to reflect the same. If you look at the logs in the 'Output' window inside VS and filter it to show output from 'Npm', you can see what command VS executed for this action. In this case it should say 'npm update <package_name> --save'. 

2. If I try your workflow and update the package.json file with a newer version and then run the 'Update npm package' it doesn't do anything. But if you notice the output window again, you would notice that the command executed by npm is the same as (1). 

I believe the problem is that when you update the package.json file beforehand, npm fails to pick it as a package to be updated and hence nothing happens. The correct workflow here is to do this directly from the solution explorer *without* modifying package.json beforehand, which is what I did in (1). 

I also verified this by trying the 'npm outdated' command. It should list all the packages that can be updated in your project and basically tells you what will happen if you run 'npm update'. Try this command in workflows (1) and (2). You should notice that (1) will list you the package you wish to update but (2) would not. 

Please let me know what you find!

rakatyal:
I made the experience better in the referenced PR above. Closing this issue. Kindly re-open if you have further comments.

Mikejo5000:
Looks good. Thanks!

PR #2190, opened 2019-07-18
Do not disable the Install npm package menu option

rakatyal:
This PR is in relation to this issue:
https://github.com/microsoft/nodejstools/issues/2129

We disable the 'Install Missing npm package' in the VS right-click menu when the package is present. The only way to update to a newer version of the package inside VS for the customers is to use the 'Update npm package' menu option. 

However this option does not work if the user first updates the package.json file with the new package version and then tries the menu option (more details in the comment on the issue). This is just due to the fact how npm update behaves. 

Due to this, we shouldn't disable the 'Install Missing npm package' option to allow users to use this option in such cases. This PR also changes the menu option text to 'Install npm package' instead to reflect that this just runs the 'npm install' command.

### DIFF
--- a/Nodejs/Product/Nodejs/NodejsTools.vsct
+++ b/Nodejs/Product/Nodejs/NodejsTools.vsct
@@ -180,7 +180,7 @@
         <CommandFlag>DynamicVisibility</CommandFlag>
         <Strings>
           <CommandName>cmdidNpmInstallModules</CommandName>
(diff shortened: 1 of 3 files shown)


rakatyal:
@armanio123: Yes, the menu would say 'Install npm Package(s)' and it would run the 'npm install' command.

PR #2223, opened 2019-08-02
Added tags for Vue templates

armanio123:
Fixes Vue templates not appearing on the "New Project Dialog".

### DIFF
--- a/Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptVuejsApp/TypeScriptVuejsApp.vstemplate
+++ b/Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptVuejsApp/TypeScriptVuejsApp.vstemplate
@@ -12,6 +12,9 @@
     <NumberOfParentCategoriesToRollUp>1</NumberOfParentCategoriesToRollUp>
     <PromptForSaveOnCreation>true</PromptForSaveOnCreation>
     <TemplateGroupID>Node.js</TemplateGroupID>
+    <LanguageTag>Typescript</LanguageTag>
+    <PlatformTag>Windows</PlatformTag>
+    <ProjectTypeTag>Web</ProjectTypeTag>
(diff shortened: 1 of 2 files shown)


Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptVuejsApp/TypeScriptVuejsApp.vstemplate
```diff
@@ -12,6 +12,9 @@
     <NumberOfParentCategoriesToRollUp>1</NumberOfParentCategoriesToRollUp>
     <PromptForSaveOnCreation>true</PromptForSaveOnCreation>
     <TemplateGroupID>Node.js</TemplateGroupID>
+    <LanguageTag>Typescript</LanguageTag>
+    <PlatformTag>Windows</PlatformTag>
```
amcasey: Why Windows?  Seems like this would work anywhere?

armanio123: I agree but there's no option for 'any' on the templates and the others are misleading. Here's the documentation [Built-in tags](https://docs.microsoft.com/en-us/visualstudio/ide/template-tags?view=vs-2019#built-in-tags). The other templates are tagged as Windows as well.

amcasey: The example in the docs makes it look like you can tag multiple platforms.

armanio123: I suppose I can add Linux and MacOS as well as per Node.js support. I'll fix it in second.

PR #2224, opened 2019-08-03
Updated template tags

armanio123:
Update the project template tags to include more platforms.

### DIFF
--- a/Nodejs/Product/Nodejs/ProjectTemplates/Express4App/ExpressApp.vstemplate
+++ b/Nodejs/Product/Nodejs/ProjectTemplates/Express4App/ExpressApp.vstemplate
@@ -12,6 +12,11 @@
     <NumberOfParentCategoriesToRollUp>1</NumberOfParentCategoriesToRollUp>
     <PromptForSaveOnCreation>true</PromptForSaveOnCreation>
     <PreviewImage>Preview.png</PreviewImage>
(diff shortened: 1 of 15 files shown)


Nodejs/Product/Nodejs/ProjectTemplates/Express4App/ExpressApp.vstemplate
```diff
@@ -12,6 +12,11 @@
     <NumberOfParentCategoriesToRollUp>1</NumberOfParentCategoriesToRollUp>
     <PromptForSaveOnCreation>true</PromptForSaveOnCreation>
     <PreviewImage>Preview.png</PreviewImage>
+    <LanguageTag>Javascript</LanguageTag>
+    <PlatformTag>Windows</PlatformTag>
+    <PlatformTag>Linux</PlatformTag>
+    <PlatformTag>macOS</PlatformTag>
```
amcasey: I would have guessed "MacOS", but I'm assuming you copied this from somewhere.

armanio123: It's the what autocomplete shows. The [documentation](https://docs.microsoft.com/en-us/visualstudio/ide/template-tags?view=vs-2019#built-in-tags) shows this as all lowercase but I don't think it matters as it is displaying the tags correctly .

Issue #2127, opened 2019-01-23
Success mocha test marked Red


genusP:
I have some mocha unit tests that are successed and  marked as fail (red). When I run mocha from CLI, all test green.

##### Expected Behavior 
All tests marked green.
##### Actual Behavior 
Some test marked redl

- NTVS Version: 1.4.21001.1 Commit Hash:8dd15923800d931b153ab9e4de74e42a74eba5e6
- Visual Studio Version: 15.Release/15.9.5+28307.280
- Node.js Version: v8.11.4

##### Steps to Reproduce
1.  Clone repo https://github.com/genusP/ntvsIssue
2.  Press Run all button in Test explorer

jkager:
I see the same behavior on my machine, and I can reproduce it with @genusP's solution. All tests pass according to the mocha CLI output, but some are displayed as failed in the test runner:

![image](https://user-images.githubusercontent.com/46070227/52346026-f3d94a00-2a1e-11e9-8552-072e6019e999.png)

![image](https://user-images.githubusercontent.com/46070227/52344721-ca6aef00-2a1b-11e9-9fbf-325a837c4a44.png)

All tests are implemented identically:
![image](https://user-images.githubusercontent.com/46070227/52344832-19b11f80-2a1c-11e9-9b51-49227d54adbf.png)

The tests fail using the `vstest.console.exe` CLI too:

![image](https://user-images.githubusercontent.com/46070227/52345917-a0ff9280-2a1e-11e9-8c1e-590e1841beba.png)

And when calling the CLI with the /Diag-Flag, one more test fails:

![image](https://user-images.githubusercontent.com/46070227/52346115-2e42e700-2a1f-11e9-9c8e-3110b2e8a9d9.png)

Here is the output of the diagnostic log: [VsTestDiag.log](https://github.com/Microsoft/nodejstools/files/2836576/VsTestDiag.log)

My setup:
* Visual Studio Enterprise 2017, Version 15.9.6
* Node.js Tools 1.4.21001.1 Commit Hash:8dd15923800d931b153ab9e4de74e42a74eba5e6
* Node.js v10.15.1

commonsensesoftware:
I see the same thing happening. I'm using:

* VisualStudio.16.Release/16.0.1+28803.156
* TypeScript Tools   16.0.10306.2001
* Node.js Tools   1.5.10301.2 Commit Hash:9c34eb81d73db6babeb71674feb1cb437dba9f2f

[...]

PR #2225, opened 2019-08-22
Changed how unit tests shows on the Test Explorer

armanio123:
Changes how the tests are displayed on the Test Explorer. Now the tree will look like:
```
<folder\fileName>
    <Test Group and Subgroup> or <global scope>
        <Test name>
```

For example, the mocha test:
```js
// file: mocha\UnitTest1.js

var assert = require('assert');

it('Test A', function () {
    assert.ok(true, "This shouldn't fail");
});

describe('Test Suite 1', function () {
    it('Test A', function () {
        assert.ok(true, "This shouldn't fail");
    });

    describe('Sub Suite 1', function () {
        it('Test A', function () {
            assert.ok(true, "This shouldn't fail");
        });
    });
});
```
Will show on the test explorer:
image

This is standarized for all of the test frameworks supported.

Also fixes an issue when running tests with suites.

### DIFF
--- a/Nodejs/Product/TestAdapter/SerializationHelpers.cs
+++ b/Nodejs/Product/TestAdapter/SerializationHelpers.cs
@@ -27,7 +27,7 @@ internal sealed class ResultObject
 internal sealed class TestEvent
 {
     public string type { get; set; }
-    public string title { get; set; }
+    public string fullyQualifiedName { get; set; }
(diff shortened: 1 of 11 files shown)


ozyx:
is it fixing this issue? #2127

armanio123:
@ozyx Yes, this also fixes the issue #2127

PR #2230, opened 2019-09-18
SpaRoot msbuild property support

radiy:
Test discovery not working for the default react template - dotnet new react.
Test adapter assumes that node_modules folder will be in root of a project, but in case of the react template it in ClientApp folder defined by SpaRoot property.
This patch adds code for handling SpaRoot property and overriding work directory based on it.


joj:
Overridden by more recent changes. Closing.

PR #2247, opened 2019-11-20
Create workflow to lint templates

rachelgshaffer:
Sets up a workflow to perform linting on the template files

Uses [Run ESLint action](https://github.com/marketplace/actions/run-eslint)

I think that because this is coming from my fork that the `main.yml` file is not being recognized and the check isn't running. I have another PR https://github.com/microsoft/nodejstools/pull/2248 for setting up the file that does the initialization.


.github/workflows/main.yml
```diff
@@ -0,0 +1,13 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Lint template files
+      uses: "stefanoeb/eslint-action@master"
+      with:
+        args: "Nodejs/Product/Nodejs/ProjectTemplates/**/*.js Nodejs/Product/Nodejs/ProjectTemplates/**/*.ts Nodejs/Product/Nodejs/Templates/**/*.js Nodejs/Product/Nodejs/Templates/**/*.ts"
```
armanio123: Can we lint `tsx`, `jsx`, `vue`, `html` files as well to cover everything where js/ts might be present? Is it worth it?

.eslintrc
```diff
@@ -0,0 +1,164 @@
...
+  },
+  "env": {
+    "amd": true,
+    "browser": true,
+    "jquery": true,
+    "node": true,
+    "es6": true, // This enables ES6 global variables AND ES6 syntax
+    "worker": true
+  },
+  "rules": {
```
armanio123: Why not use the recommended ones?

.eslintrc
```diff
@@ -0,0 +1,164 @@
...
+    // The below are some, but not all, of the rules from eslint:recommended https://github.com/eslint/eslint/blob/master/conf/eslint-recommended.js (all errors set to warning)
```
armanio123: Why not report errors? will it break the build?

.github/workflows/main.yml
rachelgshaffer: Ah good catch for `tsx` and `jsx` - I'll fix that. but there isn't support for `vue` or `html` files.

.eslintrc
rachelgshaffer: Error here means the severity that is reported to the user

.eslintrc
rachelgshaffer: This is the default file we put on user's disk to be used when a local project `eslintrc` is not present. We don't use the recommended ones for a few reasons:
1. They all display as errors which we don't want in VS. @minestarks and I discussed this and the justification is that lint issues shouldn't look more severe than JS issues (which can be configured to be turned off).
2. Some may have been left out because we don't deem them important enough to report to the user

Of course users can always change their own version but this is the default shipped so this is what we want the templates to be clean against

rachelgshaffer:
The way I set up the action here won't work. It doesn't run actions from forks. I am going to move the linting step to the devops build. I'll let you know when this is all cleaned up.

Issue #2176, opened 2019-05-31
NullReferenceException in AttemptUpdateAsync


sharwell:
When connecting to a Live Share session, I'm observing a NullReferenceException in PackageJsonTestContainerDiscoverer.AttemptUpdateAsync which causes the Live Share session to fail. At minimum, this exception should not be propagating through the `PackageJsonTestContainerDiscoverer.OnActiveWorkspaceChangedAsync`.

##### Actual Behavior 

- NTVS Version: 1.5.10424.1 c3ce0ae0
- Visual Studio Version: 16.2.0 Preview 1.0
- Node.js Version: Unknown

##### Steps to Reproduce
1. Open a Live Share link from another user

amcasey:
Thanks, @sharwell!  I'm guessing `GetIndexWorkspaceService` returns null for the Live Share workspace.

IlyaBiryukov:
There are Live Share customers hitting this issue. Can you please prioritize it?
https://github.com/MicrosoftDocs/live-share/issues/2970, 
https://devdiv.visualstudio.com/DevDiv/_queries/edit/1049398/?triage=true
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1049869

amcasey:
Thanks, @IlyaBiryukov!  FYI @minestarks

minestarks:
@IlyaBiryukov  thanks - it's on my radar, I'll make sure we're treating it with high priority

pedropablo:
This is fixed but not released yet, right? I had to uninstall node.js development to recover live share functionality :(

minestarks:
This will be in the next 16.5 Preview. Sorry for the frustrating experience!

PR #2255, opened 2020-01-14
Fixing the LiveShare null reference exception

rakatyal:
This PR fixes https://github.com/microsoft/nodejstools/issues/2176. 

The indexService is null for a LiveShare workspace so this adds a null check to get rid of the exception. 

TestExplorer support with LiveShare is currently only enabled for LiveShare insiders and is still a little buggy for C#. We should revisit this to light it up when it becomes stable.

### DIFF
--- a/Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
+++ b/Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
@@ -66,20 +66,26 @@ namespace Microsoft.NodejsTools.Workspace
             if (workspace != null)
             {
                 var indexService = workspace.GetIndexWorkspaceService();
-                var filesDataValues = await indexService.GetFilesDataValuesAsync<string>(NodejsConstants.TestRootDataValueGuid);
-
-                lock (this.containerLock)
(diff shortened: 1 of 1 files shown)


rakatyal:
@amcasey @minestarks

Nodejs/Product/Nodejs/Workspace/PackageJsonTestContainerDiscoverer.cs
```diff
@@ -66,20 +66,26 @@ private async Task AttemptUpdateAsync()
             if (workspace != null)
             {
                 var indexService = workspace.GetIndexWorkspaceService();
-                var filesDataValues = await indexService.GetFilesDataValuesAsync<string>(NodejsConstants.TestRootDataValueGuid);
-
-                lock (this.containerLock)
+                // This returns null for LiveShare workspace so adding a check.
```
amcasey: Can you please add a comment about the consequences of skipping the work below?  I believe it just means no tests will be discovered and the Test Explorer will be empty?

rakatyal: That's correct. Will add!

PR #2259, opened 2020-02-14
Launching v3 debugger under a feature flag

Shenniey:
We've have a new debug adapter that can be used to debug Node projects (see [here](https://github.com/microsoft/vscode-js-debug)). This PR is to enable launching that debug adapter under a feature flag.

### DIFF
--- a/Nodejs/Product/Nodejs/Nodejs.csproj
+++ b/Nodejs/Product/Nodejs/Nodejs.csproj
@@ -1139,6 +1139,9 @@
     <PackageReference Include="Microsoft.CSharp">
       <Version>4.5.0</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.Internal.VisualStudio.Shell.Interop.14.0.DesignTime">
+      <Version>14.0.25023-setup</Version>
+    </PackageReference>
     <PackageReference Include="Microsoft.TestPlatform.ObjectModel">
       <Version>16.0.1</Version>
     </PackageReference>
--- a/Nodejs/Product/Nodejs/Nodejs.csproj
+++ b/Nodejs/Product/Nodejs/Nodejs.csproj
@@ -1140,7 +1140,7 @@
       <Version>4.5.0</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.Internal.VisualStudio.Shell.Interop.14.0.DesignTime">
-      <Version>14.0.25023-setup</Version>
+      <Version>14.0.0</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.TestPlatform.ObjectModel">
       <Version>16.0.1</Version>
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
@@ -25,6 +25,7 @@ using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudioTools.Project;
+using Microsoft.Internal.VisualStudio.Shell.Interop;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.NodejsTools.Project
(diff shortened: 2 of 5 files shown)


Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -1139,6 +1139,9 @@
     <PackageReference Include="Microsoft.CSharp">
       <Version>4.5.0</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.Internal.VisualStudio.Shell.Interop.14.0.DesignTime">
+      <Version>14.0.25023-setup</Version>
```
amcasey: I'm kind of surprised there's a suffix on the version number.  For the LS, we're just using `14.0.0`.

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -215,6 +217,12 @@ internal static string GetFullUrl(string host, int port)
             return builder.ToString();
         }
 
+        internal static bool ShouldUseV3CdpDebugger()
```
amcasey: Is this only called from the UI thread?

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -215,6 +217,12 @@ internal static string GetFullUrl(string host, int port)
...
+        {
+            var featureFlagsService = (IVsFeatureFlags)ServiceProvider.GlobalProvider.GetService(typeof(SVsFeatureFlags));
+            return featureFlagsService != null && featureFlagsService.IsFeatureEnabled("JavaScript.Debugger.V3CdpDebugAdapter", false);
```
amcasey: I think it's fine to require a VS restart to pick up a feature flag change (I'm not even sure it _can_ change during a session), so I'd cache the result after retrieving the first time.  I might even compute it eagerly if that let me guarantee I was on the UI thread.

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -280,7 +288,7 @@ private void StartWithChromeV2Debugger(string file, string nodePath, bool should
             var debugTargets = new[] {
                 new VsDebugTargetInfo4() {
                     dlo = (uint)DEBUG_LAUNCH_OPERATION.DLO_CreateProcess,
-                    guidLaunchDebugEngine = WebKitDebuggerV2Guid,
+                    guidLaunchDebugEngine = ShouldUseV3CdpDebugger() ? JsCdpDebuggerV3Guid : WebKitDebuggerV2Guid,
```
amcasey: Maybe the helper should just return the engine GUID?

Shenniey: Done!

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
Shenniey: Caching might get confusing though since the feature flag we're using (under Tools->Options->Environment->Preview Features) can change within a session of VS (though not within a debugging session in VS). I've added a check to make sure we're on the UI thead though! If we're not on the UI thread, it just doesn't launch the debugger and shows an error dialog about needing to be on the UI thread.

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
Shenniey: Yep! This is called in: 

1) NodejsProjectLauncher.cs, which, when walking up the callstack, is ultimately fulfilling the launch functionality of . Assuming Microsoft.ViualSudio.Shell.Interop.IVsDebuggableProjectCfg, which is guaranteed UI thread.

2) JsFileDebugLaunchTargetProvider, which is called by Microsoft.VisualStudio.Workspace.Extensions.VS.Debug.IVsDebugLaunchTargetProvider, and assuming this IVsDebugLaunchTargetProvider is guaranteed UI thread, JsFileDebugLaunchTargetProvider usage is fine.

3) NodeJsDebugLaunchTargetProvider, and assuming Microsoft.VisualStudio.Workspace.Extensions.VS.Debug.IVsDebugLaunchTargetProvider2 is guaranteed UI thread, NodeJsDebugLaunchTargetProvider usage is fine

Nodejs/Product/Nodejs/Nodejs.csproj
Shenniey: Fixed!

PR #2260, opened 2020-02-28
launching node using tools option

Shenniey:
We want to launch the new node debug adapter based on the debug tools option


Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -222,11 +223,24 @@ internal static Guid GetDebuggerGuid()
...
+                if (userDebuggerOption == null)
+                {
+                    return WebKitDebuggerV2Guid;
+                }
+                else
+                {
+                    return ((int) userDebuggerOption != 0) ? JsCdpDebuggerV3Guid : WebKitDebuggerV2Guid;
+                }
+            }
+            catch (Exception e)
```
armanio123: Since `e` is not used you should just have:
```suggestion
            catch (Exception)
```

PR #1937, opened 2018-04-20
Add support for using javascript unit tests in other project types

paulvanbrenk:
This enables the javascript unit test components to work with other
project types than njsproj.

If a project has a TestRoot and a TestFramework specified, we use those
to locate the files containing the unit tests. This also includes changes
to the node project system to make those 2 properties accessible.

PR #2269, opened 2020-06-13
Remove unused code from TestAdapterImpl

uniqueiniquity:
##### Bug
VS MEF logs include the following:
```
Microsoft.NodejsTools.TestAdapter.TestMethodResolver.ctor(discoverer): expected exactly 1 export matching constraints:
    Contract name: Microsoft.NodejsTools.TestAdapter.TestContainerDiscoverer
    TypeIdentityName: Microsoft.NodejsTools.TestAdapter.TestContainerDiscoverer
but found 0.
   part definition Microsoft.NodejsTools.TestAdapter.TestMethodResolver
```

##### Fix
Fixed the exported type of the `TestContainerDiscover` to be the class, rather than the implemented interface. However, the entrypoint of the `TestMethodResolver` was never correctly implemented, so I've fixed it to return null immediately now that it's being successfully invoked.

##### Testing
Validated that the MEF log is clean, that `GetCurrentTest` is now successfully invoked, and that returning `null` is handled successfully by the caller.

**Edit:**
After discussing the code review based on the comments, we decided to just pull out all unused code, rather than just this one entry point. After looking through the `TestAdapterImpl` project, it appears to have two MEF entrypoints - the one that caused this issue, and an implementation of `IRunFromContextFileExtensions`. I've deleted the non-functioning implementation of `ITestMethodResolver`, as well as all code that was only referenced by it, and left the implementation of `IRunFromContextFileExtensions`. I've also left the `VisualStudioApp` class, which is used by `TestAdapter`.


Nodejs/Product/TestAdapterImpl/TestMethodResolver.cs
```diff
@@ -31,6 +31,9 @@ internal class TestMethodResolver : ITestMethodResolver
 
         public string GetCurrentTest(string filePath, int line, int lineCharOffset)
         {
+            // TODO: finish implementing this method
```
armanio123, line 34: Is this pending for the PR? If not, shouldn't we remove the unreachable code after the `return null`?

uniqueiniquity: Nope, not pending for this PR. I'm happy to remove it; I didn't initially because it looked like someone had started working on it. But I agree now that it should be removed.

uniqueiniquity: Actually, I didn't realize how old that unfinished code was. I'm _definitely_ removing it.

uniqueiniquity: I've removed the unreachable code as suggested, as well as some more unused code.
I've tried to strike a balance in deleting unused code here; technically, this class does nothing, so we could delete the whole thing. But in case someone _did_ want to implement this, I left the barebones structure. Let me know if you think I should keep ripping stuff out.

uniqueiniquity: Alright, I've ripped the whole thing out.

uniqueiniquity:
cc @amcasey in case you're interested.

uniqueiniquity:
Looks like #1937 moved a lot of the functioning contents of this project to `TestAdapter`.

PR #2273, opened 2020-07-02
Added argument validation

armanio123:
Fixes an exception when the user is setting an invalid path on the Node.js location field on the project properties.

Now it will display a message indicating that the path is "invalid/not found".


armanio123:
We don't have. On the test I have executed, what worked was using invalid characters like a single quote.

PR #2284, opened 2020-08-17
Fix mocha test adapter failing to set extended timeout in debugging

yegorpetrov:
There's a bug: mocha tests time out in 2 seconds even under debugger. It can be reproduced by stepping through a mocha test for longer than two seconds.

I found that the mocha test adapter contains a check for the `v8debug` object to detect debugging scenarios. That interface seems to have been deprecated and/or EOL'd. At least it's not available in v12.14.1. The Inspector API, on the other hand, has been available since version 8, and thus it should be a safe replacement.

Here's my environment where I found and successfully patched the bug:
- Microsoft Visual Studio Community 2019 Version 16.6.0
- Node.js Tools 1.5.20317.1 Commit Hash:3e70368beb9630c811076c051f4c9a59b45d7c10
- node --version v12.14.1

### DIFF
--- a/Nodejs/Product/TestAdapter/TestFrameworks/mocha/mocha.js
+++ b/Nodejs/Product/TestAdapter/TestFrameworks/mocha/mocha.js
@@ -3,6 +3,7 @@
 var EOL = require('os').EOL;
 var fs = require('fs');
 var path = require('path');
+var inspector = require('inspector');
 // Choose 'tap' rather than 'min' or 'xunit'. The reason is that
 // 'min' produces undisplayable text to stdout and stderr under piped/redirect, 
 // and 'xunit' does not print the stack trace from the test.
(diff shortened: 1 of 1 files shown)


Nodejs/Product/TestAdapter/TestFrameworks/mocha/mocha.js
```diff
@@ -205,7 +206,7 @@ function getMochaOptions(projectFolder) {
     }
 
     // set timeout to 10 minutes, because the default of 2 sec is too short for debugging scenarios
-    if (typeof v8debug === 'object') {
+    if (inspector.url()) {
```
armanio123, line 209: Thanks for the fix.

I noticed that the `inspector` module is marked as `Stability: 1 - Experimental` and not recommended for production. Instead, I think is a good idea to follow TypeScript's implementation.

Ignore the VS Code variables. Here's the implementation: https://github.com/microsoft/TypeScript/blob/598e9b2e88153399bab4902f986910226b0a8496/src/compiler/sys.ts#L1252

PR #2295, opened 2020-10-13
Add angular test adapter

armanio123:
Adds the capability to run and discovers Angular tests.

Details:
- Changed the behavior of how it discovers tests. For Angular only, it will search for an `angular.json` file instead of the user manually specifying the test adapter. This is done with the purpose of facilitating the configuration but also for identifying the root path of the tests. Currently, it's only discoverable if the file is on the root path; this is an NTVS limitation that I will work on a separate PR.
- During execution, I have changed to send all of the selected tests instead of a by file basis. The by file basis strategy is not useful for angular as it requires to spin up a server every time, Causing execution to appear slow.
- Excluded searching for tests on node_modules folders. Affects all test adapters.
- Added support for `esproj`.

Extra notes:
- There are a bunch of todos that still require to be worked out. I'll be addressing those on a different PR as they are not required for Angular tests to work.
- No support yet for .NET Core. Will be coming next.
- Pending discussion of it's a good idea to implement discovery and execution by config file instead of manual configuration fot the rest of the tests adapters.

### DIFF
--- a/Nodejs/Product/Nodejs/NodejsConstants.cs
+++ b/Nodejs/Product/Nodejs/NodejsConstants.cs
@@ -17,7 +17,7 @@ namespace Microsoft.NodejsTools
         public const string NodejsProjectExtension = ".njsproj";
         public const string CSharpProjectExtension = ".csproj";
         public const string VisualBasicProjectExtension = ".vbproj";
-        public const string AngularProjectExtension = ".esproj";
+        public const string JavaScriptProjectExtension = ".esproj";
 
         public const string JavaScript = "JavaScript";
         public const string CSS = "CSS";
--- a/Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
+++ b/Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
@@ -4,7 +4,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using Microsoft.Build.Framework.XamlTypes;
 using Microsoft.NodejsTools.TestAdapter.TestFrameworks;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
--- a/Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
+++ b/Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
@@ -4,11 +4,11 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using Microsoft.Build.Framework.XamlTypes;
 using Microsoft.NodejsTools.TestAdapter.TestFrameworks;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
-using Microsoft.VisualStudioTools;
 using MSBuild = Microsoft.Build.Evaluation;
 
 namespace Microsoft.NodejsTools.TestAdapter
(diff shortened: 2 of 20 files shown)


Nodejs/Product/Nodejs/NodejsConstants.cs
```diff
@@ -17,6 +17,7 @@ internal static class NodejsConstants
         public const string NodejsProjectExtension = ".njsproj";
         public const string CSharpProjectExtension = ".csproj";
         public const string VisualBasicProjectExtension = ".vbproj";
+        public const string AngularProjectExtension = ".esproj";
```
uniqueiniquity, line 20: Is `.esproj` the new JSPS project extension? If so, let's not specifically refer to it as "Angular".

Nodejs/Product/TestAdapter/TestFrameworkFactory.cs
```diff
@@ -0,0 +1,111 @@
...
+                    }
+                }
+            }
+
+            return configItems.Any() ? configItems : testItems;
+        }
+
+        private static (string, string)? GetTestFrameworkAndFilePath(Project project, ProjectItem projectItem, string projectRoot)
+        {
+            var testRoot = project.GetProperty(NodeProjectProperty.TestRoot)?.EvaluatedValue;
```
uniqueiniquity, line 63: Do we really need to make these two calls for every item in a project?

uniqueiniquity (commented):
I'm pretty confused by the TestFrameworkFactory implementation - let's discuss when you have a chance.

I've just gone through the C# parts of this - I'll have to go through the JS parts in a second pass.

Nodejs/Product/Nodejs/NodejsConstants.cs
armanio123: Agree. Any suggested name for it?

uniqueiniquity: I think either `JavaScriptProjectExtension` to match our common name for it (JSPS) or `ECMAScriptProjectExtension` to patch the `es` part is fine.

Nodejs/Product/TestAdapter/TestFrameworks/find_tests.js
```diff
@@ -23,5 +26,4 @@ try {
 } catch (exception) {
     console.log("NTVS_ERROR:TestFramework (" + process.argv[2] + ") threw an exception processing (" + process.argv[3] + "), " + exception);
     process.exit(1);
-}
-process.exit(0);
```
uniqueiniquity, line 27: Rather than deleting this, would it make sense for each framework to specify whether they're sync or async and, in the former case, do this, but in the latter case, explicitly exit when the promise resolves? This might be fine, I was just a little surprised that find_tests was suddenly returning a promise, since the calling pattern doesn't suggest that as the case at all. I also think without handling that, you might not be catching exceptions it's throwing in the same way that you are for the synchronous frameworks.

Nodejs/Product/TestAdapter/TestFrameworks/Angular/vsKarmaReporter.js
```diff
@@ -0,0 +1,116 @@
+﻿// @ts-check
+
+const http = require('http');
+const path = require('path');
+
+const jasmineReporterPath = path.resolve(process.env.VSTESTADAPTERPATH, 'jasmineReporter.js');
+const testCases = JSON.parse(process.env.TESTCASES);
+const isDiscovery = process.env.ISDISCOVERY === 'true';
+
+const vsKarmaReporter = function (baseReporterDecorator, config, logger, emitter) {
```
uniqueiniquity, line 10: Any particular reason you chose to write this as an ES5 class?
I feel like it would be much more readable if we could just use ES6 classes.

Nodejs/Product/TestAdapter/TestFrameworks/Angular/karmaConfig.js
```diff
@@ -0,0 +1,42 @@
+﻿// @ts-check
+
+const path = require('path');
+
+const karmaConfigPath = path.resolve(process.env.PROJECTPATH, 'karma.conf.js');
```
uniqueiniquity, line 5: Will it be documented anywhere that this has to be in the project root?

uniqueiniquity (changes requested):

Nodejs/Product/Nodejs/NodejsConstants.cs
armanio123: `JavaScriptProjectExtension` it is. Done.

Nodejs/Product/TestAdapter/TestFrameworkFactory.cs
armanio123: We do not. Nonetheless I feel like placing them inside the method makes sense as they are part of the `project` variable. Otherwise we'll need to extract them on the caller and send them as a parameter.

Is your concern based on performance? I don't think it should be a problem, although I admit I don't know much of `GetProperty` to have an argument.

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
```diff
@@ -91,6 +91,30 @@ public void RunTests(IEnumerable<TestCase> tests)
             ValidateArg.NotNull(tests, nameof(tests));
             this.cancelRequested.Reset();
 
+            var testFramework = tests.First().GetPropertyValue(JavaScriptTestCaseProperties.TestFramework, defaultValue: TestFrameworkDirectories.ExportRunnerFrameworkName);
+            if (testFramework == "Angular")
```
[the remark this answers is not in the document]
armanio123: Done. Extracted to a `const`.

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
```diff
@@ -91,6 +91,30 @@ public void RunTests(IEnumerable<TestCase> tests)
...
+            {
+                RunAllTests(tests);
+            }
+            else
```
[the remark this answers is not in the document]
armanio123: It is that in Angular, is better to run all of the tests in a given "project" (as defined in `angular.json` and `karma.conf.js`), otherwise we'll initialize karma, browsers, cache, etc, multiple times. I have added a small comment on the line.

Technically other languages don't need this processing. I'm not entirely sure why is implemented like this. I believe some time ago there used to be a push for parallelizing this process, but I prefer to let the test frameworks do their stuff (like in the case angular implementation).

Hope this answers your questions.

Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
```diff
@@ -61,74 +60,13 @@ private void DiscoverTestsCore(IEnumerable<string> sources, IMessageLogger logge
                     }
 
                     FrameworkDiscoverer frameworkDiscoverer = null;
+                    var testItems = new Dictionary<string, HashSet<string>>(StringComparer.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
armanio123: No reason, I must have refactored this and missed it. Fixed.

Nodejs/Product/TestAdapter/TestFrameworkFactory.cs
```diff
@@ -0,0 +1,111 @@
...
+                string testFrameworkName = null;
+                // TODO: Consider it can also be inside a <root>/.config/ folder and will configure the <root> folder.
+                if (projectItem.EvaluatedInclude.Contains("angular.json"))
+                {
+                    testFrameworkName = "angular";
+                }
+                // TODO: Add configuration files for jest, mocha , jasmine. Consider that some frameworks can be configured using package.json
+                // Also some frameworks have more than one filename to define a configuration.
+                // Tape is the only framework we support that doesn't have a configuration file. Decide what to do about that.
+                if (testFrameworkName != null)
```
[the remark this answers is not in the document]
armanio123: Makes sense. Changed.

Nodejs/Product/TestAdapter/TestFrameworks/Angular/karmaConfig.js
armanio123: Based on new evidence this will change. Actually the karma config path is defined in `angular.json`. I'm working on implementing that.

Nodejs/Product/TestAdapter/TestFrameworks/NodejsTestInfo.cs
```diff
@@ -10,8 +10,6 @@ public sealed class NodejsTestInfo
     {
         public NodejsTestInfo(string testPath, string suite, string testName, string testFramework, int line, int column, string projectRootDir)
         {
-            Debug.Assert(testPath.EndsWith(".js", StringComparison.OrdinalIgnoreCase) || testPath.EndsWith(".jsx", StringComparison.OrdinalIgnoreCase));
-
             var testFileRelative = CommonUtils.GetRelativeFilePath(projectRootDir, testPath);
```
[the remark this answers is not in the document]
armanio123: I believe removing this assertion makes sense, as this information should come from the test framework. In the case of angular the default paths will have a `.ts` extension. This is true for other test frameworks as well. E.g. jest might return tests with `.ts` or `.tsx` as extension. 

Those are the only extensions I have seen but I wouldn't be surprised is somebody comes with something else. (looking at `.mjs`)

Let me know if you feel otherwise.

Nodejs/Product/TestAdapter/TestFrameworks/Angular/vsKarmaReporter.js
armanio123: You mean `= function` the `$inject` and `export`? If that's the case, I just followed the documentation on creating Karma reporters. I didn't saw any example that didn't use this so I just rolled with it. But I do agree with you that having an ES6 class is nicer to eye. 

Let's do a quick call on this.

uniqueiniquity (approved):

PR #2296, opened 2020-10-29
fix: source map overrides changed to dictionary

EricCornelson:
The "sourceMapPathOverrides" field in WebDiagnosticsComponents changed from a string array to a Dictionary<string, string>. Reflecting that change here.

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
@@ -652,7 +652,7 @@ namespace Microsoft.NodejsTools.Project
         public object Server { get; set; } 
 
         [JsonProperty("sourceMapPathOverrides")]
-        public string[] SourceMapPathOverrides { get; set; }
+        public Dictionary<string, string> SourceMapPathOverrides { get; set; }
 
         [JsonProperty("restart")]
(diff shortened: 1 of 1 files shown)


Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -669,7 +669,6 @@ internal class NodePinezorroDebugLaunchConfig
         public object toPwaChromeServerConfig()
         {
             this.Console = "internalConsole";
-            this.SourceMapPathOverrides = new string[0];
```
amcasey, line 672: How come this is no longer necessary?

EricCornelson: Looks like it is! Adding it back for now, and taking a look at why it's required for us and not vscode.

amcasey:
This code should only have to deserialize messages in the new format, right?  It will never be used with code that still sends arrays?

EricCornelson:
> This code should only have to deserialize messages in the new format, right? It will never be used with code that still sends arrays?

We'll have to coordinate the insertion with the WebDiagnosticComponents insertion, but yes, otherwise there shouldn't be any mismatch.

amcasey (approved):

PR #2297, opened 2020-10-29
fix: extra slash in devenv path for launching vsix project

EricCornelson:
Removing an extra slash in the devenv path which seems to break launching the vsix project


minestarks (approved):

amcasey (approved):
Weird, Windows usually ignores double slashes.

PR #2299, opened 2020-11-03
Add .NET Core Angular unit test support

armanio123:
- Adds support for identifying Angular/Karma/Jasmine unit tests in .NET Core projects. Works for Visual Studio and dotnet cli.
- Fixes issue with identifying a root path.
- Some refactoring.

### DIFF
--- a/Nodejs/Product/TestAdapter/SerializationHelpers.cs
+++ b/Nodejs/Product/TestAdapter/SerializationHelpers.cs
@@ -40,22 +40,22 @@ internal sealed class TestCaseObject
         this.testFile = string.Empty;
         this.workingFolder = string.Empty;
         this.projectFolder = string.Empty;
-        this.configPath = string.Empty;
+        this.configDirPath = string.Empty;
     }
 
-    public TestCaseObject(string framework, string fullyQualifiedName, string testFile, string workingFolder, string projectFolder, string configPath)
+    public TestCaseObject(string framework, string fullyQualifiedName, string testFile, string workingFolder, string projectFolder, string configDirPath)
     {
         this.framework = framework;
         this.fullyQualifiedName = fullyQualifiedName;
         this.testFile = testFile;
         this.workingFolder = workingFolder;
         this.projectFolder = projectFolder;
-        this.configPath = configPath;
+        this.configDirPath = configDirPath;
     }
     public string framework { get; set; }
     public string fullyQualifiedName { get; set; }
     public string testFile { get; set; }
     public string workingFolder { get; set; }
     public string projectFolder { get; set; }
-    public string configPath { get; set; }
+    public string configDirPath { get; set; }
 }
--- a/Nodejs/Product/TestAdapter/SerializationHelpers.cs
+++ b/Nodejs/Product/TestAdapter/SerializationHelpers.cs
@@ -40,19 +40,22 @@ internal sealed class TestCaseObject
         this.testFile = string.Empty;
         this.workingFolder = string.Empty;
         this.projectFolder = string.Empty;
+        this.configPath = string.Empty;
     }
 
-    public TestCaseObject(string framework, string fullyQualifiedName, string testFile, string workingFolder, string projectFolder)
+    public TestCaseObject(string framework, string fullyQualifiedName, string testFile, string workingFolder, string projectFolder, string configPath)
     {
         this.framework = framework;
         this.fullyQualifiedName = fullyQualifiedName;
         this.testFile = testFile;
         this.workingFolder = workingFolder;
         this.projectFolder = projectFolder;
+        this.configPath = configPath;
     }
     public string framework { get; set; }
     public string fullyQualifiedName { get; set; }
     public string testFile { get; set; }
     public string workingFolder { get; set; }
     public string projectFolder { get; set; }
+    public string configPath { get; set; }
 }
--- a/Nodejs/Product/TestAdapter/TestFrameworks/Angular/Angular.js
+++ b/Nodejs/Product/TestAdapter/TestFrameworks/Angular/Angular.js
@@ -10,7 +10,7 @@ process.env.VSTESTADAPTERPATH = __dirname;
 const vsKarmaConfigFilePath = path.resolve(__dirname, "./karmaConfig.js");
 
 function getTestProjects(configFile) {
-    const configPath = path.dirname(configFile);
+    const configDirPath = path.dirname(configFile);
 
     const angularProjects = [];
     const angularConfig = require(configFile);
@@ -20,14 +20,14 @@ function getTestProjects(configFile) {
         const karmaConfigFilePath = project.architect.test
             && project.architect.test.options
             && project.architect.test.options.karmaConfig
-            && path.resolve(configPath, project.architect.test.options.karmaConfig);
+            && path.resolve(configDirPath, project.architect.test.options.karmaConfig);
 
         if (karmaConfigFilePath) {
             angularProjects.push({
-                angularConfigPath: configPath,
+                angularConfigDirPath: configDirPath,
                 karmaConfigFilePath,
                 name: projectName,
-                rootPath: path.join(configPath, project.root),
+                rootPath: path.join(configDirPath, project.root),
             });
         }
     }
@@ -39,9 +39,9 @@ const find_tests = async function (configFiles, discoverResultFile) {
     const projects = [];
 
     for (const configFile of configFiles.split(';')) {
-        const configPath = path.dirname(configFile);
+        const configDirPath = path.dirname(configFile);
 
-        if (!detectPackage(configPath, '@angular/cli')) {
+        if (!detectPackage(configDirPath, '@angular/cli')) {
             continue;
         }
 
@@ -58,7 +58,7 @@ const find_tests = async function (configFiles, discoverResultFile) {
         // on a lock file for building the project, that might be the reason.
         await new Promise((resolve, reject) => {
             const ngTest = fork(
-                path.resolve(project.angularConfigPath, "./node_modules/@angular/cli/bin/ng"),
+                path.resolve(project.angularConfigDirPath, "./node_modules/@angular/cli/bin/ng"),
                 [
                     'test',
                     project.name,
@@ -69,7 +69,7 @@ const find_tests = async function (configFiles, discoverResultFile) {
                         ...process.env,
                         PROJECT: JSON.stringify(project)
                     },
-                    cwd: project.angularConfigPath,
+                    cwd: project.angularConfigDirPath,
                 }).on('message', message => {
                     testsDiscovered.push(message);
 
@@ -101,16 +101,16 @@ const run_tests = async function (context) {
 
     // Get all the projects
     const projects = [];
-    const angularConfigPaths = new Set();
+    const angularConfigDirectories = new Set();
     for (let testCase of context.testCases) {
-        if (!angularConfigPaths.has(testCase.configPath)) {
-            angularConfigPaths.add(testCase.configPath);
+        if (!angularConfigDirectories.has(testCase.configDirPath)) {
+            angularConfigDirectories.add(testCase.configDirPath);
 
-            if (!detectPackage(testCase.configPath, '@angular/cli')) {
+            if (!detectPackage(testCase.configDirPath, '@angular/cli')) {
                 continue;
             }
 
-            projects.push(...getTestProjects(`${testCase.configPath}/angular.json`));
+            projects.push(...getTestProjects(`${testCase.configDirPath}/angular.json`));
         }
     }
 
@@ -122,7 +122,7 @@ const run_tests = async function (context) {
         // on a lock file for building the project, that might be the reason.
         await new Promise((resolve, reject) => {
             const ngTest = fork(
-                path.resolve(project.angularConfigPath, "./node_modules/@angular/cli/bin/ng"),
+                path.resolve(project.angularConfigDirPath, "./node_modules/@angular/cli/bin/ng"),
                 [
                     'test',
                     project.name,
@@ -133,7 +133,7 @@ const run_tests = async function (context) {
                         ...process.env,
                         PROJECT: JSON.stringify(project)
                     },
-                    cwd: project.angularConfigPath,
+                    cwd: project.angularConfigDirPath,
                     stdio: ['ignore', 1, 2, 'ipc'] // We need to ignore the stdin as NTVS keeps it open and causes the process to wait indefinitely.
                 }).on('message', message => {
                     context.post({
--- a/Nodejs/Product/TestAdapter/TestFrameworks/Angular/Angular.js
+++ b/Nodejs/Product/TestAdapter/TestFrameworks/Angular/Angular.js
@@ -7,7 +7,7 @@ const { fork } = require("child_process");
 
 process.env.VSTESTADAPTERPATH = __dirname;
 
-const vsKarmaConfigPath = path.resolve(__dirname, "./karmaConfig.js");
+const vsKarmaConfigFilePath = path.resolve(__dirname, "./karmaConfig.js");
 
 function getTestProjects(configFile) {
     const configPath = path.dirname(configFile);
@@ -17,16 +17,17 @@ function getTestProjects(configFile) {
     for (const projectName of Object.keys(angularConfig.projects)) {
         const project = angularConfig.projects[projectName];
 
-        const karmaConfigPath = project.architect.test
+        const karmaConfigFilePath = project.architect.test
             && project.architect.test.options
             && project.architect.test.options.karmaConfig
             && path.resolve(configPath, project.architect.test.options.karmaConfig);
 
-        if (karmaConfigPath) {
+        if (karmaConfigFilePath) {
             angularProjects.push({
                 angularConfigPath: configPath,
-                karmaConfigPath,
+                karmaConfigFilePath,
                 name: projectName,
+                rootPath: path.join(configPath, project.root),
             });
         }
     }
@@ -61,7 +62,7 @@ const find_tests = async function (configFiles, discoverResultFile) {
                 [
                     'test',
                     project.name,
-                    `--karmaConfig=${vsKarmaConfigPath}`
+                    `--karmaConfig=${vsKarmaConfigFilePath}`
                 ],
                 {
                     env: {
@@ -78,7 +79,7 @@ const find_tests = async function (configFiles, discoverResultFile) {
                     ngTest.send({});
                 }).on('error', err => {
                     reject(err);
-                }).on('exit', (code) => {
+                }).on('exit', code => {
                     resolve(code);
                 });
         });
@@ -125,7 +126,7 @@ const run_tests = async function (context) {
                 [
                     'test',
                     project.name,
-                    `--karmaConfig=${vsKarmaConfigPath}`
+                    `--karmaConfig=${vsKarmaConfigFilePath}`
                 ],
                 {
                     env: {
@@ -142,8 +143,8 @@ const run_tests = async function (context) {
                     });
 
                     ngTest.send({});
-                }).on('exit', () => {
-                    resolve();
+                }).on('exit', code => {
+                    resolve(code);
                 }).on('error', err => {
                     reject(err);
                 });
--- a/Nodejs/Product/TestAdapter/TestFrameworks/Angular/Angular.js
+++ b/Nodejs/Product/TestAdapter/TestFrameworks/Angular/Angular.js
@@ -103,14 +103,14 @@ const run_tests = async function (context) {
     const projects = [];
     const angularConfigPaths = new Set();
     for (let testCase of context.testCases) {
-        if (!angularConfigPaths.has(testCase.projectFolder)) {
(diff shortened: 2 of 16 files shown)


Nodejs/Product/TestAdapter/TestFrameworks/Angular/Angular.js
```diff
@@ -17,16 +17,17 @@ function getTestProjects(configFile) {
...
             && path.resolve(configPath, project.architect.test.options.karmaConfig);
 
-        if (karmaConfigPath) {
+        if (karmaConfigFilePath) {
             angularProjects.push({
                 angularConfigPath: configPath,
-                karmaConfigPath,
+                karmaConfigFilePath,
                 name: projectName,
+                rootPath: path.join(configPath, project.root),
```
amcasey, line 30: Why `path.join`, rather than `path.resolve`, as above?

Nodejs/Product/TestAdapter/TestFrameworks/Angular/vsKarmaReporter.js
```diff
@@ -48,7 +48,7 @@ const vsKarmaReporter = function (baseReporterDecorator, config, logger, emitter
         // Increment the amount of test cases found.
         testCaseCount++;
 
-        const fullFilePath = path.join(path.dirname(project.karmaConfigPath), result.fileLocation.relativeFilePath);
+        const fullFilePath = path.join(project.rootPath, result.fileLocation.relativeFilePath);
```
amcasey, line 51: I think this is the old value plus `project.root`, but I'd find an example helpful.

Nodejs/Product/TestAdapter/TestFrameworks/Angular/vsKarmaReporter.js
```diff
@@ -64,6 +64,7 @@ const vsKarmaReporter = function (baseReporterDecorator, config, logger, emitter
             filepath: fullFilePath,
             line: result.fileLocation.line,
             column: result.fileLocation.column,
+            configPath: project.angularConfigPath,
```
amcasey, line 67: I'm having a little trouble with the naming convention.  It seems like "file path" is a file and "path" is a directory?

Nodejs/Product/TestAdapter/SerializationHelpers.cs
```diff
@@ -40,19 +40,22 @@ public TestCaseObject()
         this.testFile = string.Empty;
         this.workingFolder = string.Empty;
         this.projectFolder = string.Empty;
+        this.configPath = string.Empty;
     }
 
-    public TestCaseObject(string framework, string fullyQualifiedName, string testFile, string workingFolder, string projectFolder)
+    public TestCaseObject(string framework, string fullyQualifiedName, string testFile, string workingFolder, string projectFolder, string configPath)
```
amcasey, line 46: It seems strange to have this ctor when the properties are writeable.  Anyone consuming it could just use an object initializer, no?  And that would eliminate the need to keep the signature up-to-date?

Nodejs/Product/TestAdapter/TestFrameworks/Angular/vsKarmaReporter.js
amcasey, line 67: This is the (only?) place where the new configPath that gets threaded through everywhere is actually specified?  Where is it consumed?

Nodejs/Product/TestAdapterNetStandard/TestAdapterNetStandard.csproj
```diff
@@ -78,6 +78,18 @@
     <Content Include="..\TestAdapter\TestFrameworks\Jest\jestReporter.js" Link="TestFrameworks\Jest\jestReporter.js">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="..\TestAdapter\TestFrameworks\Angular\Angular.js" Link="TestFrameworks\Angular\Angular.js">
```
amcasey, line 81: These don't seem to be new - were they just missed?

Nodejs/Product/TestAdapter/TestFrameworks/Angular/Angular.js
```diff
@@ -156,18 +157,18 @@ const run_tests = async function (context) {
...
 
 function detectPackage(projectFolder, packageName) {
-    try {
-        const packagePath = path.join(projectFolder, 'node_modules', packageName);
-        const pkg = require(packagePath);
+    const packagePath = path.join(projectFolder, 'node_modules', packageName);
 
-        return pkg;
-    } catch (ex) {
+    if (!fs.existsSync(packagePath)) {
```
amcasey, line 162: `existsSync` won't do the same module resolution as `require` or handle things like matching `index.js` if the argument is a directory.  Does that matter?

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
```diff
@@ -147,10 +119,76 @@ void ProcessFiles(string fileAbsolutePath)
...
+                };
+            }
+
+            return default;
+        }
+
+        private Dictionary<string, HashSet<string>> GetTestItems(string projectRoot, string testFramework, string outDir)
+        {
+            // If we find ts or tsx files, get the JS file and return.
+            var files = Directory.EnumerateFiles(projectRoot, "*.ts?", SearchOption.AllDirectories)
```
amcasey, line 161: Doesn't `?` usually mean "exactly one character"?  It seems like you want "at most one character".

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
```diff
@@ -147,10 +119,76 @@ void ProcessFiles(string fileAbsolutePath)
...
+                .Where(x => !x.Contains("\\node_modules\\"));
```
amcasey, line 162: Anecdotally, you may find this is (much) slower than doing your own directory traversal and not descending into node_modules folders.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
```diff
@@ -147,10 +119,76 @@ void ProcessFiles(string fileAbsolutePath)
...
+            if (files.Any())
+            {
+                var testFilePaths = files
+                    .Where(x => TypeScriptHelpers.IsTypeScriptFile(x))
```
amcasey, line 166: Doesn't the search pattern guarantee `IsTypeScriptFile`?

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
```diff
@@ -147,10 +119,76 @@ void ProcessFiles(string fileAbsolutePath)
...
+                    .Select(x => TypeScriptHelpers.GetTypeScriptBackedJavaScriptFile(x, projectRoot, outDir));
```
amcasey, line 167: Nit: "x" is not a great name for a path.  Maybe "p"?

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
amcasey: Oh, because it could be ".tsq"?

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
```diff
@@ -147,10 +119,76 @@ void ProcessFiles(string fileAbsolutePath)
...
+                return new Dictionary<string, HashSet<string>>()
+                {
+                    {
+                        testFramework,
+                        new HashSet<string>(testFilePaths)
+                    }
+                };
+            }
+
+            files = Directory.EnumerateFiles(projectRoot, "*.js", SearchOption.AllDirectories)
```
amcasey, line 178: Not ".jsx"?

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
amcasey: If you decide to do your own traversal, you might want to build up the JS list at the same time and drop it if you don't need it.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
```diff
@@ -147,10 +119,76 @@ void ProcessFiles(string fileAbsolutePath)
...
+                        TestFrameworkDirectories.AngularFrameworkName,
+                        new HashSet<string>(files)
+                    }
+                };
+            }
+
+            return default;
+        }
+
+        private Dictionary<string, HashSet<string>> GetTestItems(string projectRoot, string testFramework, string outDir)
```
amcasey, line 158: This is a really expensive way to represent the data being returned, but maybe it will make more sense once I've read the rest of the file.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
```diff
@@ -147,10 +119,76 @@ void ProcessFiles(string fileAbsolutePath)
...
+            foreach (var testFx in testItems.Keys)
+            {
+                var testFramework = frameworkDiscoverer.GetFramework(testFx);
+                if (testFramework == null)
+                {
+                    logger.SendMessage(TestMessageLevel.Warning, $"Ignoring unsupported test framework '{testFx}'.");
+                    continue;
+                }
+
+                var fileList = testItems[testFx];
```
amcasey, line 133: Does it matter that these won't be ordered consistently?

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
```diff
@@ -147,10 +119,76 @@ void ProcessFiles(string fileAbsolutePath)
             }
         }
 
-        private void DiscoverTests(HashSet<string> fileList, TestFramework testFramework, ITestCaseDiscoverySink discoverySink, IMessageLogger logger, string nodeExePath, string projectFullPath)
+        private void DiscoverTests(Dictionary<string, HashSet<string>> testItems, FrameworkDiscoverer frameworkDiscoverer, ITestCaseDiscoverySink discoverySink, IMessageLogger logger, string nodeExePath, string projectFullPath)
         {
-            var discoverWorker = new TestDiscovererWorker(projectFullPath, NodejsConstants.ExecutorUri, nodeExePath);
-            discoverWorker.DiscoverTests(fileList, testFramework, logger, discoverySink);
+            foreach (var testFx in testItems.Keys)
```
amcasey, line 124: Consider iterating over key-value pairs.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
```diff
@@ -147,10 +119,76 @@ void ProcessFiles(string fileAbsolutePath)
...
+                var testFramework = frameworkDiscoverer.GetFramework(testFx);
+                if (testFramework == null)
+                {
+                    logger.SendMessage(TestMessageLevel.Warning, $"Ignoring unsupported test framework '{testFx}'.");
+                    continue;
+                }
+
+                var fileList = testItems[testFx];
+
+                var discoverWorker = new TestDiscovererWorker(projectFullPath, NodejsConstants.ExecutorUri, nodeExePath);
```
amcasey, line 135: Do we need a new one for each iteration?

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
amcasey: I'm still not getting it - it looks like the dictionary will always have either 0 or 1 elements.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
```diff
@@ -70,73 +72,43 @@ public void DiscoverTests(IEnumerable<string> sources, IDiscoveryContext discove
...
-                    }
 
+                    // Prioritize configuration files over manually setup tests files.
+                    var testItems = this.GetConfigItems(projectHome);
                     if (!testItems.Any())
                     {
-                        foreach (var file in Directory.EnumerateFiles(testFolder, "*.js", SearchOption.AllDirectories))
+                        var testFramework = propertyGroup.Descendants(NodeProjectProperty.TestFramework).FirstOrDefault()?.Value;
+                        var testRoot = propertyGroup.Descendants(NodeProjectProperty.TestRoot).FirstOrDefault()?.Value;
+                        var outDir = propertyGroup.Descendants(NodeProjectProperty.TypeScriptOutDir).FirstOrDefault()?.Value ?? "";
```
amcasey, line 83: Personally, I'd put the `?? ""` at the use site, to maintain the parallelism.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
```diff
@@ -70,73 +72,43 @@ public void DiscoverTests(IEnumerable<string> sources, IDiscoveryContext discove
 
                 foreach (var (projectFile, propertyGroup) in projects)
                 {
-                    var testFramework = propertyGroup.Descendants(NodeProjectProperty.TestFramework).FirstOrDefault()?.Value;
```
amcasey, line 73: I didn't carefully read the old implementation, so I don't know if it did something different.  Let me know if I should.

amcasey (commented):
Some thoughts.

Nodejs/Product/TestAdapter/TestFrameworks/Angular/Angular.js
armanio123: No particular reason, I think for this scenario both will work ok.

My first though is that we just need to merge the paths since `configPath` will be a full path and `root` will be relative; so I just went with join.

Nodejs/Product/TestAdapter/SerializationHelpers.cs
armanio123: I suppose in a way the consumer will be force to consider the variable value when initializing, but I don't have a particular preference. I could change it if you feel is more readable.

Nodejs/Product/TestAdapter/TestFrameworks/Angular/Angular.js
armanio123: It matters, `existsSync` wont load the file, it will only check it it exists. `require` will load it memory, execute and return the export. The latter will unnecessarily consume more resources, so I refactored.

Nodejs/Product/TestAdapter/TestFrameworks/Angular/vsKarmaReporter.js
armanio123: This will be the `angular.json` file path plus the `project.root` defined on it. I can add a comment.

Nodejs/Product/TestAdapter/TestFrameworks/Angular/vsKarmaReporter.js
armanio123: I feel you, I can rename it. Even during development I had some trouble understanding myself. But yes, "path" is just the directory so I'll name it like that.

armanio123: Yes, this is where we specify it.
During discovery, this will be consumed by the VS Test Adapter in `TestDiscovererWorker.cs:81`.
During execution this will be a property and used here: `TestExecutorWorker.cs:387`.

Hope it makes sense.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
armanio123: Actually is already handled downstream. So I have removed it.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
armanio123: Of course. I guess switching context on from JS affected this.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
armanio123: It doesn't matter.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
armanio123: Good call, we do not. Fixed.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
armanio123: This is a copy paste from the other tests adapters (where they might contain different test frameworks). Fixed.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
armanio123: Based on the Microsoft documentation, `?` means: "Zero or one character in that position."

I guess anybody can do the implementation of wildcards as they understand it. I suppose there's no standard about them.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
armanio123: Yeah, I consider that to be a scenario. I'll add a TODO and address that on an oncoming PR.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
armanio123: Sure.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
armanio123: `.jsx` requires a precompilation process, most likely not understood by this test adapter. I feel like for that scenario we will not be able to assume where the output is going to be.

Let me know if you want to discuss further about this. I could use brainstorm about it as I do believe it might be a valid scenario.

Nodejs/Product/TestAdapterNetStandard/TestAdapterNetStandard.csproj
armanio123: They where not. This are links to the existing `js` files to add support for .NET Core. Adding them in this project will include them on the nuget package generated.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
armanio123: The only difference is that it considers now configuration files and has been simplified.

Nodejs/Product/TestAdapter/TestFrameworks/Angular/Angular.js
amcasey: I think `require` does two things that `existsSync` does not: it loads the file - which, I agree, is undesirable - but it also does extra resolution to find the file and I don't know whether that's important.  If we're passing an absolute path to a file (not a directory), it probably isn't.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
amcasey: Weird.  Thanks for confirming.

Nodejs/Product/TestAdapterNetStandard/NetCoreTestDiscoverer.cs
amcasey: I can't say that I've ever used a jsx file myself, but I definitely had the impression that they could be consumed as-is, without pre-processing.  Happy to discuss further.

amcasey (approved):
My only remaining question is about existsSync vs require.  Otherwise, LGTM.

Nodejs/Product/TestAdapter/TestFrameworks/Angular/Angular.js
armanio123: I suppose if we want to be super thorough on checking the dependencies that behavior might be desirable.  We are looking into the directory. I admit I don't have enough knowledge to now how npm resolve things, I'm not confident to change it until looking further.

IMO, I don't think we need to be that meticulous when checking for dependencies as this is only for showing a custom message if they don't exists. If for any reason the directory exists but not the package, we still get an error message anyway. 

Let me know if you think otherwise, I can change it back if you feel it adds some value to have.

PR #2300, opened 2020-11-10
Accessibility bug fixes

armanio123:
Fixes bugs:
- Keyboard tab stop is provided for non actionable content. Keyboard tab stop is provided for the entire section
- MAS4.2.1: The Name property of a focusable element must not be null.

### DIFF
--- a/Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallWindow.xaml
+++ b/Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallWindow.xaml
@@ -296,6 +296,7 @@
             <ContentControl Style="{StaticResource SearchStyle}" HorizontalAlignment="Stretch" VerticalAlignment="Stretch" IsTabStop="False" Focusable="False" />
             <Grid Grid.Row="1" Background="{Binding ElementName=packageList, Path=Background}">
                 <ListView x:Name="packageList"
+                          AutomationProperties.Name="{x:Static resx:NpmInstallWindowResources.PackageListName}"
                           ItemsSource="{Binding FilteredPackages}"
                           SelectedItem="{Binding SelectedPackage}"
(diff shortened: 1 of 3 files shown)


Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallWindow.xaml
```diff
@@ -343,7 +344,7 @@
                               Content="{Binding Path=SelectedPackage}"
                               ContentTemplate="{StaticResource PackageInfoTemplate}"
                               IsTabStop="True" AutomationProperties.Name="{Binding Path=SelectedPackage.Name}"
-                              TabIndex="3" />
+                              Focusable="False"/>
```
jessetrinity, line 347: What is the difference between `IsTabStop` (just above) and `Focusable` here?

Also, is the general intent to not stop on the vertical scrollbar while tabbing?

armanio123: I didn't consider the scrolling! I don't know the answer for that, I'll follow up with the accessibility folks. 

In regards to `IsTabStop` and `Focusable` seems like the first is the right solution if we don't want to stop on the control. Depending on their answer about the scrolling, I'll fix this.

armanio123: Based on discussion with the accessibility experts there should be no change here. Ideally it should be only focusable when the scroll is visible, but since this is the default behavior of the control and changing it seems to be a ton of effort, for now, we should leave it as is.

Rollback the changes.

armanio123:
Submitted a revision. Could you please review again?

jessetrinity (approved):
Looks like this resolves the naming issue. Not sure about the tab stop bit without doing manual testing.

PR #2301, opened 2020-11-12
Update TS and Node types versions for TypeScript Template

jessetrinity:
We need at least TS `4.1` to fix the occurrence of [this](https://github.com/microsoft/TypeScript/issues/40951) issue in VS.

Two things
1) These versions were acquired via `npm install --save-dev typescript @types/node`. There is nothing special about them, they are just the latest published at this time. As long as the major version is `^4` we can change these to anything reasonable.
2) We should probably also update these versions for the other `TypeScript*` templates. Any reason not to?


jessetrinity:
Added JSTS first responders.

minestarks (approved):

EricCornelson (approved):

armanio123 (approved):
Thanks for updating all the templates

PR #2302, opened 2020-11-12
Add test adapter telemetry

armanio123:
Adds a telemetry event triggered when discovery starts.

It will capture information of the test adapter used (angular, mocha, jasmine, etc.) and the test adapter name (.NET Core, .NET Framework, NTVS, etc).

### DIFF
--- a/Nodejs/Product/TestAdapter/DotNetTestDiscoverer.cs
+++ b/Nodejs/Product/TestAdapter/DotNetTestDiscoverer.cs
@@ -1,0 +1,15 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
+
+namespace Microsoft.NodejsTools.TestAdapter
+{
+    // Keep in sync the method TypeScriptHelpers.IsSupportedTestProjectFile if there's a change on the supported projects.
+    [FileExtension(NodejsConstants.CSharpProjectExtension), FileExtension(NodejsConstants.VisualBasicProjectExtension)]
+    [DefaultExecutorUri(NodejsConstants.ExecutorUriString)]
+    public sealed class DotNetTestDiscoverer : ProjectTestDiscoverer, ITestDiscoverer
+    {
+        public override string TestDiscovererName => nameof(DotNetTestDiscoverer);
+    }
+}
--- a/Nodejs/Product/TestAdapter/DotNetTestDiscoverer.cs
+++ b/Nodejs/Product/TestAdapter/DotNetTestDiscoverer.cs
@@ -5,7 +5,7 @@ using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
 
 namespace Microsoft.NodejsTools.TestAdapter
 {
-    // Keep in sync the method TypeScriptHelpers.IsSupportedTestProjectFile if there's a change on the supported projects.
+    /// Keep in sync the method <see cref="TypeScript.TypeScriptHelpers.IsSupportedTestProjectFile(string)" /> if there's a change on the supported projects.
     [FileExtension(NodejsConstants.CSharpProjectExtension), FileExtension(NodejsConstants.VisualBasicProjectExtension)]
     [DefaultExecutorUri(NodejsConstants.ExecutorUriString)]
     public sealed class DotNetTestDiscoverer : ProjectTestDiscoverer, ITestDiscoverer
--- a/Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
+++ b/Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
@@ -110,7 +110,7 @@ namespace Microsoft.NodejsTools.TestAdapter
 
                 var fileList = testItems[testFx];
 
-                discoverWorker.DiscoverTests(fileList, testFramework, logger, discoverySink);
+                discoverWorker.DiscoverTests(fileList, testFramework, logger, discoverySink, nameof(ProjectTestDiscoverer));
             }
         }
     }
--- a/Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
+++ b/Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
@@ -12,11 +12,9 @@ using MSBuild = Microsoft.Build.Evaluation;
 
 namespace Microsoft.NodejsTools.TestAdapter
 {
-    // We require to put a non-existent file extension to avoid duplicate discovery executions.
-    [FileExtension("NTVS_NonExistentFileExtension")]
-    public class ProjectTestDiscoverer : ITestDiscoverer
+    public abstract class ProjectTestDiscoverer
     {
-        public virtual string TestDiscovererName => nameof(ProjectTestDiscoverer);
+        public abstract string TestDiscovererName { get; }
 
         public void DiscoverTests(IEnumerable<string> sources, IDiscoveryContext discoveryContext, IMessageLogger logger, ITestCaseDiscoverySink discoverySink)
         {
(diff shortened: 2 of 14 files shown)


Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
```diff
@@ -12,11 +12,12 @@
 
 namespace Microsoft.NodejsTools.TestAdapter
 {
-    // Keep in sync the method TypeScriptHelpers.IsSupportedTestProjectFile if there's a change on the supported projects.
-    [FileExtension(NodejsConstants.NodejsProjectExtension), FileExtension(NodejsConstants.CSharpProjectExtension), FileExtension(NodejsConstants.VisualBasicProjectExtension), FileExtension(NodejsConstants.JavaScriptProjectExtension)]
-    [DefaultExecutorUri(NodejsConstants.ExecutorUriString)]
-    public partial class ProjectTestDiscoverer : ITestDiscoverer
+    // We require to put a non-existent file extension to avoid duplicate discovery executions.
+    [FileExtension("NTVS_NonExistentFileExtension")]
+    public class ProjectTestDiscoverer : ITestDiscoverer
```
uniqueiniquity, line 17: `abstract` ?

Nodejs/Product/TestAdapter/DotNetTestDiscoverer.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
+
+namespace Microsoft.NodejsTools.TestAdapter
+{
+    // Keep in sync the method TypeScriptHelpers.IsSupportedTestProjectFile if there's a change on the supported projects.
```
amcasey, line 8: I like to use `cref` for cases like this - it enables things like GTD and FAR.

Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
```diff
@@ -12,11 +12,12 @@
...
     {
+        public virtual string TestDiscovererName => nameof(ProjectTestDiscoverer);
```
amcasey, line 19: `abstract`?

Nodejs/Product/TestAdapter/TestDiscovererWorker.cs
```diff
@@ -26,14 +27,16 @@ public TestDiscovererWorker(string testSource, string nodeExePath)
...
         {
             var fileList = Directory.EnumerateFiles(testFolderPath, "*.js", SearchOption.AllDirectories).Where(x => !x.Contains(NodejsConstants.NodeModulesFolder));
-            this.DiscoverTests(fileList, testFx, logger, discoverySink);
+            this.DiscoverTests(fileList, testFx, logger, discoverySink, testDiscovererName);
         }
 
-        public void DiscoverTests(IEnumerable<string> fileList, TestFramework testFx, IMessageLogger logger, ITestCaseDiscoverySink discoverySink)
+        public void DiscoverTests(IEnumerable<string> fileList, TestFramework testFx, IMessageLogger logger, ITestCaseDiscoverySink discoverySink, string testDiscovererName)
         {
+            TelemetryHelper.LogTestDiscoveryStarted(testFx.Name, testDiscovererName);
```
amcasey, line 38: The framework name comes from a fixed list we control?

amcasey (commented):
Minor questions.

Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
armanio123: This class is instantiated in: https://github.com/microsoft/nodejstools/blob/587d04559aaef7288c51ef8420fff67c38257006/Nodejs/Product/TestAdapter/ProjectTestExecutor.cs#L30

This happens when the users launched execution from the vstest cli.

Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
armanio123: Same as above, the class is instantiated so we declare the name for it. Inheritors replace it value, hence the virtual.

Nodejs/Product/TestAdapter/TestDiscovererWorker.cs
armanio123: The name comes from a list of directories we expect to exists within the VS installation. Technically, the user might be able to add more directories.

Sounds like we might need to fix to specific names?

Nodejs/Product/TestAdapter/DotNetTestDiscoverer.cs
armanio123: Added.

Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
amcasey: My personal preference would be to make this type abstract and make a new concrete type that extends it.  More generally, it's easier to reason about type hierarchies if everything is either abstract or sealed.

Nodejs/Product/TestAdapter/TestDiscovererWorker.cs
amcasey: Life would be simpler if we didn't report any user-specified strings, yes.  If that's complicated, grab me and we'll figure something out.

Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
armanio123: I don't mind one or another, so I change das suggested. I considered it before but decided to have the lease amount of classes was better for maintainability, but I do agree that having explicit abstract and sealed classes.

The new class name is `DefaultTestDiscoverer` because I'm not 100% sure when it is called. I have only seen it when using the vstest cli, but the documentation doesn't state exactly in what conditions it's called: "Runs 'all' the tests present in the specified 'sources'.". Let me know if you have a suggestion on the name.

uniqueiniquity (approved):

Nodejs/Product/TestAdapter/TestDiscovererWorker.cs
armanio123: I have added a simple whitelist before calling the method. Hope that's enough.

Nodejs/Product/TestAdapter/ProjectTestDiscoverer.cs
amcasey: "Default" seems good.  Thanks!

Nodejs/Product/TestAdapter/TestDiscovererWorker.cs
```diff
@@ -26,14 +38,20 @@ public TestDiscovererWorker(string testSource, string nodeExePath)
...
             var fileList = Directory.EnumerateFiles(testFolderPath, "*.js", SearchOption.AllDirectories).Where(x => !x.Contains(NodejsConstants.NodeModulesFolder));
-            this.DiscoverTests(fileList, testFx, logger, discoverySink);
+            this.DiscoverTests(fileList, testFx, logger, discoverySink, testDiscovererName);
         }
 
-        public void DiscoverTests(IEnumerable<string> fileList, TestFramework testFx, IMessageLogger logger, ITestCaseDiscoverySink discoverySink)
+        public void DiscoverTests(IEnumerable<string> fileList, TestFramework testFx, IMessageLogger logger, ITestCaseDiscoverySink discoverySink, string testDiscovererName)
         {
+            // If it's a framework name we support, is safe to submit the string to telemetry.
+            if (supportedFrameworks.Contains(testFx.Name))
```
amcasey, line 50: It might be interesting to report "other" if it's not a supported framework, so we get a sense of how common that is.

amcasey (approved):
LGTM.  Thanks for the updates!

Nodejs/Product/TestAdapter/TestDiscovererWorker.cs
armanio123: Good idea. Added,

PR #2303, opened 2020-11-19
Added unit tests for Nodejs projects test adapters.

armanio123:
Added some basic tests for all supported Test Adapters.

Running all tests takes about 30 seconds so far. My plan is to eventually add them to the build pipeline as a requirement for merging.

Please note that there's a small change on how we capture telemetry on the TestAdapters. No privacy impacted.

More test for the rest of the scenarios are coming. This PR is intended for an early review and discussion.


armanio123:
We most likely need to visit the open source compliance for the Moq package I'm using.

Nodejs/Tests/TestAdapter.Tests/TestProjectFactory.cs
```diff
@@ -0,0 +1,129 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using static Microsoft.NodejsTools.TestAdapter.TestFrameworks.TestFramework;
+
+namespace TestAdapter.Tests
+{
+    public class TestProjectFactory
```
uniqueiniquity, line 11: Does this really need to be a stateful class?
In other words, why not just make this class static and have the caller pass in `projectName` where needed?

Nodejs/Tests/TestAdapter.Tests/TestProjectFactory.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            switch (this.projectName)
+            {
+                case ProjectName.NodeAppWithTestsConfiguredOnProject:
+                case ProjectName.NodeAppWithTestsConfiguredPerFile:
+                case ProjectName.NodeAppWithAngularTests:
+                    return Path.Combine(this.GetProjectDirPath(), $"{projectNameString}.njsproj");
+            }
+
+            throw new ArgumentException();
```
uniqueiniquity, line 43: Add a message to this exception so it's clear what unexpected project name was observed?

Nodejs/Tests/TestAdapter.Tests/TestCaseFactory.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using Microsoft.NodejsTools;
+using Microsoft.NodejsTools.TestAdapter;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using static Microsoft.NodejsTools.TestAdapter.TestFrameworks.TestFramework;
+
+namespace TestAdapter.Tests
+{
+    public class TestCaseFactory
```
uniqueiniquity, line 10: Similar comment here to `TestProjectFactory` - I'd suggest making this static and introducing some sort of `TestCaseOptions` struct that can be passed to `GetTestCase`. At that point, you might even move `GetTestCase` into the other class as well.

Nodejs/Tests/TestAdapter.Tests/TestAdapter.Tests.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Build" Version="16.8.0" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="16.8.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.7.1" />
+    <PackageReference Include="Moq" Version="4.15.1" />
```
uniqueiniquity, line 13: Moq should be fine from a compliance perspective - the license is fine.

Nodejs/Tests/TestAdapter.Tests/TestAdapter.Tests.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <PackageReference Include="MSTest.TestAdapter" Version="2.1.1" />
+    <PackageReference Include="MSTest.TestFramework" Version="2.1.1" />
+    <PackageReference Include="coverlet.collector" Version="1.3.0" />
```
uniqueiniquity, line 16: Same with this package.

Nodejs/Tests/TestAdapter.Tests/TestAdapter.Tests.csproj
```diff
@@ -0,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
uniqueiniquity, line 4: Why not .NET 5? 😄

Nodejs/Tests/TestAdapter.Tests/TestProjectFactory.cs
armanio123: It doesn't need to be stateful. My first thought was to avoid sending multiple parameters but I don't mind making it static. I'll change it as suggested.

Nodejs/Tests/TestAdapter.Tests/TestProjectFactory.cs
armanio123: Added message and changed to `NotImplementedException` when a projectName has not been implemented instead.

Nodejs/Tests/TestAdapter.Tests/TestCaseFactory.cs
armanio123: Changed as suggested. Other than preference is there a reason behind changing it to static? I would admit I don't like very much to send to many parameters when a constructor can define them for all of the methods.

I don't mind changing it, just if there's something I'm missing.

uniqueiniquity: It just seems overly OOP-focused. Like... why is it a class? why do you call a method with no parameters and it spits out a test project? It feels much cleaner to me to have a function that clearly takes parameters as input and and returns the test cases generated from that input, especially since the class isn't meaningfully stateful.

I agree that it's clumsy to provide a bunch of parameters that are the same for each call here, which is why I thought it might be nice to have an `options` struct that holds them all together. That way, you can define the options up front, and hand it to each call as a single clean parameter.

Nodejs/Tests/TestAdapter.Tests/TestAdapter.Tests.csproj
armanio123: Oh! no reason I just didn't have it installed. Upgraded.

Nodejs/Tests/TestAdapter.Tests/TestProjectFactory.cs
```diff
@@ -0,0 +1,129 @@
...
+                this.GetProjectFilePath(),
+                configPath);
+
+            switch (testFramework)
+            {
+                case SupportedFramework.Jasmine:
+                    return new List<TestCase>()
+                    {
+                        testCaseFactory.GetTestCase("JasmineUnitTest.js::Test Suite 1::Test 1", "Test 1", 1, Path.Combine(this.GetProjectDirPath(), "JasmineUnitTest.js")),
+                        testCaseFactory.GetTestCase("JasmineUnitTest.js::Test Suite 1::Test 2", "Test 2", 1, Path.Combine(this.GetProjectDirPath(), "JasmineUnitTest.js")),
```
uniqueiniquity:
```suggestion
                        var filePath = Path.Combine(this.GetProjectDirPath(), "JasmineUnitTest.js");
                        testCaseFactory.GetTestCase("JasmineUnitTest.js::Test Suite 1::Test 1", "Test 1", 1, filePath),
                        testCaseFactory.GetTestCase("JasmineUnitTest.js::Test Suite 1::Test 2", "Test 2", 1, filePath)),
```

armanio123: Added suggestion. I had to add a local scope to the switch tho.

Nodejs/Tests/TestAdapter.Tests/DefaultTestDiscovererTests.cs
```diff
@@ -0,0 +1,130 @@
...
+
+            testCaseDiscoverySink.Setup(x => x.SendTestCase(It.IsAny<TestCase>()))
+                .Callback<TestCase>(x => actual.Add(x));
+
+            var testSource = TestProjectFactory.GetProjectFilePath(projectName);
+            var sources = new List<string>() { testSource };
+
+            AssureNodeModules(TestProjectFactory.GetProjectDirPath(projectName));
+
+            var testDiscoverer = new DefaultTestDiscoverer();
```
uniqueiniquity, line 94: Out of curiosity, why was this type public?
I would expect that none of the types in the test adapter assembly are public, so IVT would be needed for the tests.

uniqueiniquity (approved):

Nodejs/Tests/TestAdapter.Tests/DefaultTestDiscovererTests.cs
armanio123: I'm not sure I understand the question. You mean why `DefaultTestDiscoverer` is public? It implements `ITestDiscoverer` which is public as well, since I'm using it as well for testing, it make sense to make it public.

uniqueiniquity: Yeah, that was my question. I thought `ITestDiscoverer` wasn't our type, and even if it was, that doesn't mean the implementation needs to be public too.

My concern is that we have a public API here for no reason that someone could somehow take a dependency on.
We can use IVT to access internal types for testing purposes.

uniqueiniquity: Of course, I could be totally wrong about how our `ITestDiscoverer` implementations are invoked, in that they might have to be public for some reason I don't know about.

armanio123: You are correct, `ITestDiscoverer` is not our interface we just implement it and we don't explicitly require to be public nor the methods, afaik. The only reason we have it public is for the test project since we're referencing the class in it.

In regards to IVT, pardon my ignorance but what is that and how making it private/internal will allow us to access it?

armanio123:
I'll merge the PR and address any comments on a new one.

PR #2304, opened 2020-11-24
fix: updated typings causing build errors in templates

EricCornelson:
The latest node types have the result of `server.address()` as a `string | AddressInfo`, which was causing a build error on the default template. Updated the templates to fix this, and also standardized the imports.

### DIFF
--- a/Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureExpressApp/app.ts
+++ b/Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureExpressApp/app.ts
@@ -1,6 +1,7 @@
-import debug = require('debug');
-import express = require('express');
-import path = require('path');
+import * as debug from 'debug';
+import * as express from 'express';
+import { AddressInfo } from "net";
+import * as path from 'path';
 
 import routes from './routes/index';
 import users from './routes/user';
@@ -50,5 +51,5 @@ app.use((err, req, res, next) => { // eslint-disable-line @typescript-eslint/no-
 app.set('port', process.env.PORT || 3000);
 
 const server = app.listen(app.get('port'), function () {
-    debug('Express server listening on port ' + server.address().port);
+    debug('Express server listening on port ' + (server.address() as AddressInfo).port);
 });
--- a/Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureExpressApp/app.ts
+++ b/Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureExpressApp/app.ts
@@ -1,4 +1,3 @@
-import * as debug from 'debug';
 import * as express from 'express';
 import { AddressInfo } from "net";
 import * as path from 'path';
@@ -6,6 +5,7 @@ import * as path from 'path';
 import routes from './routes/index';
 import users from './routes/user';
 
+const debug = require('debug')('my express app');
 const app = express();
 
 // view engine setup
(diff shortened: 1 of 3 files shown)


Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureExpressApp/app.ts
```diff
@@ -50,5 +51,5 @@ app.use((err, req, res, next) => { // eslint-disable-line @typescript-eslint/no-
 app.set('port', process.env.PORT || 3000);
 
 const server = app.listen(app.get('port'), function () {
-    debug('Express server listening on port ' + server.address().port);
+    debug('Express server listening on port ' + (server.address() as AddressInfo).port);
```
amcasey, line 54: If this were DT, I'd point out that `import * as debug` does not result in a callable `debug`.

amcasey, line 54: How do we know it's not a `string`?

Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptExpressApp/app.ts
```diff
@@ -50,5 +51,5 @@ app.use((err, req, res, next) => { // eslint-disable-line @typescript-eslint/no-
 app.set('port', process.env.PORT || 3000);
 
 const server = app.listen(app.get('port'), function () {
-    debug('Express server listening on port ' + server.address().port);
+    debug('Express server listening on port ' + (server.address() as AddressInfo).port);
```
jessetrinity, line 54:
```suggestion
    debug(`Express server listening on port ${(server.address() as AddressInfo).port}`);
```

jessetrinity: This is admittedly not super helpful but since we offer this refactor anyway...

Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureExpressApp/app.ts
EricCornelson: good catch!

EricCornelson: [server.address()](https://nodejs.org/api/net.html#net_server_address) will only return a string if the server is listening on a pipe or Unix domain socket, which (at least in the unmodified template) will never be true.

Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureExpressApp/app.ts
```diff
@@ -1,10 +1,11 @@
...
-import express = require('express');
-import path = require('path');
+import * as express from 'express';
+import { AddressInfo } from "net";
+import * as path from 'path';
 
 import routes from './routes/index';
 import users from './routes/user';
 
+const debug = require('debug')('my express app');
```
amcasey, line 8: What does the `('my express app')` do?  It doesn't seem to have been in the original.

amcasey (approved):
I'm not sure I understand what we're gaining from the import changes, but this looks generally correct.

Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureExpressApp/app.ts
EricCornelson: So [debug](https://www.npmjs.com/package/debug) doesn't immediately give you a logging function, it gives you a function that takes a string and *returns* a logging function that, when called, will write to std.err with that string as a tag:
image

Previously, the template code was never `debug` with the initial tag string, so the output never actually got logged.

This should actually be updated in the JavaScript version of the template as well, now that I think of it.

amcasey: Wacky

PR #2305, opened 2020-12-01
Added execution tests to test adapters

armanio123:
Added execution tests to test adapters. Only nodejstools scenarios are covered.

Note that one of the added tests is always going to until a regression with Jasmine and a test failure with jest are fixed. Probably we want to `[Ignore]` this test until fixed, but I have decided to leave it broken so that we don't forget about it.


joj:
@armanio123 do we still want this?

armanio123:
I still think is a good idea to test this, but I'm unsure I will have the time to fix any failures. It's been some time since the PR was created maybe is no longer valid.

joj:
Closing to avoid maintenance burden, we can reinstate if we decide to do this at some point.

PR #2306, opened 2020-12-01
Add sourcelink support

jessetrinity:
We would like to add the link to the sourcefiles to the .pdb files we upload to the symbol servers. This follows the instructions here:
https://github.com/MicrosoftDocs/azure-devops-docs/blob/master/docs/artifacts/symbols/setting-up-github-sourcelinking.md

I'm not sure who will want to review this - please tag as appropriate :)

### DIFF
--- a/Nodejs/Product/Nodejs/Nodejs.csproj
+++ b/Nodejs/Product/Nodejs/Nodejs.csproj
@@ -1245,6 +1245,10 @@
     <PackageReference Include="VSLangProj80">
       <Version>8.0.50728</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.SourceLink.GitHub">
+      <Version>1.0.0</Version>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
   </ItemGroup>
   <ItemGroup>
     <ZipItem Include="Templates\Files\TypeScriptJasmineUnitTest\UnitTest.ts" />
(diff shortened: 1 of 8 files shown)


Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -1245,6 +1245,10 @@
     <PackageReference Include="VSLangProj80">
       <Version>8.0.50728</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.SourceLink.GitHub">
```
armanio123, line 1248: Should the reference be added to all projects? For example TestAdapter.csproj or Npm.csproj

jessetrinity: It should be added to anything we would expect to step into from other code. I think test assemblies are the classic example of what shouldn't be included, but everything else is probably fair game.

I guess we should also be publishing the pdbs for these? I think we currently only archive `NodeJsTools`.

jessetrinity: In the current build I see
`NodejsTools.pdb`
`NodejsTools.Npm.pdb`
`NodejsTools.PressAnyKey.pdb`
`NodejsTools.ProjectWizard.pdb`
`NodejsTools.TestAdapter.pdb`
`NodejsTools.TestAdapterImpl.pdb`
`JavaScript.TestAdapter.pdb`
`NodejsTools.WebRole.pdb`

Do you think it's reasonable to index and publish symbols for all of these?

armanio123 (approved):

PR #2318, opened 2021-02-25
Remove stale Nuget feeds and internal dependencies

minestarks:
* Remove all sources from nuget.config but nuget.org
* Remove internal dependency `Microsoft.VisualStudio.AppDesigner`
* Revert sections of #1719 which introduced a dependency on `PropPageBase` from the internal assembly. This restores the previous custom implementation of the property page.
* Revert sections of #1754 which introduced a dependency on `PropPageUserControlBase` from the internal assembly. This restores the previous implementation of File/Folder browse dialogs.

### DIFF
--- a/Nodejs/Product/Nodejs/Project/NodejsGeneralPropertyPage.cs
+++ b/Nodejs/Product/Nodejs/Project/NodejsGeneralPropertyPage.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
 using System.Runtime.InteropServices;
@@ -18,14 +18,7 @@ namespace Microsoft.NodejsTools.Project
             this.control = new NodejsGeneralPropertyPageControl(this);
         }
 
-        protected override Control CreateControl()
-        {
-            return this.control;
-        }
-
         public override Control Control => this.control;
-
-        protected override Type ControlType => typeof(NodejsGeneralPropertyPageControl);
 
         internal override CommonProjectNode Project
         {
@@ -46,8 +39,7 @@ namespace Microsoft.NodejsTools.Project
                 }
             }
         }
-
-        protected override void Apply()
+        public override void Apply()
         {
             this.Project.SetProjectProperty(CommonConstants.StartupFile, this.control.ScriptFile);
             this.Project.SetProjectProperty(NodeProjectProperty.ScriptArguments, this.control.ScriptArguments);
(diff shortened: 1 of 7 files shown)


uniqueiniquity (approved):

Nodejs/Product/Nodejs/Project/NodejsGeneralPropertyPage.cs
```diff
@@ -95,32 +87,40 @@ protected override void Apply()
...
-            this.control.WorkingDirectory = this.Project.GetUnevaluatedProperty(CommonConstants.WorkingDirectory);
-            this.control.LaunchUrl = this.Project.GetUnevaluatedProperty(NodeProjectProperty.LaunchUrl);
-            this.control.NodejsPort = this.Project.GetUnevaluatedProperty(NodeProjectProperty.NodejsPort);
-            this.control.DebuggerPort = this.Project.GetUnevaluatedProperty(NodeProjectProperty.DebuggerPort);
-            this.control.Environment = this.Project.GetUnevaluatedProperty(NodeProjectProperty.Environment);
-            this.control.StartWebBrowser = GetBoolProperty(this.Project, NodeProjectProperty.StartWebBrowser);
-            this.control.SaveNodeSettingsInProject = GetBoolProperty(this.Project, NodeProjectProperty.SaveNodeJsSettingsInProjectFile);
-            this.control.TestRoot = this.Project.GetProjectProperty(NodeProjectProperty.TestRoot);
-            this.control.TestFramework = this.Project.GetProjectProperty(NodeProjectProperty.TestFramework);
+            this.Loading = true;
```
amcasey, line 95: >this.Loading = true; [](start = 12, length = 20)

Where is this used?

amcasey (approved):
I'm assuming you confirmed that the property page(s) in question still work(s).

Nodejs/Product/Nodejs/Project/NodejsGeneralPropertyPage.cs
minestarks: There seems to be one reference to it in the base class.

https://github.com/microsoft/nodejstools/blob/0e0caef24ae5dcd9b76fbe57c617ec0ca9741ed4/Nodejs/Product/Nodejs/SharedProject/CommonPropertyPage.cs#L250

amcasey: In that case, is it thread-safe?

minestarks: Yes. VS UI controls all have UI thread affinity as far as I know. Confirmed via debugger.

amcasey: I feel like I'm missing something.  Does this method yield the UI thread so that another caller can check `IsDirty` on the UI thread?  If it just holds the UI thread until loading's done, the flag seems redundant?

minestarks: Yes, there's one thing that's not obvious from reading: setting the properties on the `control` is what causes the `IsDirty` setter to get called. This flag just seems to be for preventing excessive `OnStatusChanged` notifications from being sent as you initialize the page.

amcasey (approved):

PR #2347, opened 2021-09-03
Adding configuration to run React tests on Jest.

safacero:
We were facing a problem on JSPS using Jest programmatically without these configurations and the addition of the NODE_ENV environment variable. This may have been caused for some update on Jest and it may be good to think in the future on using `jest.config.js` if there is one on the project folder.

### DIFF
--- a/Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
+++ b/Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
@@ -37,8 +37,10 @@ const find_tests = function (testFileList, discoverResultFile, projectFolder) {
 };
 
 const run_tests = function (context) {
+    let projectFolder = context.testCases[0].projectFolder;
+    process.env.NODE_ENV =  process.env.NODE_ENV || 'test';
     return new Promise(async resolve => {
-        const jest = detectPackage(context.testCases[0].projectFolder, 'jest');
+        const jest = detectPackage(projectFolder, 'jest');
         if (!jest) {
             return resolve();
         }
@@ -51,14 +53,15 @@ const run_tests = function (context) {
             });
         }
 
-        const config = {
+        let projectFolderJsonReady = projectFolder.replaceAll('\\', '\\\\');
+        const argv = {
             json: true,
             reporters: [[__dirname + '/jestReporter.js', { context }]],
-            testMatch: [context.testCases[0].testFile]
-        };
+            config : `{"roots":["<rootDir>/src"],"collectCoverageFrom":["src/**/*.{js,jsx,ts,tsx}","!src/**/*.d.ts"],"setupFiles":["${projectFolderJsonReady}\\\\node_modules\\\\react-app-polyfill\\\\jsdom.js"],"setupFilesAfterEnv":["<rootDir>/src/setupTests.js"],"testMatch":["<rootDir>/src/**/__tests__/**/*.{js,jsx,ts,tsx}","<rootDir>/src/**/*.{spec,test}.{js,jsx,ts,tsx}"],"testEnvironment":"jsdom","testRunner":"${projectFolderJsonReady}\\\\node_modules\\\\jest-circus\\\\runner.js","transform":{"^.+\\\\.(js|jsx|mjs|cjs|ts|tsx)$":"${projectFolderJsonReady}\\\\node_modules\\\\react-scripts\\\\config\\\\jest\\\\babelTransform.js","^.+\\\\.css$":"${projectFolderJsonReady}\\\\node_modules\\\\react-scripts\\\\config\\\\jest\\\\cssTransform.js","^(?!.*\\\\.(js|jsx|mjs|cjs|ts|tsx|css|json)$)":"${projectFolderJsonReady}\\\\node_modules\\\\react-scripts\\\\config\\\\jest\\\\fileTransform.js"},"transformIgnorePatterns":["[/\\\\\\\\]node_modules[/\\\\\\\\].+\\\\.(js|jsx|mjs|cjs|ts|tsx)$","^.+\\\\.module\\\\.(css|sass|scss)$"],"modulePaths":[],"moduleNameMapper":{"^react-native$":"react-native-web","^.+\\\\.module\\\\.(css|sass|scss)$":"identity-obj-proxy"},"moduleFileExtensions":["web.js","js","web.ts","ts","web.tsx","tsx","json","web.jsx","jsx","node"],"watchPlugins":["jest-watch-typeahead/filename","jest-watch-typeahead/testname"],"resetMocks":true,"rootDir":"${projectFolderJsonReady}"}`,
+        }
 
         try {
-            await jest.runCLI(config, [context.testCases[0].projectFolder]);
+            await jest.runCLI(argv, [projectFolder]);
         } catch (error) {
             logError(error);
         }


Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -37,8 +37,10 @@ const find_tests = function (testFileList, discoverResultFile, projectFolder) {
 };
 
 const run_tests = function (context) {
+    let projectFolder = context.testCases[0].projectFolder;
```
PranavSenthilnathan, line 40: These can be `const`.

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -37,8 +37,10 @@ const find_tests = function (testFileList, discoverResultFile, projectFolder) {
...
+    process.env.NODE_ENV =  process.env.NODE_ENV || 'test';
```
PranavSenthilnathan, line 41: What does this do? Could add a comment here.

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -51,14 +53,15 @@ const run_tests = function (context) {
...
...dom.js"],"setupFilesAfterEnv":["<rootDir>/src/setupTests.js"],"testMatch":["<rootDir>/src/**/__tests__/**/*.{js,jsx,ts,tsx}","<rootDir>/src/**/*.{spec,test}.{js,jsx,ts,tsx}"],"testEnvironment":"jsdom","testRunner":"${projectFolderJsonReady}\\\\node_modules\\\\jest-circus\\\\runner.js","transform":{"^.+\\\\.(js|jsx|mjs|cjs|ts|tsx)$":"${projectFolderJsonReady}\\\\node_modules\\\\react-scripts\\\\config\\\\jest\\\\babelTransform.js","^.+\\\\.css$":"${projectFolderJsonReady}\\\\node_modules\\\\react-scripts\\\\config\\\\jest\\\\cssTransform.js","^(?!.*\\\\.(js|jsx|mjs|cjs|ts|tsx|css|json)$)":"${projectFolderJsonReady}\\\\node_modules\\\\react-scripts\\\\config\\\\jest\\\\fileTransform.js"},"transformIgnorePatterns":["[/\\\\\\\\]node_modules[/\\\\\\\\].+\\\\.(js|jsx|mjs|cjs|ts|tsx)$","^.+\\\\.module\\\\.(css|sass|scss)$"],"modulePaths":[],"moduleNameMapper":{"^react-native$":"react-native-web","^.+\\\\.module\\\\.(css|sass|scss)$":"identity-obj-proxy"},"moduleFileExtensions":["web.js","js","web.ts","ts","web.tsx","tsx","json","web.jsx","jsx","node"],"watchPlugins":["jest-watch-typeahead/filename","jest-watch-typeahead/testname"],"resetMocks":true,"rootDir":"${projectFolderJsonReady}"}`,
```
PranavSenthilnathan, line 60: This is a little hard to read. The usual approach to this is to first create the json object then stringify it:

```
const config = { ... };
const argv = {
  ...
  config: JSON.stringify(config)
};
```

As a bonus this lets you get rid of the `replaceAll`.

PranavSenthilnathan: We might have to find a minimal set of options. It doesn't look like some of these should have to be here. Also looking at the paths, I see a lot of "/src/" folders - while this might be a convention, I don't think this is a required project structure for react tests to work (what if I structured my code to be a "source/" directory instead for example).

PranavSenthilnathan (changes requested):

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -51,14 +53,15 @@ const run_tests = function (context) {
             });
         }
 
-        const config = {
+        let projectFolderJsonReady = projectFolder.replaceAll('\\', '\\\\');
+        const argv = {
             json: true,
             reporters: [[__dirname + '/jestReporter.js', { context }]],
-            testMatch: [context.testCases[0].testFile]
```
armanio123, line 57: I believe we launch a run for each file, so removing the `context.testCases[0].testFile` and instead adding all found tests will cause to run them multiple times. 

I agree running this only once will be a better approach, but some changes on NTVS will be required for that.

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
armanio123, line 60: Consider that this options might change if a configuration is present. Most likely we'll need to check of such configuration and set up accordingly.

armanio123 (changes requested):

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
safacero: All the information I found is that we can use this variable to add more log on development or test mode and less logging on production mode, so it's quicker, but I'm unsure how Node handles this internally, still digging! I tested changing the value of this variable and see if there was any time improvement running the test, no significant changes.

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -51,14 +56,20 @@ const run_tests = function (context) {
             });
         }
 
-        const config = {
+        let config = {
+            setupFilesAfterEnv:["<rootDir>/src/setupTests.js"],
```
safacero, line 60: Unsure why, but if I remove this line I get the following:
image

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
safacero: Yep, just realized I was adding all of them! Changed it to the test file but inside the config variable.

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
safacero: I'm a little unsure what to do with the "/src", I agree with @armanio123 that we need to suit this accordingly if there's already a config file on the project's folder, but how can we handle the fact that the user can change the name of the folder and maybe not have a config file? (Or am I understanding it wrong? That can also be the case :D ) Because from my tests creating a React app, it does create the folder "/src".

safacero: @armanio123 just added this support in the new changes!! Please review when you can :D

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -116,5 +123,36 @@ function logError() {
     console.error.apply(console, errorArgs);
 }
 
+function readConfigs(projectFolder, context)
+{
+    // First look for Jest.config.js, otherwise look at the package.json under "jest" tag
+    var userConfig;
```
armanio123, line 129: Use `let` instead.

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -116,5 +123,36 @@ function logError() {
...
+    {
+        userConfig = require(packageJsonPath).jest;
+        mergeConfigs();
+        return userConfig;
+    }
+
+    function mergeConfigs()
+    {
+        // If no config was found OR the user doesn't have these tags set up, add it.
+        if(!userConfig) userConfig = {};
```
armanio123, line 150: You could also use the logical assignment `||=`. It's supported since node 15.14.0, but I don't remember how far back do we support as well.

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -116,5 +123,36 @@ function logError() {
     console.error.apply(console, errorArgs);
 }
 
+function readConfigs(projectFolder, context)
+{
+    // First look for Jest.config.js, otherwise look at the package.json under "jest" tag
+    var userConfig;
+    const jestConfigPath = projectFolder + "\\jest.config.js";
```
armanio123, line 130: I don't remember, how will this work if we have multiple projects in one solution? Do we get a new `projectFolder` for each project?

armanio123: Should `jest.config.ts` be considered as well?

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -116,5 +123,36 @@ function logError() {
...
+    const packageJsonPath = projectFolder + "\\package.json";
+
+    if(fs.existsSync(jestConfigPath))
```
armanio123, line 133: I'm guessing the `jest.conf.js` file takes precedence over `package.json` by default. Is it the same behavior with plane Jest?

I mean, does it merge both configs if `package.json` and `jest.conf.js` have settings?

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -116,5 +123,36 @@ function logError() {
...
+        userConfig = require(packageJsonPath).jest;
+        mergeConfigs();
+        return userConfig;
+    }
+
+    function mergeConfigs()
+    {
+        // If no config was found OR the user doesn't have these tags set up, add it.
+        if(!userConfig) userConfig = {};
+        if(!userConfig.setupFilesAfterEnv) userConfig.setupFilesAfterEnv = ["<rootDir>/src/setupTests.js"];
```
armanio123, line 151: Seems like this configurations are the defaults used by React rather than the ones used by Jest. 

I'm not sure if this is the correct approach for NTVS as assuming a user is creating a plain JS project and adding jest support, this configurations would be incorrect for him.

armanio123 (changes requested):

Nodejs/Product/TestAdapter/TestFrameworkFactory.cs
```diff
@@ -91,18 +91,6 @@ private static (string, string)? GetTestFrameworkAndFilePath(Project project, Pr
                 fileAbsolutePath = CommonUtils.GetAbsoluteFilePath(projectRoot, projectItem.EvaluatedInclude);
             }
 
-            // Check if file is a typecript file. If so, get the javascript file. The javascript file needs to be in the same path and name.
-            // It doesn't work with bundlers or minimizers. Also, project needs to be build in order to have the js file created.
-            var typeScriptTest = TypeScriptHelpers.IsTypeScriptFile(fileAbsolutePath);
```
safacero, line 96: @armanio123 can you confirm to me that all this is doing is getting the .js and why that was a decision back then? I'm passing the .ts and .tsx now, do you think removing this code will cause other issues?

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
safacero: After new changes, we're not configuring this by hand anymore, we're letting `react-scripts` give us the config and, if not, adding the one from the user or passing just the `testMatch` config that we were doing previously.

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -50,15 +56,17 @@ const run_tests = function (context) {
...
-        const config = {
+        
+        let config = readConfigs(projectFolder, context);
+        
+        const argv = {
             json: true,
             reporters: [[__dirname + '/jestReporter.js', { context }]],
-            testMatch: [context.testCases[0].testFile]
-        };
+            config : JSON.stringify(config)
```
PranavSenthilnathan, line 65: I'm assuming putting `testMatch` in the config has the same behavior as having it outside (like before)?

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -110,6 +118,46 @@ function detectPackage(projectFolder, packageName) {
...
+    }
+
+    // Else, for other frameworks, first prioritize package.json config if jest config exists there under "jest".
+    // If not, try to find jest.config.js. If no file found, finally use a basic config.
+    if(fs.existsSync(packageJsonPath))
+    {
+        config = require(packageJsonPath).jest;
+    }
+
+    if(fs.existsSync(jestConfigPath) && !config)
```
PranavSenthilnathan, line 150:
```suggestion
    if(!config && fs.existsSync(jestConfigPath))
```

PranavSenthilnathan (changes requested):

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -110,6 +118,45 @@ function detectPackage(projectFolder, packageName) {
...
+    // It already deals with override from package.json config, if there's any.
+    const pathToReactConfig = path.join(projectFolder, 'node_modules/react-scripts/scripts/utils/createJestConfig.js');
+    if(fs.existsSync(pathToReactConfig))
+    {
+        const createJestConfig = require(pathToReactConfig);
+        config = createJestConfig(
+            relativePath => path.join(projectFolder, 'node_modules/react-scripts/', relativePath),
+            projectFolder,
+            false
+        )
```
armanio123, line 137: nit: semicolon missing.

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -110,6 +118,45 @@ function detectPackage(projectFolder, packageName) {
...
+    {
+        config = require(packageJsonPath).jest;
+    }
+
+    if(!config && fs.existsSync(jestConfigPath))
+    {
+        config = require(jestConfigPath);
+    }
+
+    if(!config) config = {};
```
armanio123, line 154: nit: You can use the logical OR assignment (`||=`). In node is supported since version 15.14.0, although I don't remember which version of Node we support.

Nodejs/Product/TestAdapter/TestFrameworkFactory.cs
armanio123: I think this will affect other test frameworks that are not Jest as they depend of this for typescript support. As an option, we might be able to filter out for Jest only.

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
```diff
@@ -37,8 +37,14 @@ const find_tests = function (testFileList, discoverResultFile, projectFolder) {
 };
 
 const run_tests = function (context) {
+    const projectFolder = context.testCases[0].projectFolder;
+
+    // NODE_ENV sets the environment context for Node, it can be development, production or test and it needs to be set up for jest to work.
+    // If no value was assigned, assign test.
+    process.env.NODE_ENV =  process.env.NODE_ENV || 'test';
```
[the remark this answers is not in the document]
safacero: The test continues according to my test!

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
safacero: jest.config.ts failed on the tests because apparently it changes the way we can read the configs depending on the TS version. For now, if we think it's a good thing I suggest creating a new work item for it! (:

Nodejs/Tests/TestAdapter.Tests/TestProjectFactory.cs
```diff
@@ -17,6 +17,8 @@ public enum ProjectName
             NodeAppWithTestsConfiguredOnProject,
             NodeAppWithTestsConfiguredPerFile,
             NodeAppWithAngularTests,
+            reactappwithjesttestsjavascript,
```
armanio123, line 20: I would recommend adding a comment for why is this lowercase. Somebody in the future might try to change it.

Nodejs/Tests/TestAdapter.Tests/TestProjectFactory.cs
```diff
@@ -102,12 +113,32 @@ private static List<TestCase> GetTestCases(ProjectName projectName, SupportedFra
...
+                            var filePath = Path.Combine(GetProjectDirPath(projectName), "src", "App.test.tsx");
+                            return new List<TestCase>()
+                            {
+                                GetTestCase(testCaseOptions, "src\\App.test.tsx::global::renders learn react link", "renders learn react link", 6, filePath)
+                            };
+                        }
+                        else
+                        {
+                            var filePath = Path.Combine(GetProjectDirPath(projectName), "JestUnitTest.js");
+                            return new List<TestCase>()
```
armanio123, line 136: nit: Formatting

armanio123 (approved):
Tests are passing manually but failing the build due to:  
```
NU1201: Project Microsoft.NodejsTools.TestAdapter is not compatible with net5.0 (.NETCoreApp,Version=v5.0). Project Microsoft.NodejsTools.TestAdapter supports: net472 (.NETFramework,Version=v4.7.2)
```

We can address this on another PR.

Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
armanio123: A new work item for me makes sense. Lets merge this one first.

Nodejs/Product/Nodejs/Nodejs.csproj
```diff
@@ -1148,7 +1148,7 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Newtonsoft.Json">
-      <Version>12.0.2</Version>
+      <Version>13.0.0</Version>
```
PranavSenthilnathan, line 1151: It looks like some places we use 13.0.1 and some 13.0.0. I think VS updated and then reverted versions.. can you update these to the latest version? I think the latest VS has 13.0.0 again but you should double check.

Nodejs/Product/TestAdapter/TestFrameworkFactory.cs
```diff
@@ -93,14 +93,18 @@ private static (string, string)? GetTestFrameworkAndFilePath(Project project, Pr
 
             // Check if file is a typecript file. If so, get the javascript file. The javascript file needs to be in the same path and name.
             // It doesn't work with bundlers or minimizers. Also, project needs to be build in order to have the js file created.
-            var typeScriptTest = TypeScriptHelpers.IsTypeScriptFile(fileAbsolutePath);
-            if (typeScriptTest)
+            // Note: for Jest we're just passing the .ts or .tsx, no need for .js
+            if(testFrameworkName != "Jest")
```
PranavSenthilnathan, line 97:
```suggestion
            if (testFrameworkName != "Jest")
```

PR #2350, opened 2021-11-30
Fixing perf issue in TestFilesUpdateWatcher

EricCornelson:
Currently, we use JTF.Run() on each time we get a FilesChanged event, or a DirectoryChangedEx event. This blocks the current (threadpool) thread until the work is complete. This can cause threadpool starvation, and has done so for users, as evidenced by telemetry.

See Issue # 1393794 (Azure DevOps)

### DIFF
--- a/Nodejs/Product/TestAdapterImpl/TestFilesUpdateWatcher.cs
+++ b/Nodejs/Product/TestAdapterImpl/TestFilesUpdateWatcher.cs
@@ -4,6 +4,8 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Threading;
+using Microsoft.NodejsTools.Telemetry;
 using Microsoft.VisualStudio;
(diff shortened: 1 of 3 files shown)


armanio123 (approved):

Nodejs/Product/TestAdapterImpl/TestFilesUpdateWatcher.cs
```diff
@@ -156,17 +171,18 @@ private static WatcherChangeTypes ConvertVSFILECHANGEFLAGS(uint flag)
 
         private void CheckDisposed()
         {
-            if (this.disposed)
+            if (this.disposed == DISPOSED)
```
joj, line 174: Should this be IsDisposed?

EricCornelson: Probably should be, for consistency

PR #2353, opened 2022-01-24
Discover devenv process id without relying on environment

joj:
NTVS sets an _NTVS_PID variable on package initialize. JSPS doesn't do
that. Use the _NTVS_PID variable is present. If it isn't we'll search
the stack for a parentid of devenv instead. Search 8 processes app
(which is a bit extreme, but it's better to search a few extras). If
after 8 searches there is no devenv (or we fail to find a devenv process
at all), we return 0 and consider that the tests are not using VS.

### DIFF
--- a/Nodejs/Product/TestAdapter/TestExecutorWorker.cs
+++ b/Nodejs/Product/TestAdapter/TestExecutorWorker.cs
@@ -2,8 +2,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Management;
 using System.Net.NetworkInformation;
 using System.Runtime.InteropServices;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


Nodejs/Product/TestAdapter/TestExecutorWorker.cs
```diff
@@ -316,11 +318,47 @@ void RecordEnd(TestCaseResult test, ResultObject resultObject)
...
         }
 
+        private int GetParentProcessId(uint processId, int counter = 0)
+        {
+            var query = string.Format("SELECT ParentProcessId FROM Win32_Process WHERE ProcessId = {0}", processId);
+            var search = new ManagementObjectSearcher("root\\CIMV2", query);
+            var results = search.Get().GetEnumerator();
+            results.MoveNext();
+
+            var parentId = (uint)results.Current["ParentProcessId"];
```
armanio123, line 328: Could the cast throw an exception if there's no `ParentProcessId` or will always have a parent?

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
```diff
@@ -316,11 +318,47 @@ void RecordEnd(TestCaseResult test, ResultObject resultObject)
...
+
+            try
+            { 
+                var parentName = Process.GetProcessById((int)parentId).ProcessName;
+
+                if (parentName == "devenv")
+                    return (int)parentId;
+                if (counter < 8) //devenv.exe is usually the 4th process app, so don't search forever
+                    return GetParentProcessId(parentId, counter++);
```
armanio123, line 337: nit: I prefer to use a for loop as I feel is more readable. Feel free to ignore this anyway is just preference.

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
```diff
@@ -316,11 +318,47 @@ void RecordEnd(TestCaseResult test, ResultObject resultObject)
...
             var pid = Environment.GetEnvironmentVariable(NodejsConstants.NodeToolsProcessIdEnvironmentVariable);
-            return pid == null ? false : int.TryParse(pid, out processId);
+
+            if (pid != null)
+            {
+                return pid == null ? false : int.TryParse(pid, out processId);
+            }
+            else
+            {
+                processId = GetParentProcessId((uint)Process.GetCurrentProcess().Id);
```
armanio123, line 359: Could you add a comment stating the need for checking the process Id?

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
```diff
@@ -316,11 +318,47 @@ void RecordEnd(TestCaseResult test, ResultObject resultObject)
             }
         }
 
+        private int GetParentProcessId(uint processId, int counter = 0)
+        {
+            var query = string.Format("SELECT ParentProcessId FROM Win32_Process WHERE ProcessId = {0}", processId);
+            var search = new ManagementObjectSearcher("root\\CIMV2", query);
```
armanio123, line 324: I noticed on the documentation that `ManagementObjectSearcher` "cannot be used on by partially trusted code". Is ther a scenario where this might affect us?

armanio123 (changes requested):

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
joj: There will always be a parent, but I'm still making the try block include everything (commit coming) because that makes more sense.

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
joj: I feel you, totally, but I still think this is cleaner.

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
joj: I thought that including in the PR was enough, I will add the same comment I added here :)

Nodejs/Product/TestAdapter/TestExecutorWorker.cs
joj: I looked it up. Partially trusted code is a sandbox like UWP. I think that will never be the case with VS, and also this is going away on net5, so when VS upgrades it will no longer be an issue. ManagementObjectServer is available in net5 and net6, and CAS is not a constraint there.

armanio123 (approved):
Is there a test case added for JSPS for this regresion?

joj:
Nope, I'll create a task for it.

joj:
Reviewing the backlog, there is a test and a bug. I'll make sure that gets resolved.

PR #2354, opened 2022-01-26
Switched to launching using PineZorro

joj:
The new server protocol in vs-debug-js has issues, and when launching a
browser via the new server protocol, breakpoints are not set correctly.
Switched to launching using the debugger to avoid this issue while still
having debug support for the browser.


Shenniey (approved):
Looks good! 😄 The only thing I'd add is that we may want to consider how the logging works. I don't remember exactly, but I think right now all instances using WebDiagnostics/js-debug through PineZorro write to a single visualstudio-js-debugger.txt file, so I'm not sure if they're both appending to it or if only one manages to write to it and then we lose the logging for the other. 

Not super important but just something to keep in mind!

PR #2356, opened 2022-01-27
Support Jasmine 4.0

PranavSenthilnathan:
Jasmine 4.0 has two major breaking changes that affect us.

1) addSpecFiles is deprecated and addMatchingSpecFiles should be used instead

2) topSuite() now returns a metadata object containing the Suite instead of returning the Suite itself

In addition, the version update also exposed a bug where we call `loadSpecs` and don't await the result (previously this was a void returning function but now it's an async function).

### DIFF
--- a/Nodejs/Product/TestAdapter/TestFrameworks/Jasmine/jasmine.js
+++ b/Nodejs/Product/TestAdapter/TestFrameworks/Jasmine/jasmine.js
@@ -123,10 +123,10 @@ async function find_tests(testFileList, discoverResultFile, projectFolder) {
             jasmineInstance.specFiles = [];
             jasmineInstance.addMatchingSpecFiles([testFile]);
             
-			var p = jasmineInstance.loadSpecs();
-			if (p instanceof Promise) {
-				await p;
-			}
+            var p = jasmineInstance.loadSpecs();
+            if (p instanceof Promise) {
+                await p;
+            }
 
             var topSuite = jasmineInstance.env.topSuite();
             // In Jasmine 4.0+ the Suite object is not top level anymore and is instead in the suite_ property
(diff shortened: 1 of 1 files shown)


Nodejs/Product/TestAdapter/TestFrameworks/Jasmine/jasmine.js
```diff
@@ -107,38 +107,44 @@ function enumerateSpecs(suite, testList, testFile) {
...
+                await p;
             }
-        });
 
-        var fd = fs.openSync(discoverResultFile, 'w');
-        fs.writeSync(fd, JSON.stringify(testList));
-        fs.closeSync(fd);
+            var topSuite = jasmineInstance.env.topSuite();
+            // In Jasmine 4.0+ the Suite object is not top level anymore and is instead in the suite_ property
+            if (topSuite && topSuite.suite_) {
```
PranavSenthilnathan, line 133: Not sure about the idiomatic way of checking this, so I'm open to changing it.

Nodejs/Product/TestAdapter/TestFrameworks/Jasmine/jasmine.js
```diff
@@ -107,38 +107,44 @@ function enumerateSpecs(suite, testList, testFile) {
...
+    
+    var jasmineInstance = initializeJasmine(Jasmine, projectFolder);
+    setSpecFilter(jasmineInstance, _ => false);
+
+    var testList = [];
+    for (var testFile of testFileList.split(";")) {
+        try {
+            jasmineInstance.specDir = "";
+            jasmineInstance.specFiles = [];
+            jasmineInstance.addMatchingSpecFiles([testFile]);
```
armanio123, line 124: Will this create a compatibility issue with the previous version or this method also exists pre version 4?

noshlag (approved):
I think this looks fine.

Nodejs/Product/TestAdapter/TestFrameworks/Jasmine/jasmine.js
PranavSenthilnathan: Updated to check if it exists and fall back to the old function if it doesn't.

armanio123 (approved):

PR #2365, opened 2022-03-12
Refactor npm install --save command

armanio123:
The save command has been deprecated for some time now so the UI doesn't make sense anymore. 

I have refactored to instead remove the save checkbox and added the "no save" command on the combo box.

This also fixes an accessibility issue that the "(recommended)" text was not read by the screen reader, since I have removed it.


noshlag (approved):

minestarks:
@armanio123 could you post a screenshot of the changes? 

(Nit: "Refactor" implies just a code reorganization / no behavior change, so I was initially confused by the title)

armanio123:
This is how the npm install window looks like now:
image

The dependency type now contains "No Save" to avoid adding the dependency to `package.json`:
image

PR #2371, opened 2022-03-28
Removed ProgressBar and disabled options

armanio123:
Removed the Progress bar when installing a package from the npm window. Instead, now we disable the options and the install button and enable them back once the installation is complete.

The progress should still be visible from the background tasks area on Visual Studio. Details can still be seen on the npm output window.

This is how the window looks once the Install Package button has been clicked:
image


zkat (approved):
looks like this does what it says on the tin, so 👍🏻

PR #2382, opened 2022-04-11
stop depending on _requiredBy

zkat:
npm@7 removed this feature with its new resolver. This was always an internal
implementation detail and should have never been relied upon.

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1237993

### DIFF
--- a/Nodejs/Product/Npm/SPI/NodeModules.cs
+++ b/Nodejs/Product/Npm/SPI/NodeModules.cs
@@ -30,27 +30,15 @@ namespace Microsoft.NodejsTools.Npm.SPI
             {
                 Debug.Assert(this.allModules.Count == 0, "Depth is 0, but top-level modules have already been added.");
 
+                var rootJson = parent.PackageJson;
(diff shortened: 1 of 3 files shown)


joj (approved):

Nodejs/Product/Npm/SPI/NodeModules.cs
```diff
@@ -125,13 +113,11 @@ private bool AddModuleIfNotExists(IRootPackage parent, string moduleDir, bool sh
 
             IPackage package = moduleInfo.Package;
 
-            if (package == null || depth == 1 || !moduleInfo.RequiredBy.Contains(parent.Path))
+            if (package == null || depth == 1)
```
armanio123, line 116: What was the purpose in looking into the path? Isn't this the logic that tells the user that a package is a dependency but has not been installed yet?

zkat: no this seems to be more about being defensive about adding too many things to RequiredBy? It didn't seem to affect anything?

PR #2383, opened 2022-04-18
Ignore exceptions when unloading already unloaded projects

armanio123:
This issue happens during integration tests. If a project has already been unloaded, we ignore the exception and just continue to dispose.


noshlag (approved):
Looks good to me.  Maybe we should record some telemetry here to diagnose how often this is happening?

PR #2384, opened 2022-04-20
Hidding templates with very low usage to simplify selection

joj:


armanio123:
Why can't we remove them instead? Also, do we have any data that backs up they are low usage?

joj:
We will eventually remove them, but I wanted to keep them for a bit. We do have the data.

armanio123 (approved):

PR #2391, opened 2022-05-23
Removed legacy NTVS debugger

joj:


Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
```diff
@@ -84,14 +83,9 @@ public int LaunchFile(string file, bool debug)
             {
                 if (nodeVersion >= new Version(8, 0))
                 {
-                    StartWithChromeV2Debugger(file, nodePath, shouldStartBrowser, browserPath);
+                    StartWithChromeDebugger(file, nodePath, shouldStartBrowser, browserPath);
                     TelemetryHelper.LogDebuggingStarted("ChromeV2", nodeVersion.ToString());
                 }
-                else
```
EricCornelson, line 90: It looks like this will silently fail on node versions < 8.0. Should we show an error message or something?

EricCornelson (approved):

Nodejs/Product/Nodejs/Project/NodejsProjectLauncher.cs
joj: That's fair but I wouldn't block the merge on this. I'll create a work item to track it and add a message if needed.

joj: Usage count for Node6 in the log is 0.9%. I will add an error message to catch these few.

PR #2396, opened 2022-06-13
Add support for customizing Jest config with react-app-rewired

acostadelivra:
Loads the react-app-rewired library (if present) in order to customize the Jest config provided by react-scripts.  This allows customizing the config further than what react-scripts normally allows, without ejecting.

This is normally accomplished via the command line like "npx react-app-rewired test".  I think requiring the overrides/jest.js file is the most appropriate way to load it directly.

### DIFF
--- a/Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
+++ b/Nodejs/Product/TestAdapter/TestFrameworks/Jest/jest.js
@@ -129,6 +129,12 @@ function readConfigs(projectFolder, context)
     const pathToReactConfig = path.join(projectFolder, 'node_modules/react-scripts/scripts/utils/createJestConfig.js');
     if(fs.existsSync(pathToReactConfig))
     {
+        const pathToReactAppRewired = path.join(projectFolder, 'node_modules/react-app-rewired/overrides/jest.js');
+        if (fs.existsSync(pathToReactAppRewired)) {
+            // Support react-app-rewired.  This overrides the normal createJestConfig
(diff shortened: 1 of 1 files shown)


joj:
Hi! I'm trying to better understand your scenario. Is this just a convenient way of avoiding "npx react-app-rewired test" or is it required for what you're doing? 
The code in this repo is shipped with Visual Studio, an so we try to be as minimal as we can when running dependencies, for security reasons. I do understand this is just one more dependency and this file already has some, but this change may make react-app-rewired invisible in some scenarios, which is something we'd rather avoid. 
Thanks! J

acostadelivra:
We're trying to get our test results in Test Explorer rather than through the command line, so in that sense it's required.  There seems to be no way to override the react-scripts config in the current TestAdapter code as far as I can tell.  I completely understand the security concern though, and I think it can be worked around.

react-app-rewired just loads a 'config-overrides.js' file in the project root (similar to jest.config.js) and uses it to transform the react-scripts output.
Instead of requiring the react-app-rewired package, it should be simple enough to just apply the config-overrides without it.
I think it would still be appropriate to check for react-app-rewired's existence in node_modules in case a user just happens to have an unrelated config-overrides.js in their project root though.

Let me know if this sounds like a better plan and I'll go ahead implement it.

acostadelivra:
Sorry for the long delay, I lost track of this PR.  I've updated it so that react-app-rewired is never directly executed.  Instead just the user supplied config-overrides script is used.  This should be equivalent and less invasive.

PR #2402, opened 2022-06-28
General fixes on templates

joj:


minestarks (approved):

Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptVuejsApp/TypeScriptVuejsApp.vstemplate
```diff
@@ -37,7 +37,6 @@
         <ProjectItem>shims-vue.d.ts</ProjectItem>
       </Folder>
       <ProjectItem>babel.config.js</ProjectItem>
-      <ProjectItem>.browserslistrc</ProjectItem>
```
armanio123, line 40: What's the reason this file got removed? 
The template is based on the vue cli output, and I feel like we should only remove it if the cli changed it as well.

joj: browserslist is configuration. I don't think we should include anything that starts with a "." except for vscode stuff.

armanio123: Why do we shouldn't include it? Browserlist is used for setting up the supported browsers on other tools used on the project.

I'm on the position that we shouldn't modify the default templates, unless is necessary.

PR #2403, opened 2022-07-07
Fixing a bug after Jasmine update.

safacero:
Fixing the Jasmine bug after an update that moved the 'it()' function to another place in their internal code.

Issue #
Angular tests were not being found.

##### Bug
Jasmine made some internal changes and moved some logic related to the 'it()' function that we use to describe and run the tests. Internally, NTVS needs to replace Jasmine's 'it()' function and after their changes on update, this replacement was failing.

##### Fix
We found out on their code that there's an 'it_()' that seems to be the old 'it()'. After making this change, our tests were being found again.

##### Testing
The Angular test are now being found, tested locally.


safacero:
@joj @armanio123 I also made this change on TestAdapter.tests, it was needed for the project to build locally, but I was unsure if it was just on my machine or if I should push as well. Should I?
image

armanio123 (approved):
Thanks for working on this!

joj:
Build works locally and on CI, so we're ok :)

PR #2408, opened 2022-08-12
Add ESLint default configurations

MariaSolOs:
Adding the necessary linting dependencies and the most basic configurations to the project templates.

Note that I couldn't find all of the templates in the new project dialog, so I didn't update those.

### DIFF
--- a/Nodejs/Product/Nodejs/ProjectTemplates/AzureExpress4App/package.json
+++ b/Nodejs/Product/Nodejs/ProjectTemplates/AzureExpress4App/package.json
@@ -20,5 +20,12 @@
   },
   "engines": {
     "node": "~6.10.x"
+  },
+  "devDependencies": {
+    "eslint": "^8.21.0",
+    "eslint-plugin-node": "^11.1.0"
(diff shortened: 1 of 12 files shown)


Nodejs/Product/Nodejs/ProjectTemplates/AzureExpress4App/package.json
```diff
@@ -20,5 +20,10 @@
   },
   "engines": {
     "node": "~6.10.x"
+  },
+  "devDependencies": {
+    "eslint": "^8.21.0"
+  },
+  "eslintConfig": {
```
armanio123, line 27: What does an empty `eslintConfig` field means?

MariaSolOs: It means that we're explicitly defining an ESLint configuration (which will allow the server to pick it up and not prompt the user to run the plugin wizard) but not indicating the rules to use.

joj (approved):

PR #2411, opened 2022-10-25
Moved IVsFileChangeEx construction outside of the constructor

joj:
IVsFileChangeEx is expected to be requested from the main thread. This may not always be the case on the constructor, and it's resulting in failures while opening VS. Moving the call to first use (which will usually be AddFileWatch or AddFolderWatch) results in the service being requested always in the right context.

Issue #1640631

### DIFF
--- a/Nodejs/Product/TestAdapterImpl/TestFilesUpdateWatcher.cs
+++ b/Nodejs/Product/TestAdapterImpl/TestFilesUpdateWatcher.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.Threading;
 using Microsoft.NodejsTools.Telemetry;
 using Microsoft.VisualStudio;
+using Microsoft.VisualStudio.OLE.Interop;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel;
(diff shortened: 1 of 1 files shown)


Nodejs/Product/TestAdapterImpl/TestFilesUpdateWatcher.cs
```diff
@@ -181,19 +183,19 @@ private void CheckDisposed()
...
+                    this.fileWatcher.Value.UnadviseFileChange(cookie);
                 }
                 foreach (var cookie in this.watchedFolders.Values)
                 {
-                    this.fileWatcher.UnadviseDirChange(cookie);
+                    this.fileWatcher.Value.UnadviseDirChange(cookie);
                 }
                 this.watchedFiles.Clear();
                 this.watchedFolders.Clear();
-                this.fileWatcher = null;
```
PranavSenthilnathan, line 196: Might be worth keeping this null assignment just so we know the object isn't being used.

PranavSenthilnathan (approved):

Nodejs/Product/TestAdapterImpl/TestFilesUpdateWatcher.cs
joj: The Lazy can't be assigned to. We do have a DISPOSED variable and I'm checking for IsValueCreated before clearing things out to avoid creating the value just to clear it.

PR #2418, opened 2022-11-16
Added speech action to HelpListItem

armanio123:
Addes a speech action to each Item on the npm package list.

Fixes accessibility bug: A11y_When using down arrow to the list items screen reader is not announcing the information of the more data present beside the list items which is selected._JSTS in VS_Install New npm Packages_ScreenReader.

### DIFF
--- a/Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallWindow.xaml
+++ b/Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallWindow.xaml
@@ -44,10 +44,11 @@
                         <Setter Property="AutomationProperties.Name" Value="{Binding Path=Name}" />
                         <Setter Property="AutomationProperties.HelpText">
                             <Setter.Value>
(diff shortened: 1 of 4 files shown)


Nodejs/Product/Nodejs/NpmUI/NpmPackageInstallWindow.xaml
```diff
@@ -44,10 +44,11 @@
                         <Setter Property="AutomationProperties.Name" Value="{Binding Path=Name}" />
                         <Setter Property="AutomationProperties.HelpText">
                             <Setter.Value>
-                                <MultiBinding StringFormat="{} {0}, {1}, {2}">
+                                <MultiBinding StringFormat="{} {0}, {1}, {2}, {3}">
```
armanio123, line 47: Feels a bit weird to add this as a parameter binding instead of just the text but I didn't find another way to make it work.

PranavSenthilnathan (approved):

PR #2430, opened 2022-12-06
Moved Telemetry to be within the VS process instead of the test console

joj:
Telemetry right now is reported in the test console, which is running outside of devenv.exe and incapable of actually publishing telemetry. While I find what the proper pattern is for publishing telemetry from a test adapter, I'm moving the reporting to VS. The only item we're publishing is what test framework is being used on discovery.

Issue #

### DIFF
--- a/Nodejs/Product/TestAdapterImpl/TestContainerDiscoverer.cs
+++ b/Nodejs/Product/TestAdapterImpl/TestContainerDiscoverer.cs
@@ -6,6 +6,7 @@ using System.ComponentModel.Composition;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using Microsoft.Internal.VisualStudio.Shell;
 using Microsoft.NodejsTools.TypeScript;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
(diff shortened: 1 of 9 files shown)


Nodejs/Product/TestAdapterImpl/TestContainerDiscoverer.cs
```diff
@@ -289,6 +290,11 @@ public IEnumerable<ITestContainer> GetTestContainers(IVsProject project)
                     return latest;
                 });
 
+            if (!string.IsNullOrEmpty(testFramework))
+            { 
+                Microsoft.NodejsTools.Telemetry.TelemetryHelper.LogTestDiscoveryStarted(testFramework);
```
armanio123, line 295: Is this still on starting a discovery? if not, I would change the name of the log

armanio123 (approved):

Nodejs/Product/TestAdapterImpl/TestContainerDiscoverer.cs
joj: It's debatable :). This gets the test containers and that happens on discovery (execution already has them because of discovery). The most accurate would be "Container Discovery Started" but that's mostly an implementation detail. I prefer to have it as Discovery Started in case I can shift it to the right location in the container.

PR #2447, opened 2023-01-20
Prevent loading output panes on non-jsts projects

joj:
CMake projects are showing tsc and npm output windows because we are too eager in initializing those panes on any case where a file is opened while building (which is standard for CMake projects). I'm switching to only initialize if we're actually going to include information in the panes.

### DIFF
--- a/Nodejs/Product/Nodejs/Workspace/TypeScriptActionProviderFactory.cs
+++ b/Nodejs/Product/Nodejs/Workspace/TypeScriptActionProviderFactory.cs
@@ -60,8 +60,6 @@ namespace Microsoft.NodejsTools.Workspace
             {
                 await this.workspaceContext.JTF.SwitchToMainThreadAsync();
 
-                this.outputPane.InitializeOutputPanes();
-
                 var actions = new List<IFileContextAction>();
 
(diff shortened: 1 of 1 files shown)


minestarks:
I looked into whether we can disable this code from loading entirely, but I couldn't find a way to filter the MEF export even further.  This seems like the most practical approach.

More importantly though... does the feature that this file is supposed to implement ("build .ts file") even work? I dug into the code and it seems to rely on [tsc.exe](https://github.com/microsoft/nodejstools/blob/c56da573e15e51e201b9d6344e066b708b72177b/Nodejs/Product/Nodejs/TypeScriptHelpers/TypeScriptCompile.cs#L14), which is not a file that we ship anymore (we dropped support for it 4-5 years ago).

joj:
I honestly don't know. This pane will load when we're "building" (I use that term very lightly) and a ts file or tsconfig.json is shown. We may be able to just remove the pane but I'd rather keep it for now, only not loading when it's not needed (which, to be fair, may be always)

minestarks (approved):
As I mentioned I think we may want to remove this altogether at some point if it doesn't work anyway.

joj:
I'm thinking (based on that) that it may make sense to move this to Extras. I'll add it to the list :)

PR #2455, opened 2023-03-15
using MVC5 guid to differentiate between web and console app

Shenniey:


Shenniey:
Sorry! Accidental PR from wrong branch. Pls ignore.

PR #2467, opened 2023-05-29
fix angular test discovery

PranavSenthilnathan:
The test settings previously always used IPv4 for the listening endpoint. This meant in Node 17+ the tests would not run since node would probe with IPv6. This PR changes it to use localhost which makes the test use the same address that node would use. I tested this change with Node 16 and Node 18 and they both work. Since the change is only to the default values, the user can pass their own values and they will take precedence.

### DIFF
--- a/Nodejs/Product/TestAdapter/TestFrameworks/Angular/karmaConfig.js
+++ b/Nodejs/Product/TestAdapter/TestFrameworks/Angular/karmaConfig.js
@@ -19,6 +19,11 @@ module.exports = function (config) {
     // Replace all reporters
     config.reporters = ['vsKarmaReporter'];
 
+    if (!config.hostname && !config.listenAddress) {
+        // The default address is IPv4 but node 17+ uses IPv6. Update the value to 'localhost' (which works with both IPv4 and IPv6) if it was not set.
+        config.hostname = config.listenAddress = 'localhost';
+    }
+
     setGrep(config);
 };
 


armanio123:
Adding link for documentation: https://github.com/karma-runner/karma/issues/3730

Nodejs/Product/TestAdapter/TestFrameworks/Angular/karmaConfig.js
```diff
@@ -19,6 +19,11 @@ module.exports = function (config) {
     // Replace all reporters
     config.reporters = ['vsKarmaReporter'];
 
+    if (!config.hostname && !config.listenAddress) {
+        // The default address is IPv4 but node 17+ uses IPv6. Update the value to 'localhost' (which works with both IPv4 and IPv6) if it was not set.
+        config.hostname = config.listenAddress = 'localhost';
```
armanio123, line 24: Karma documentation states that `hostname` is by default `localhost`. Do we need to set it up anyway?

Nodejs/Product/TestAdapter/TestFrameworks/Angular/karmaConfig.js
```diff
@@ -19,6 +19,11 @@ module.exports = function (config) {
     // Replace all reporters
     config.reporters = ['vsKarmaReporter'];
 
+    if (!config.hostname && !config.listenAddress) {
```
armanio123, line 22: If `hostname` is set but `listenAddress` is not will the discovery fail? 

Shouldn't `listenAddress` be set all the time?

Nodejs/Product/TestAdapter/TestFrameworks/Angular/karmaConfig.js
PranavSenthilnathan: The test runner actually has a warning if listenAddress is set explicitly but hostname is not, so I just set them both explicitly to get rid of that warning.

Nodejs/Product/TestAdapter/TestFrameworks/Angular/karmaConfig.js
PranavSenthilnathan: Yes it was failing because the default listenAddress is ipv4. Node started using the ipv6 address so this mismatch caused an issue.

armanio123: I mean, with this change if a user sets the `hostname` jbut not the `listenAddress`, the default will never apply to `listenAddress` causing it to fail?

PranavSenthilnathan: Yes, my reasoning was that if the user left the default values for both then their intent was just `localhost` for both. If they changed either value then it was intentional and we won't change the behavior with our options (for example, the hostname could be some server). Not sure if this is a valid assumption..

joj (approved):

Nodejs/Product/TestAdapter/TestFrameworks/Angular/karmaConfig.js
armanio123: Sounds good, thanks for clarifying.

armanio123 (approved):

PR #2491, opened 2024-03-26
adding goldbar to prompt migration to JSPS

Shenniey:


Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            uint eventCookie = 2;
+            CurrentInfoBarElement = infoBarUiFactory.CreateInfoBar(infoBarModel);
+            CurrentInfoBarElement.Advise(new InfoBarUIEvents(OnClose), out eventCookie);
+
+            infoBarHost.AddInfoBar(CurrentInfoBarElement);
+
+            void OnClose()
+            {
+                CurrentInfoBarElement.Unadvise(eventCookie);
```
PranavSenthilnathan, line 34: If the user dismisses the gold bar we should probably not show it again the next time they open the project. We have some sample code in the JSPS repo (see SdkUpdatePrompter for how to use the user file).

Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            if (ErrorHandler.Failed(vsShell.GetProperty((int)__VSSPROPID7.VSSPROPID_MainWindowInfoBarHost, out var tmp)))
+            {
+                return;
+            }
+            var infoBarHost = (IVsInfoBarHost)tmp;
+
+            CurrentInfoBarElement?.Close();
+
+            var infoBarModel = new InfoBarModel(Resources.MigrateToJspsPrompt);
```
PranavSenthilnathan, line 24: It would be a nice user experience if there was a button on the gold bar that users can click that would do the migration instead of giving them the steps to the project context menu. I think that should be possible by just passing in an `Action` into this ctor that does it.

PranavSenthilnathan (changes requested):

PranavSenthilnathan (approved):
approving so this isn't blocked from merging

Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
Shenniey: I wasn't able to use the CPS tools in JSPS to read the XML file -- noting this down to fix in the next iteration.

Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
Shenniey: Also moving this to next iteration, and merging for now.

Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
Shenniey: Turns out we don't need it! The cookie event will prevent the goldbar from showing up again if the user dismisses it.

Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
Shenniey: Added in #2493 !

PR #2493, opened 2024-03-28
applying feedback on goldbar

Shenniey:
Applying feedback from #2491.

We don't need to use the user file b/c the cookie event ensures the goldbar from showing up again if the user dismisses it.

### DIFF
--- a/Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
+++ b/Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
@@ -1,4 +1,9 @@
 ﻿using System;
+using System.Drawing.Text;
+using System.IO;
+using System.Linq;
+using System.Xml.Linq;
+using Microsoft.NodejsTools.Commands;
 using Microsoft.VisualStudio;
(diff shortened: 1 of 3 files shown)


Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
```diff
@@ -33,6 +44,11 @@ void OnClose()
             {
                 CurrentInfoBarElement.Unadvise(eventCookie); 
             }
+
+            void MigrateProjectCommand()
+            {
+                NodejsPackage.Instance.DTE.Commands.Raise(Guids.MigrateToJspsCmdSet.ToString(), PkgCmdId.cmdidJspsProjectMigrate, null, null);
```
PranavSenthilnathan, line 50: I would have expected there to be a project id passed in somewhere, but since there isn't, I'm wondering if this will work with multiple projects? Would multiple goldbars show up (one for each project)?

Shenniey: Nope, just 1 goldbar

PranavSenthilnathan (approved):

PR #2500, opened 2024-08-13
Updated package.json for test projects for compliance reasons

joj:
Most of the mock projects are fairly old, and they were triggering compliance checks. Updated packages to avoid this. This may fail tests. If it doesn't, we should consider removing these projects.


armanio123:
Can we instead have an exception on the mock projects?

armanio123 (approved):

joj:
What do you mean? Removing them completely?

armanio123:
Sorry, no I mean an exception on the compliance check.

joj:
That's a possibility, but I'd rather have all of our compliance be super accurate right now.

PR #2502, opened 2024-09-11
Persist user choice of dismissing migration gold bar.

safacero:
Issue #
Currently, if a user dismisses the migration gold bar, the bar comes up again when reloading the project. This change saves this information on the .suo file and checks every time a project is loaded.

I'm also removing old code that was supposed to check the .user file. We don't generate a .user file on our templates, this was rarely checked and was not persisting user actions.

### DIFF
--- a/Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
+++ b/Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
@@ -1,31 +1,25 @@
 ﻿using System;
-using System.IO;
-using System.Linq;
-using System.Runtime.CompilerServices;
-using System.Runtime.InteropServices;
-using System.Xml.Linq;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.OLE.Interop;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
-using IServiceProvider = System.IServiceProvider;
 
 namespace Microsoft.NodejsTools.Project
 {
-    internal abstract class MigrateToJspsInfoBar
+    internal static class MigrateToJspsInfoBar
     {
         private static IVsInfoBarUIElement CurrentInfoBarElement;
         private const string UserClosedMigrationPrompt = "UserClosedMigrationPrompt";
 
         public static void Show(IVsShell vsShell, IVsInfoBarUIFactory infoBarUiFactory, IVsSolutionPersistence solutionPersistence, string projectFileLocation)
         {
-            var userPreference = new PersistUserPreferences();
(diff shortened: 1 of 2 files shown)


Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
```diff
@@ -117,6 +63,67 @@ void MigrateProjectCommand()
...
+            // This method writes the boolean value to the .suo file
+            public int WriteUserOptions(IStream pOptionsStream, string pszKey)
+            {
+                ThreadHelper.ThrowIfNotOnUIThread();
+                if (pszKey == UserClosedMigrationPrompt)
+                {
+                    uint outBytesWritten = 0;
+
+                    // Convert the boolean to string and then to a byte array
+                    byte[] buffer = System.Text.Encoding.UTF8.GetBytes(dismissedMigration.ToString());
```
PranavSenthilnathan, line 117: Why does `dismissedMigration` need to be stored as a string? Why can't it just be stored as a 0/1?

Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
```diff
@@ -94,13 +40,13 @@ public static void Show(IVsShell vsShell, IVsInfoBarUIFactory infoBarUiFactory,
...
+            Action<string> OnClose = (userFile) =>
             {
-                CurrentInfoBarElement.Unadvise(eventCookie);
-                WritePropertyToUserFile(filePath);
+                userPreference.dismissedMigration = true;
+                solutionPersistence.SavePackageUserOpts(userPreference, UserClosedMigrationPrompt);
             };
             CurrentInfoBarElement = infoBarUiFactory.CreateInfoBar(infoBarModel);
-            CurrentInfoBarElement.Advise(new InfoBarUIEvents(OnClose, userFile), out eventCookie);
+            CurrentInfoBarElement.Advise(new InfoBarUIEvents(OnClose, ""), out eventCookie);
```
PranavSenthilnathan, line 49: Why is `""` needed? Remove from the `InfoBarUIEvents` ctor and `Action` above.

PranavSenthilnathan (changes requested):

Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
safacero: It can, updated now!

Nodejs/Product/Nodejs/Project/MigrateToJspsInfoBar.cs
safacero: Updated.

PranavSenthilnathan (approved):

PR #2510, opened 2024-09-26
First version of StartWebBrowser.

safacero:
Issue # We were not properly migration some Debug properties settings, this PR migrates the StartWebBrowser property behavior.

launch.json for a Web app with StartWebBrowser = true
image

launch.json for a Web app with StartWebBrowser = false
image

launch.json for a Console app
image

### DIFF
--- a/Nodejs/MigrateToJsps/Jsps/LaunchJsonWriter.cs
+++ b/Nodejs/MigrateToJsps/Jsps/LaunchJsonWriter.cs
@@ -89,13 +89,24 @@ namespace MigrateToJsps
                 var chromeLaunch = ChromeLaunchTemplate;
                 chromeLaunch.Url = chromeLaunch.Url.Replace("*PORT*", port);
 
-                Configuration[] launchConfigs = new Configuration[]
+                Configuration[] launchConfigs;
+
+                if (bool.Parse(njsprojFileModel.StartWebBrowser))
                 {
-                    edgeLaunch, chromeLaunch, nodeLaunch
-                };
+                    launchConfigs = new Configuration[]
+                    {
+                        edgeLaunch, chromeLaunch, nodeLaunch
+                    };
+                }
+                else
+                {
+                    launchConfigs = new Configuration[]
+                    {
+                        nodeLaunch, edgeLaunch, chromeLaunch
+                    };
+                }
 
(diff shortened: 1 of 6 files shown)


Nodejs/Product/Nodejs/ProjectTemplates/TypeScriptAzureWebRole/TypeScriptWebApp.njsproj
```diff
@@ -12,7 +12,6 @@
     <ProjectGuid>$guid1$</ProjectGuid>
     <ProjectHome>.</ProjectHome>
     <StartupFile>server.ts</StartupFile>
-    <StartWebBrowser>True</StartWebBrowser>
```
PranavSenthilnathan, line 15: these seem like local changes

safacero: Turns out it isn't! This property is duplicated in some files for some reason. You can see another one in line 23 if you expand the file! This was breaking me for picking up in C# but it was working on MSBuild because MSBuild will pick the last value it computes (if there's two).

Nodejs/MigrateToJsps/Jsps/LaunchJsonWriter.cs
```diff
@@ -89,13 +89,24 @@ public static void CreateLaunchJson(string projectDir, NjsprojFileModel njsprojF
                 var chromeLaunch = ChromeLaunchTemplate;
                 chromeLaunch.Url = chromeLaunch.Url.Replace("*PORT*", port);
 
-                Configuration[] launchConfigs = new Configuration[]
+                Configuration[] launchConfigs;
+
+                if (bool.Parse(njsprojFileModel.StartWebBrowser))
```
PranavSenthilnathan, line 94: As far as I understand, the ordering here is not going to achieve what you're looking for. When StartWebBrowser is set, we want VS to start both the browser and node  (i.e. use the compound config) and when it's not set it should just start node. Since VS will prioritize single configs, the StartWebBrowser=true scenario requires the user to select the compound config from the dropdown regardless of ordering. If we want to provide a better experience, we can put the compound config in the user file so that it will be selected by default in the dropdown when the migration finishes.

PR #2515, opened 2025-01-21
Fixing GetSdkVersion GetDirectories() bug.

safacero:
We got a bug from GetDirectories(); when no folder is found.

### DIFF
--- a/Nodejs/MigrateToJsps/Jsps/EsprojFileWriter.cs
+++ b/Nodejs/MigrateToJsps/Jsps/EsprojFileWriter.cs
@@ -114,16 +114,26 @@ namespace MigrateToJsps
 
         private static string GetSdkVersion()
         {
-            // Use the installed version of JSPS on the NuGet fallback folder.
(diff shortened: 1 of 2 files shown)


Nodejs/MigrateToJsps/Jsps/EsprojFileWriter.cs
```diff
@@ -114,16 +114,26 @@ private static JObject ReadPackageJson(string projectDir)
...
+                        Version.TryParse(Path.GetFileName(v), out var version);
+                        return version;
+                    })
+                    .OrderByDescending(v => v)
+                    .FirstOrDefault();
+            }
+            catch (Exception)
+            {
+                // If the SDK version is not found on NuGet fallback folder, use the one hardcoded on EsProjFileModel
+                return null;
```
joj, line 135: should this return the sdk constant?

joj (approved):

Nodejs/MigrateToJsps/Jsps/EsprojFileWriter.cs
safacero: Leaving here for documentation: I am intercepting sdk version, but if null it falls back to the current flow of using the hardcoded one on esprojmodel.

